PR #41, opened 2024-02-16
Chore: add github ci workflows

aaronsteers:
Adds CI workflows


aaronsteers:
@bindipankhudi - It's interesting that `test_docs.py` failed on `3.11` and not `3.10`. This might be related to the issue you were seeing today.

https://github.com/airbytehq/PyAirbyte/actions/runs/7927029670/job/21642726068?pr=41

If docs generates different output when run on py3.10 vs py3.11, then it might make sense to only run on one or the other.

PR #52, opened 2024-02-19
Feat: Streamline stream selection and support for `select_streams("*")`

aaronsteers:
This PR allows a more streamlined single-step source creation and setup.

Instead of requiring multiple steps:

```py
    source = ab.get_source(
        "source-github",
        config=GITHUB_CONFIG,
    )
    source.select_streams(["pull_requests", "issues"])
    read_result = source.read()
```

We can now do all in a single step:

```py
    read_result = ab.get_source(
        "source-github",
        config=GITHUB_CONFIG,
        streams=["pull_requests", "issues"],
    ).read()
```

If `"*"` is specified, it will be treated as "all", equivalent to `select_all_streams()`.

For tutorials, the multi-step process is often preferable, because it allows each step to be debugged in turn. However, for demos, the reverse is true: fewer lines and fewer statement makes the process easier to read and easier to understand overall.

This also allows streams to be specified in `read()`.

### DIFF
--- a/.github/workflows/python_lint.yml
+++ b/.github/workflows/python_lint.yml
@@ -68,4 +68,4 @@ jobs:
 
     # Job-specifc step(s):
     - name: Check code format
-      run: poetry run ruff format --check .
+      run: poetry run mypy .
--- a/.github/workflows/python_lint.yml
+++ b/.github/workflows/python_lint.yml
@@ -67,5 +67,5 @@ jobs:
         cache: 'poetry'
 
     # Job-specifc step(s):
-    - name: Check code format
+    - name: Check MyPy typing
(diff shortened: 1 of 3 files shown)


.github/workflows/python_lint.yml
```diff
@@ -67,5 +67,5 @@ jobs:
         cache: 'poetry'
 
     # Job-specifc step(s):
-    - name: Check code format
-      run: poetry run ruff format --check .
+    - name: Check MyPy typing
+      run: poetry run mypy .
```
aaronsteers, line 71: Drive-by change to fix this check.

bindipankhudi (approved):
Looks good!

Issue #57, opened 2024-02-21
Bug: Type conversion failure for `events` stream of `source-klaviyo` when run in PyAirbyte


aaronsteers:
Reported [in slack](https://airbytehq-team.slack.com/archives/C06FZ238P8W/p1708526473508759).

`Could not convert 'false' with type str: tried to convert to boolean`

I checked the [`events` schema](https://github.com/airbytehq/airbyte/blob/master/airbyte-integrations/connectors/source-klaviyo/source_klaviyo/schemas/events.json) for Klaviyo and it doesn't seem to have any boolean-typed fields, which is odd.

Here's the full traceback:

```
>>> cache = ab.get_default_cache()
>>> source.select_streams(["events"])
>>> result = source.read(cache=cache)
Started `source-klaviyo` read operation at 12:56:42...
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/gsb-eng/prog/envs/andromeda/lib/python3.11/site-packages/airbyte/source.py", line 567, in read
    cache.process_airbyte_messages(
  File "/Users/gsb-eng/prog/envs/andromeda/lib/python3.11/site-packages/airbyte/_processors.py", line 218, in process_airbyte_messages
    record_batch = pa.Table.from_pylist(stream_batch)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pyarrow/table.pxi", line 1876, in pyarrow.lib._Tabular.from_pylist
  File "pyarrow/table.pxi", line 5327, in pyarrow.lib._from_pylist
  File "pyarrow/table.pxi", line 3969, in pyarrow.lib.Table.from_arrays
  File "pyarrow/table.pxi", line 1448, in pyarrow.lib._sanitize_arrays
  File "pyarrow/table.pxi", line 1429, in pyarrow.lib._schema_from_arrays
  File "pyarrow/array.pxi", line 344, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 42, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 154, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 91, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Could not convert 'false' with type str: tried to convert to boolean
```

aaronsteers:
[...]

PR #58, opened 2024-02-21
Test: Add script to streamline integ test with any source

aaronsteers:
~~This is not yet tested, and needs a bit more work.~~

This is working now. 😄 

The goal here is to be able to run a sync using any source by running the script and passing the name of the source you want to test:

```
poetry run python examples/run_integ_test_source.py SOURCE_NAME
```

Or to test a specific stream:

```bash
poetry run python examples/run_integ_test_source.py SOURCE_NAME STREAMS_CSV
```

If provided, the second `STREAMS_CSV` arg is expected to be a comma-delimited string (not a file) with one or more streams. For instance: `events` (with no comma) or `events,customers,users`.

For instance, here are the commands I used for testing Source Klaviyo:

```bash
# Run all streams
poetry run python examples/run_integ_test_source.py source-klaviyo

# Run just the events stream
poetry run python examples/run_integ_test_source.py source-klaviyo events
```

### DIFF
--- a/examples/run_integ_test_source.py
+++ b/examples/run_integ_test_source.py
@@ -13,7 +13,7 @@ from google.cloud import secretmanager
 
 import airbyte as ab
 
-def get_integ_test_config(connector_name: str):
+def get_integ_test_config(secret_name: str):
     if "GCP_GSM_CREDENTIALS" not in os.environ:
         raise Exception(
             f"GCP_GSM_CREDENTIALS env variable not set, can't fetch secrets for '{connector_name}'. "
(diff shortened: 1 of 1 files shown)


aaronsteers:
First intended application is to try to repro and diagnose this issue:

- https://github.com/airbytehq/PyAirbyte/issues/57

aaronsteers:
Merging, as this appears to be working well now. I expect we will continue to iterate and improve as we go, for instance, by looping through streams one at a time, and perhaps running additional debug checks to narrow down the issue when a failure occurs.

PR #60, opened 2024-02-22
Feat: Replace Parquet File Writer with Gzipped Jsonl File Writer

aaronsteers:
Add Jsonl file writer.

This file writer better supports variable schemas that were breaking the Parquet writer.

### DIFF
--- a/airbyte/_file_writers/jsonl.py
+++ b/airbyte/_file_writers/jsonl.py
@@ -1,0 +1,65 @@
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+
+"""A Parquet cache implementation."""
+from __future__ import annotations
+
+import gzip
+from pathlib import Path
+from typing import cast
+
+import orjson
+import pyarrow as pa
+import ulid
+from overrides import overrides
+
+from airbyte._file_writers.base import (
+    FileWriterBase,
+    FileWriterBatchHandle,
+    FileWriterConfigBase,
+)
+
+
+class JsonlWriterConfig(FileWriterConfigBase):
+    """Configuration for the Snowflake cache."""
+
+    # Inherits `cache_dir` from base class
+
+
+class JsonlWriter(FileWriterBase):
+    """A Jsonl cache implementation."""
+
+    config_class = JsonlWriterConfig
+
+    def get_new_cache_file_path(
+        self,
+        stream_name: str,
+        batch_id: str | None = None,  # ULID of the batch
+    ) -> Path:
+        """Return a new cache file path for the given stream."""
+        batch_id = batch_id or str(ulid.ULID())
+        config: JsonlWriterConfig = cast(JsonlWriterConfig, self.config)
+        target_dir = Path(config.cache_dir)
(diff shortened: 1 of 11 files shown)


aaronsteers (commented):
I tested this successfully with `source-shopify` and `source-klaviyo`.

airbyte/_file_writers/jsonl.py
```diff
@@ -0,0 +1,68 @@
...
+        target_dir.mkdir(parents=True, exist_ok=True)
+        return target_dir / f"{stream_name}_{batch_id}.jsonl.gz"
+
+    @overrides
+    def _write_batch(
+        self,
+        stream_name: str,
+        batch_id: str,
+        record_batch: pa.Table,
+    ) -> FileWriterBatchHandle:
```
bindipankhudi, line 54: For my understanding, the parquet implementation has a logic for missing columns, but don't see that here, why is that ?

bindipankhudi (approved):
Very clean! Looks good to me.

airbyte/_file_writers/jsonl.py
aaronsteers: Replied in DM. TL;DR is that I was able to add them later on in the process, during insert. But for other cache types, this might be trickier.

Parquet being colunnar, it's easy and fast to add a couple blank columns. I'm trying to avoid the cost of inserting blank records to all the json records if I can, because I expect it would be not very efficient to do so.

aaronsteers:
@bindipankhudi - Just FYI, there's apparently a bug in DuckDB where if you try to load a blank json file, it will just hang indefinitely. Because of this, I had to refactor the no-data treatment slightly. I modified it so that we don't actually create a batch, but we do still loop through all the streams and finalize all of them. During the finalize step, if there are no batches, then we exit after making sure the table exists.

airbyte/caches/base.py
```diff
@@ -545,17 +545,6 @@ def _finalize_batches(
...
-            if not batches_to_finalize:
-                return {}
-
-            files: list[Path] = []
-            # Get a list of all files to finalize from all pending batches.
-            for batch_handle in batches_to_finalize.values():
-                batch_handle = cast(FileWriterBatchHandle, batch_handle)
-                files += batch_handle.files
-            # Use the max batch ID as the batch ID for table names.
-            max_batch_id = max(batches_to_finalize.keys())
```
aaronsteers, line 557: Moved this down below _ensure_table_exists() section, so we can skip it if there are no batches to process.

airbyte/caches/base.py
```diff
@@ -957,6 +958,11 @@ def register_source(
         This method is called by the source when it is initialized.
         """
         self._source_name = source_name
+        self.file_writer.register_source(
+            source_name,
+            incoming_source_catalog,
+            stream_names=stream_names,
+        )
```
aaronsteers, line 965: I don't know if file writers were getting the source registered and expected stream names initialized. This change makes sure when the cache gets its source registered, the file writer does also.

PR #62, opened 2024-02-23
Chore: Revamp anonymous usage stats

aaronsteers:
Resolves: https://github.com/airbytehq/airbyte-internal-issues/issues/6315

Changes:

- Remove the need for _get_telemetry_info() methods everywhere.
- Move discovery functions from `telemetry` module into `_util.meta`.
- Rename telemetry module under `_util` private module name.
- Add `safe_logging_dict` to generate a dictionary of non-sensitive fields, including the class name and the message reported (without `context`). Also includes what we define as 'safe_attrs', which are currently `connector_name`, `violation`, and `exit_code`.
- Move telemetry tests to `unit_tests/test_anonymous_usage_stats.py`


aaronsteers:
@bindipankhudi - This is ready for review.

bindipankhudi (approved):
All updates makes sense to me! One question - Are we thinking of something like IP to identify repeat users ? We should eventually keep track of DAUs and MAUs.

PR #65, opened 2024-02-23
Feat: Add BigQuery cache support; Fix: IPython rendering bug when using a TTY

bindipankhudi:
Added the following:
- BigQuery cache - still uses the config/cache, and parquet, so you won't be able to run the test file at the moment. i will fix that shortly.
- test file - examples/run_bigquery_faker.py

For append and replace cases, data gets written to final table and temp table also gets removed. Writing to _airbytelib_state fails due to some datetime/timestamp mismatch.


airbyte/caches/bigquery.py
```diff
@@ -0,0 +1,166 @@
...
+        json_schema_property_def: dict[str, str | dict | list],
+    ) -> sqlalchemy.types.TypeEngine:
+        """Convert a value to a SQL type.
+
+        We first call the parent class method to get the type. Then if the type is VARCHAR or BIGINT, we
+        replace it with respective BigQuery types.
+        """
+        sql_type = super().to_sql_type(json_schema_property_def)
+        # to-do: replace hardcoded return types with some sort of snowflake Variant equivalent
+        if isinstance(sql_type, sqlalchemy.types.VARCHAR):
```
bindipankhudi: Like I mentioned, I didn't find a VARIANT type. There are methods like _visit_VARCHAR and such in the bigquery libary but not for external use: https://github.com/googleapis/python-bigquery-sqlalchemy/blob/main/sqlalchemy_bigquery/base.py (line 605)

airbyte/caches/bigquery.py
```diff
@@ -0,0 +1,166 @@
...
+                        context={
+                            "temp_table_name": temp_table_name,
+                        },
+                    )
+                
+                credentials = service_account.Credentials.from_service_account_file(self.config.credentials_path)
+                
+                # timestamp columns need to be converted to datetime to work with pandas_gbq
+                # to-do: generalize the following to all columns of column type. This change is to test specically with faker source.
+                dataframe['created_at'] = pd.to_datetime(dataframe['created_at'])
```
bindipankhudi: just pointing out - datatype timestamp needs to be converted to datetime to work, it's hardcoded to faker source for now.

airbyte/_processors/sql/base.py
```diff
@@ -435,9 +434,6 @@ def _create_table(
         column_definition_str: str,
         primary_keys: list[str] | None = None,
     ) -> None:
-        if DEBUG_MODE:
-            assert table_name not in self._get_tables_list(), f"Table {table_name} already exists."
-
```
aaronsteers, line 440: This, and the below, were significantly slowing down execution, at the cost of approx. 30 seconds per call to _get_tables_list(). Removed here, and also below.

airbyte/_processors/sql/base.py
```diff
@@ -804,8 +795,8 @@ def _merge_temp_table_to_final_table(
         columns = {self._quote_identifier(c) for c in self._get_sql_column_definitions(stream_name)}
         pk_columns = {self._quote_identifier(c) for c in self._get_primary_keys(stream_name)}
         non_pk_columns = columns - pk_columns
-        join_clause = "{nl} AND ".join(f"tmp.{pk_col} = final.{pk_col}" for pk_col in pk_columns)
-        set_clause = "{nl}    ".join(f"{col} = tmp.{col}" for col in non_pk_columns)
+        join_clause = f"{nl} AND ".join(f"tmp.{pk_col} = final.{pk_col}" for pk_col in pk_columns)
+        set_clause = f"{nl}  , ".join(f"{col} = tmp.{col}" for col in non_pk_columns)
```
aaronsteers, line 799: A bug fix for an already-existing bug. Bigquery is the first to use this base method, hence finding and fixing this but as part of this PR.

airbyte/_processors/sql/base.py
```diff
@@ -908,12 +899,14 @@ def _emulated_merge_temp_table_to_final_table(
             conn.execute(update_stmt)
             conn.execute(insert_new_records_stmt)
 
-    @final
     def _table_exists(
         self,
         table_name: str,
     ) -> bool:
-        """Return true if the given table exists."""
+        """Return true if the given table exists.
```
aaronsteers, line 906: This was super slow for bigquery. I overrode this with a custom bigquery implementation which uses a different API to check for table existence.

airbyte/_processors/sql/bigquery.py
```diff
@@ -0,0 +1,204 @@
...
+    file_writer_class = JsonlWriter
+    type_converter_class = BigQueryTypeConverter
+    supports_merge_insert = True
+
+    cache: BigQueryCache
+
+    def __init__(self, cache: CacheBase, file_writer: FileWriterBase | None = None) -> None:
+        self._credentials: service_account.Credentials | None = None
+        self._schema_exists: bool | None = None
+        super().__init__(cache, file_writer)
```
aaronsteers, line 65: Adding _schema_exists and _credentials as object-level markers. The _credentials attribute holds the credentials object after we load it the first time in `_get_credentials()` and the `_schema_exists` bool helps us bypass redundant schema exists checks in _ensure_schema_exists()` .

airbyte/_processors/sql/bigquery.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # Specify the table ID (in the format `project_id.dataset_id.table_id`)
+        table_id = f"{self.cache.project_name}.{self.cache.dataset_name}.{temp_table_name}"
+
+        # Initialize a BigQuery client
+        client = bigquery.Client(credentials=self._get_credentials())
+
+        for file_path in files:
+            with Path.open(file_path, "rb") as source_file:
+                load_job = client.load_table_from_file(  # Make an API request
```
aaronsteers, line 108: I looked through several API options, and the `google-cloud-bigquery` library appears to be the only one which can natively load a jsonl file without first uploading it to GCS. The other libraries were converting the data to CSV or Parquet behind the scenes, which was causing irreconcilable typing issues.

airbyte/_processors/sql/bigquery.py
```diff
@@ -0,0 +1,204 @@
...
+
+        sql = f"CREATE SCHEMA IF NOT EXISTS {self.cache.schema_name}"
+        try:
+            self._execute_sql(sql)
+        except Exception as ex:
+            # Ignore schema exists errors.
+            if "already exists" not in str(ex):
+                raise
+
+        self._schema_exists = True
```
aaronsteers, line 145: This implementation doesn't check - it just runs 'CREATE IF NOT EXISTS ...'. And storing the marker "True" in `self._schema_exists` means that at most we'll run this once per cache object.

airbyte/caches/_catalog_manager.py
```diff
@@ -50,7 +51,9 @@ class StreamState(Base):  # type: ignore[valid-type,misc]
     stream_name = Column(String)
     table_name = Column(String, primary_key=True)
     state_json = Column(String)
-    last_updated = Column(DateTime(timezone=True), onupdate=func.now(), default=func.now())
+    last_updated = Column(
+        DateTime(timezone=True), onupdate=datetime.now(utc), default=datetime.now(utc)
+    )
```
aaronsteers, line 56: Bigquery doesn't have a 'now()' function, so I modified to use Python to determine the timestamp before inserting.

aaronsteers (approved):
This is working well now. Some implementation details were just not possible with the sqlalchemy library, so those are implemented using the `google-cloud-bigquery` library.

aaronsteers:
@bindipankhudi - This one is ready to merge. 🚀 

Thanks for working out all the implementation details and type conversion issues. I noted my changes in a review above.

I'll probably merge this in today or tomorrow, since it contains some other bug fixes and optimizations which will be helpful broadly.

bindipankhudi:
Thank you so much for working out all the issues here!!  🙏I really
appreciate it!! I was out for the evening so just saw this. I read all your
comments which make sense to me. I will look at the PR in detail tomm
morning, but feel free to merge anytime since there are so many other
changes in here.

examples/run_bigquery_faker.py
```diff
@@ -0,0 +1,64 @@
...
+
+warnings.filterwarnings("ignore", message="Cannot create BigQuery Storage client")
+
+
+# load secrets from GSM using the GCP_GSM_CREDENTIALS env variable
+secret_client = secretmanager.SecretManagerServiceClient.from_service_account_info(
+    json.loads(os.environ["GCP_GSM_CREDENTIALS"])
+)
+
+# to-do: make sure database & account id are read from PY_AIRBYTE_BIGQUERY_CREDS
```
bindipankhudi: nit.. :) change the comment to match the secret name we are using

bindipankhudi (commented):
Everything looks great! Great call on using the Google Bigquery library!

examples/run_bigquery_faker.py
aaronsteers: Great catch. Thanks! I think we can actually remove this comment...

```suggestion
```

PR #67, opened 2024-02-25
Fix: Resolve conflicting schema issues by removing dependency on PyArrow

aaronsteers:
Resolves: #89 

We've discovered through user testing that the PyArrow dependency, and PyArrow itself, are inherently fragile when it comes to JSON schemas with variable typing. PyArrow exceeds at handling large volumes of typed data very quickly, but there's no way to process ragged and potentially conflicting types without reverting json types to strings.

Paqruet, similarly, has problems rendering json structures without child nodes. So, something like "`extra_properties: {}`" would cause a runtime error.

We've recently moved our file writer to a JSONL file writer, in order to better handle ragged, non-populated objects, and unpredictable types. However, we still have a dependency on `pyarrow`, which is problematic for some users' workloads.

This PR will remove the `pyarrow` dependency and replace it with a flow that simply writes JSONL lines to local files as quickly as possible.


aaronsteers:
Performance snapshot, prior to changes (approx 15K records per second):

```
AJ-Steers-MacBook-Pro---GGHWM7QWPJ:airbyte-lib-private-beta ajsteers$ poetry run python ./examples/run_faker.py

Enter the value for secret 'DUMMY_SECRET': 
Installing Faker source...
Faker source installed.
Connection check succeeded for `source-faker`.
Started `source-faker` read operation at 09:23:55...

                                                    Read Progress                                                     

Started reading at 17:23:55.                                                                                          

Read 500,100 records over 33 seconds (15,154.5 records / second).                                                     

Wrote 500,100 records over 53 batches.                                                                                

Finished reading at 17:24:28.                                                                                         

Started finalizing streams at 17:24:28.                                                                               

Finalized 53 batches over 3 seconds.                                                                                  

Completed 3 out of 3 streams:                                                                                         

 • purchases                                                                                                          
 • products                                                                                                           
 • users                                                                                                              

Completed writing at 17:24:32. Total time elapsed: 36 seconds                                                         

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Completed `source-faker` read operation at 09:24:32.
Stream products: 100 records
Stream users: 250000 records
Stream purchases: 250000 records
AJ-Steers-MacBook-Pro---GGHWM7QWPJ:airbyte-lib-private-beta ajsteers$ 
```

aaronsteers:
Refactoring is functionally complete - although still some cleanup to do.

Performance is boosted by approximately 30% - from 15K records per second to 20K records per second.

```
AJ-Steers-MacBook-Pro---GGHWM7QWPJ:PyAirbyte ajsteers$ poetry run python ./examples/run_faker.py

Enter the value for secret 'DUMMY_SECRET': 
Installing Faker source...
Faker source installed.
Connection check succeeded for `source-faker`.
Started `source-faker` read operation at 10:51:43...
Completed `source-faker` read operation at 10:52:13.
Stream products: 100 records
Stream users: 250000 records
Stream purchases: 250000 records

                                               Read Progress                                               

Started reading at 18:51:43.                                                                               

Read 500,100 records over 25 seconds (20,004.0 records / second).                                          

Wrote 500,100 records over 51 batches.                                                                     

Finished reading at 18:52:09.                                                                              

Started finalizing streams at 18:52:09.                                                                    

Finalized 51 batches over 3 seconds.                                                                       
                                                    ...                                                    AJ-Steers-MacBook-Pro---GGHWM7QWPJ:PyAirbyte ajsteers$ 
```

aaronsteers:
@bindipankhudi - Tests are passing locally and lint cleanup is complete.

I am taking another pass through the code myself but wanted to ping you of the latest status.

bindipankhudi:
Sounds great! I am taking a look also, but don't wait on it. Feel free to
check-in,

airbyte/_batch_handles.py
```diff
@@ -0,0 +1,83 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
+"""Batch handle class."""
```
aaronsteers, line 2: This class has been refactored and moved out. 

It now handles its own file creation and deletion. Deletion is generic, and creation is passed as a function ("file_opener") to the contstructor.

This class now also tracks whether the batch handle itself has been finalized or not - which makes tracking easier and more consistent across parts of the codebase which need to know this.

airbyte/_batch_handles.py
```diff
@@ -0,0 +1,83 @@
...
+        """Return the open file writer, if any, or None."""
+        return self._open_file_writer
+
+    def close_files(self) -> None:
+        """Close the file writer."""
+        if self.open_file_writer is None:
+            return
+
+        # with suppress(Exception):
+        self.open_file_writer.close()
```
aaronsteers:
```suggestion
        with suppress(Exception):
            self.open_file_writer.close()
```

airbyte/_processors/file/parquet.py
aaronsteers, line 1: FYI - I'm considering whether to delete this file entirely. (Probably will, but not sure if I'll do it here in this PR or a subsequent one.)

airbyte/_processors/sql/base.py
```diff
@@ -160,11 +161,11 @@ def register_source(
...
-        self.file_writer.register_source(
-            source_name,
-            incoming_source_catalog,
-            stream_names=stream_names,
-        )
+        # self.register_source(
+        #     source_name,
+        #     incoming_source_catalog,
+        #     stream_names=stream_names,
+        # )
```
aaronsteers:
```suggestion
```

airbyte/_processors/base.py
```diff
@@ -203,150 +187,44 @@ def process_airbyte_messages(
...
-        # We are at the end of the stream. Process whatever else is queued.
-        for stream_name, stream_batch in stream_batches.items():
-            batch_df = pd.DataFrame(stream_batch)
-            record_batch = pa.Table.from_pandas(batch_df)
-            self._process_batch(stream_name, record_batch)
-            progress.log_batch_written(stream_name, len(stream_batch))
-
-        all_streams = list(self._pending_batches.keys())
-        # Add empty streams to the streams list, so we create a destination table for it
-        for stream_name in self.expected_streams:
```
bindipankhudi, line 215: are we still tackling empty streams downstream ?

airbyte/_processors/file/parquet.py
aaronsteers: (Deleted.)

bindipankhudi (approved):
Quite some refactoring. Looks great and cleaner than before. One question - What was the reason to use PyArrow in the first place. Is this because we were using Parquet at the time?

airbyte/_processors/base.py
aaronsteers: @bindipankhudi - Great question - and I just double checked. Yes, we are still running regardless. Here's how it works now in the `SqlProcessorBase` class:

1. `write_all_stream_data()` is called at the end of `process_airbyte_messages()`.
2. `write_all_stream_data()` calls `write_stream_data()` on all values contained in `expected_streams`, so no streams are missed even data is not received for the stream in question.
3. `write_stream_data()` does its processing flow (including `_ensure_schema_exists()` even if there are no records/batches to process.

PR #96, opened 2024-03-05
Chore: Add `.gitignore` rules for secrets dirs

aaronsteers:
If we run `connectors_ci`, it will create these directories. This PR ensures we keep git history safe from accidentally committed secrets in `secrets` or `.secrets` dirs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
+# Directories and subdirectories called 'secrets' or '.secrets'
+.secrets
+secrets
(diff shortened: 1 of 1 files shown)


aaronsteers:
@bindipankhudi - FYI. I'm going to go ahead and merge this to prevent secrets usage while I begin working with the `ci_credentials` tool.

No action needed, but wanted to tag you for visibility.

Issue #35, opened 2024-02-12
🐛 Bug (non-blocking): Colab prints annoying/confusing version conflicts during `pip install`


aaronsteers:
Colab prints the below confusing messages of version conflicts during pip install.

These look like fatal errors, but they are not. I haven't been able to find a way to suppress these but perhaps we can relax or adjust some version constraints to reduce the noise here:

 
![image](https://github.com/airbytehq/airbyte-lib-private-beta/assets/18150651/df1f3c77-13d0-405b-aa76-47379315660a)

aaronsteers:
Resolved by #97:

> Test successfully with this notebook:
> 
> https://colab.research.google.com/drive/1pBT1RVf16psMRY3zzSSDr3H_GsyDNR6r#scrollTo=TEzG3kXznVtz
> 
> Before this change, we get many conflict errors and Colab prompts us we might need to restart the notebook:
> 
> > <img width="711" alt="image" src="https://github.com/airbytehq/PyAirbyte/assets/18150651/ca992bc3-ea9b-4105-adfb-9ab67771fd9c">
> 
> 
> After this change, the log is much cleaner:
> 
> > <img width="684" alt="image" src="https://github.com/airbytehq/PyAirbyte/assets/18150651/10c7038f-fa4a-4b93-8cf2-6b1c1e52076c">
>

PR #97, opened 2024-03-05
Chore: Loosen version requirements to match Colab defaults

aaronsteers:
Resolves: #35 

In the latest version, complains about version conflicts and asks users to reload the Notebook.

This PR aims to widen supported versions for `google.auth` and `pandas`, to avoid conflicts with the default Colab environment.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2795,4 +2795,4 @@ files = [
 [metadata]
 lock-version = "2.0"
 python-versions = ">=3.9,<4.0"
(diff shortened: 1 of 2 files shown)


aaronsteers:
Before this change, we get many conflict errors and Colab prompts us we might need to restart the notebook:

> image


After this change, the log is much cleaner:

> image

aaronsteers:
This is tested and should be safe. Merging so it can be part of pre-release testing this morning...

PR #98, opened 2024-03-05
Fix: Resolve Python 3.9 compatibility issues, expand CI test matrix

aaronsteers:
This adds Python 3.9 and 3.11 back into the test matrix.

Findings:

- Python 3.11 worked without any issue.
- Python 3.9 was failing because Pydantic wants to parse/evaluate type hints for classes derived from `pydantic.BaseModel`.
  - I disabled the lint rule that dissallowed the legacy syntaxes `Union[x | y]` and `Optional[x]`.
  - In any subclasses of `BaseModel`, I reverted to legacy-style syntax declarations, to appease the Pydantic eval statements.

### DIFF
--- a/.github/workflows/python_pytest.yml
+++ b/.github/workflows/python_pytest.yml
@@ -48,10 +48,9 @@ jobs:
     strategy:
       matrix:
         python-version: [
-          # TODO: Re-enable 3.9 and 3.11 once we have stable tests across all versions.
-          # '3.9',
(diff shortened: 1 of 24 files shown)


aaronsteers:
@bindipankhudi - Another FYI above. No action needed. Going forward, we'll get CI failures if Python 3.9 or Python 3.11 tests fail (in addition to Python 3.10).

PR #109, opened 2024-03-08
Feat: Clean up public API and improve docs

### LINKED ISSUE #22: Clean up public interface
Currently, the source and cache classes expose lots of internals in their constructors and methods because they need to communicate with each other.

However this should be an implementation detail that's not exposed to the user:

```mermaid
classDiagram
    RecordProcessor <|-- SQLCacheBase
    SQLCacheBase <|-- DuckDBCacheBase
    DuckDBCacheBase <|-- DuckDBCache
    SQLCacheBase <|-- SnowflakeCache
    SQLCacheBase <|-- PostgresCache
    class Source{
        __init__(executor: Executor)
        read(cache: SQLCacheBase)
    }
    class SQLCacheBase{
        register_source(source: Source)
        get_state()
    }
    Source --> SQLCacheBase
```

Things like executor, register_source and get_state should not be visible by the user, but they are also not protected/private methods, as they are required for the source/cache interaction.

To fix this, let's hide these things for the end user and only expose clean interfaces:
* Things that are only used within a class can be made protected (like `discovered_catalog`, `configured_catalog`, `executor`)
* Don't expose the actual `Source` class, but only an interface which doesn't know about the constructor (the actual source class becomes `_Source`). The user should use the factory function to create the source.
* Introduce a new class `Cache` which has an `internal_instance: SQLCacheBase` property which is not document via `@private` annotation in the docstring: https://pdoc.dev/docs/pdoc.html#control-what-is-documented - this way it's hidden from the reference docs
* The `Cache` class delegates all public methods to its instance. When communicating with a cache (calling register_source or get_state), the `_Source` class is directly accessing the `internal_instance` to call these methods

This way the user can only "see" the intended public interface while the components can still communicate with each other. All implementation details of caches are hidden away as well.

aaronsteers:
Resolves #22 

This cleans up the public API while improving docs readability and coverage.

### DIFF
--- a/airbyte/__init__.py
+++ b/airbyte/__init__.py
@@ -2,27 +2,34 @@
 
 .. include:: ../README.md
 
+## API Reference
+
 """
 from __future__ import annotations
 
-from airbyte import caches, datasets, registry, secrets
-from airbyte._factories.connector_factories import get_source
+from airbyte import caches, datasets, documents, exceptions, results, secrets, sources
 from airbyte.caches.bigquery import BigQueryCache
 from airbyte.caches.duckdb import DuckDBCache
-from airbyte.caches.factories import get_default_cache, new_local_cache
+from airbyte.caches.util import get_default_cache, new_local_cache
 from airbyte.datasets import CachedDataset
-from airbyte.registry import get_available_connectors
 from airbyte.results import ReadResult
 from airbyte.secrets import SecretSource, get_secret
-from airbyte.source import Source
+from airbyte.sources import registry
+from airbyte.sources.base import Source
+from airbyte.sources.registry import get_available_connectors
+from airbyte.sources.util import get_source
 
 
 __all__ = [
     # Modules
     "caches",
     "datasets",
+    "documents",
+    "exceptions",
     "registry",
+    "results",
     "secrets",
+    "sources",
     # Factories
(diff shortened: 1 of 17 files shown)


airbyte/__init__.py
```diff
@@ -2,27 +2,34 @@
...
     "datasets",
+    "documents",
+    "exceptions",
     "registry",
+    "results",
     "secrets",
+    "sources",
     # Factories
     "get_available_connectors",
     "get_default_cache",
```
aaronsteers, line 35: Important to note: No top-level members of the `airbyte` module have been removed.

The hope is that this won't break users at all, unless they are importing from some lower-level module that has been renamed. There are very few of these - e.g. `airbyte.sources.factories -> airbyte.sources.util`.

Also - `factories` did not feel right to me. To be inclusive of "factory functions" as well as other "helper utilities", I've regrouped these under "util" submodules, where appropriate.

aaronsteers:
https://www.loom.com/share/3abe276bc44241259f951f910fbf54a5?sid=d0403b50-1180-447f-8075-f012519a4527

bindipankhudi (approved):
All makes sense to me. (I watched the Loom). Docs are looking great! It's really cool that we can update usage section in docs by updating in code!

PR #112, opened 2024-03-08
Docs: Add usage samples for each cache option

aaronsteers:
These usage samples will render at the top of each cache's module reference page.

This was very quick to create. I started it accidentally when recording my last loom. Most of the example code was written by copilot actually. 😄 🤖


bindipankhudi (approved):
Very cool!

Issue #108, opened 2024-03-07
🐛 Bug: Mixed case column names from Source S3 is causing empty data when loaded to DuckDB 


divyagowdab:
I tried to read data from a s3 source. The bucket had one csv file. Attached the sample csv file.
In the attached file there are 2 colums starting with upper case (Age and Credit_rating). For those 2 columns the duck db does not have data. (attached screenshot)
If i change all the column names to lower case then everything works fine.
And also if the column names are hyphen separated then "missing column" error is thrown.
[TestData.csv](https://github.com/airbytehq/PyAirbyte/files/14529523/TestData.csv)
<img width="1081" alt="Screenshot 2024-03-08 at 1 43 47 AM" src="https://github.com/airbytehq/PyAirbyte/assets/96168512/86aeabd7-3198-4ce5-aad3-62cbdd4f3142">

aaronsteers:
@divyagowdab - This is very helpful feedback - thank you.

We will see if I can reproduce this using the `File` source, and will report back.

aaronsteers:
@divyagowdab - We have a PR ready for review, which should resolve this: #114

PR #114, opened 2024-03-08
Fix: Resolve issue where mixed-case stream properties would result in missing data

aaronsteers:
Resolves: #108

The behavior implemented is as follows:

- Prefer lower case column names in tables, as before.
- Expect mixed-case property names in the incoming data from the source. 

On its own this creates a small problem, which was that accessing the results of Source.get_records() would require different casing than was required to read from data synced to the cache with Source.read().

The way I was able to resolve this difference, is to introduce a case-insensitive dictionary implementation - which as of now is only implemented in the output of Source.get_records().

This mirrors the database behavior of (1) preferring lower-casing for column names, while (2) still allowing case-insensitive matching by default.


airbyte/sources/base.py
```diff
@@ -329,29 +329,21 @@ def get_records(self, stream: str) -> LazyDataset:
...
 
         def _with_logging(records: Iterable[dict[str, Any]]) -> Iterator[dict[str, Any]]:
             self._log_sync_start(cache=None)
             yield from records
             self._log_sync_success(cache=None)
 
-        def _with_missing_columns(records: Iterable[dict[str, Any]]) -> Iterator[dict[str, Any]]:
-            """Add missing columns to the record with null values."""
-            for record in records:
-                existing_properties_lower = lower_case_set(record.keys())
```
aaronsteers, line 342: This "missing columns" treatment is now included in the new "normalize_records" function.

tests/integration_tests/fixtures/source-test/source_test/run.py
```diff
@@ -17,8 +17,8 @@
                     "$schema": "http://json-schema.org/draft-07/schema#",
                     "type": "object",
                     "properties": {
-                        "column1": {"type": "string"},
-                        "column2": {"type": "number"},
+                        "Column1": {"type": "string"},
+                        "Column2": {"type": "number"},
```
aaronsteers, line 21: Modified the test data to ensure coverage of mixed-case top level properties.

airbyte/_processors/sql/duckdb.py
```diff
@@ -93,9 +94,9 @@ def _write_files_to_new_table(
         columns_type_map = indent(
             "\n, ".join(
                 [
-                    f"{self._quote_identifier(c)}: "
-                    f"{self._get_sql_column_definitions(stream_name)[c]!s}"
-                    for c in columns_list
+                    f"{self._quote_identifier(prop_name)}: "
+                    f"{self._get_sql_column_definitions(stream_name)[prop_name.lower()]!s}"
```
bindipankhudi: is there a reason we are not using the LowercaseNormalizer class here and doing .lower()?

bindipankhudi (approved):
Looks great!

airbyte/_processors/sql/duckdb.py
aaronsteers: Good catch! No reason. There are probably a few more cases like this actually, where we can standardize the logic in one place.

An advantage of doing so will be that later, more edge cases can be all handled in one place - for instance, handling of punctuation and special characters.

aaronsteers: Fixed now. ✅

Issue #105, opened 2024-03-07
Add anonymized reporting of failed installs


aaronsteers:
Today we rely on users to let us know if connector installation failures occur.

We should send a telemetry event in these cases (assuming `DO_NOT_TRACK` is not set) so that we can better understand where problems may exist in the the compatibility of specific connectors.

PR #115, opened 2024-03-09
Feat: Telemetry - Added "install" event type to track failed installs

bindipankhudi:
Related to https://github.com/airbytehq/PyAirbyte/issues/105

* Added a new install type event (Currently info send for failed installs is same as what is send for failed syncs)
* Confirmed that segment is receiving these events (sample install event below) 
* Confirmed that new install table is created in bigquery automatically. (screenshot below)
* All tests are passing. 

Screenshot 2024-03-10 at 7 48 31 PM
Screenshot 2024-03-10 at 6 19 47 PM


airbyte/_util/telemetry.py
```diff
@@ -74,12 +74,17 @@
...
 
 
-class SyncState(str, Enum):
+class EventState(str, Enum):
     STARTED = "started"
     FAILED = "failed"
     SUCCEEDED = "succeeded"
 
 
+class EventType(str, Enum):
```
bindipankhudi, line 175: added this enum in case we want to add more types in future.

airbyte/_util/telemetry.py
```diff
@@ -137,9 +158,11 @@ def get_env_flags() -> dict[str, Any]:
 
 
 def send_telemetry(
-    source: Source,
+    source: Source | None,
```
bindipankhudi: Source param can't be mandatory anymore since we don't have enough information to create a Source in most cases when install failures happens. Instead, a source name is passed in those cases.

airbyte/_util/telemetry.py
```diff
@@ -158,6 +180,14 @@ def send_telemetry(
         "application_hash": one_way_hash(meta.get_application_name()),
         "flags": get_env_flags(),
     }
+
```
bindipankhudi, line 266: added this check since both source, and name params can be None.

airbyte/sources/util.py
```diff
@@ -108,26 +111,50 @@ def get_source(
...
+        _log_install_state(None, name, state=EventState.FAILED, exception=e)
+        raise
 
 
 __all__ = [
     "get_source",
 ]
+
+
+def _log_install_state(
```
bindipankhudi, line 142: This is not a class, but prefixed the method with underscore since it's for internal use.

aaronsteers:
Noyce! 🙌

airbyte/_util/telemetry.py
aaronsteers: Wdyt or re-typing `source` as `Source | str`? (For now, at least, and later `Source | str | None` when/if we have events that don't have any source at all.)

In this case, we'd replace the below null checks with `isinstance()` checks - taking one code path if `Source` is `str` and another if it's an object.

tests/unit_tests/test_anonymous_usage_stats.py
```diff
@@ -33,9 +33,11 @@ def test_telemetry_track(monkeypatch):
 
     telemetry.send_telemetry(
         source=source_test,
+        name="source_test",
         cache=cache,
-        state="started",
+        state=telemetry.EventState.STARTED,
```
aaronsteers, line 37: Nice 👍

aaronsteers (changes requested):
This looks awesome. Nice work! One requested change, inline.

airbyte/_util/telemetry.py
```diff
@@ -158,6 +171,12 @@ def send_telemetry(
...
+    if isinstance(source, str):
```
bindipankhudi, line 267: really weird that isinstance(source, Source) keeps giving error saying "Source is not defined" so I did the check on str instead.

airbyte/_util/telemetry.py
bindipankhudi: oh yea, makes a lot of sense! updated.

airbyte/_util/telemetry.py
aaronsteers: Interesting! I think your solution to switch makes sense.

One possible explanation would be if `isinstance()` doesn't get access to the `Source` class because it is imported in the `if TYPE_CHECKING` block.

This way of importing will support type annotations but might not be referenceable directly in code. (Avoids circular dependencies and unnecessary resource loading.)

So, I think it is good how you have it but that's my hypothesis, fwiw.

bindipankhudi: yea, that makes sense.

airbyte/sources/util.py
```diff
@@ -106,28 +109,49 @@ def get_source(
...
+    name: str,
+    state: EventState,
+    exception: Exception | None = None,
+) -> None:
+    """Log an install event."""
+    print(f"{state.value} `{name}` install operation at {pendulum.now().format('HH:mm:ss')}...")
+    send_telemetry(
```
aaronsteers: I think we might already be printing the install start/stop notification. Need to double-check though...

aaronsteers: Confirmed, we are printing start/complete messages here:

https://github.com/airbytehq/PyAirbyte/blob/7c70f14b56e68f2d34108da3b5fc2f88cefaec7e/airbyte/_executor.py#L205-L229

What do you think of removing the print from this method?

```suggestion
    send_telemetry(
```

aaronsteers (approved):
Looks great! One suggestion inline. Otherwise, I think this is ready to go. 🚀

PR #116, opened 2024-03-11
Unpinning Airbyte CDK to get freebie upgrades before 1.0

natikgadzhi:
Okay hear me out. Good news is that we might get free dependabot-triggered updates for the CDK, and get latest bugfixes. That will come handy as we wire up declarative sources! 

Bad news is that we might break things, but we will unpin the CDK in all sources, and we have `Poetry.lock` to protect us from yolo-bug-problems.


aaronsteers:
@natikgadzhi - Rejecting this and replacing with a bump to latest version:

- #116 

Happy to take a dependabot-like auto-update if we have that option open up. However, I don't think we can leave this unpinned until CDK reaches a stable 1.0 status.

As specified in semver:

https://semver.org/

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

PR #124, opened 2024-03-12
Chore: Add anonymous user ID in tracking events

aaronsteers:
This is ready for review.

1. On first run, PyAirbyte will check for an existing analytics file.
2. If not found, and if `DO_NOT_TRACK` is not set, a new ID will be generated and will be stored in a yaml file at `~/.airbyte/analytics.yml`
3. If `AIRBYTE_ANALYTICS_ID` env var is set, then the provided ID will be used instead of a randomly generated one. In this case, the provided analytics ID will also be saved to the analytics file, overriding any previously set value.
4. This PR updates all of our CI jobs to pass in a (newly created) repository variable ([here](https://github.com/airbytehq/PyAirbyte/settings/variables/actions)), which contains a newly generated ULID that is specific to our CI flows.

### DIFF
--- a/airbyte/_util/telemetry.py
+++ b/airbyte/_util/telemetry.py
@@ -105,7 +105,7 @@ def _setup_analytics() -> str | bool:
         _ANALYTICS_FILE.write_text(yaml.dump(new_file_contents))
     except Exception as ex:
         print(
-            f"Failed to create the analytics file at '{_ANALYTICS_FILE}'. " f"Error was: {str(ex)}"
+            f"Failed to create the analytics file at '{_ANALYTICS_FILE}'. " f"Error was: {ex!s}"
         )
     print(
         "Anonymous usage reporting is enabled. For more information or to opt out, please"
--- a/airbyte/_util/telemetry.py
+++ b/airbyte/_util/telemetry.py
@@ -37,10 +37,12 @@ from contextlib import suppress
 from dataclasses import asdict, dataclass
 from enum import Enum
 from functools import lru_cache
-from typing import TYPE_CHECKING, Any
+from pathlib import Path
+from typing import TYPE_CHECKING, Any, cast
 
 import requests
 import ulid
+import yaml
 
 from airbyte import exceptions as exc
 from airbyte._util import meta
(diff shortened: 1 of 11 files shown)


aaronsteers:
When I run locally:

We write the file as plain text but it is intended to be parsed as yaml. The reason for writing as plain text is that this allows us to more easily control formatting and the header comment text.

```
AJ-Steers-MacBook-Pro---GGHWM7QWPJ:.airbyte ajsteers$ pwd
/Users/ajsteers/.airbyte
AJ-Steers-MacBook-Pro---GGHWM7QWPJ:.airbyte ajsteers$ cat analytics.yml
# This file is used by PyAirbyte to track anonymous usage statistics.
# For more information or to opt out, please see
# - https://docs.airbyte.io/pyairbyte/anonymized-usage-statistics
anonymous_user_id: 01HRT5XN7NPS12DWZX5RTXHAE8
```

I realize we still need to have a real URL to point people to. The one above is not a real page as of yet.

airbyte/_util/telemetry.py
```diff
@@ -73,6 +75,76 @@
...
+def _setup_analytics() -> str | bool:
+    """Set up the analytics file if it doesn't exist.
+
+    Return the anonymous user ID or False if the user has opted out.
+    """
+    anonymous_user_id: str | None = None
+    if _ENV_ANALYTICS_ID in os.environ:
+        # If the user has chosen to override their analytics ID, use that value and
+        # remember it for future invocations.
+        anonymous_user_id = os.environ[_ENV_ANALYTICS_ID]
```
bindipankhudi, line 102: Will we allow users to add anything for this ? Does it make sense to make sure it's not malicious, or not super long.

aaronsteers: Good question. I wasn't worrying much about malicious actors - but I could optionally add a check like a regex expression - such as "must be alpha-numeric and under a specific character length..."

If we want to add some checks/protections, I think we should put them both in the file parser and also in the env var parser - since both would be easy to inject different values.

I'm okay also to keep this for a later improvement - but it isn't much work either way. Lmk if you it think it's worth adding protections in this phase.

bindipankhudi (approved):
Looks great! :) On minor comment, non-blocking.

airbyte/_util/telemetry.py
bindipankhudi: later sounds good! no point blocking on this.

PR #125, opened 2024-03-13
Chore: Add telemetry tracking flag for `langchain-airbyte`, fix incorrect telemetry URL in first-time execution

aaronsteers:
An overly simple implementation, admittedly, but a decent start that should work as expected in most cases.

Note: I haven't tested this yet. I may try manually testing in a notebook.

### DIFF
--- a/airbyte/_util/meta.py
+++ b/airbyte/_util/meta.py
@@ -36,6 +36,9 @@ def is_langchain() -> bool:
 
     This checks for the presence of the 'langchain-airbyte' package.
 
+    TODO: A more robust check would inspect the call stack or another flag to see if we are actually
+          being invoked via LangChain, vs being installed side-by-side.
+
     This is cached for performance reasons.
     """
     return "langchain-airbyte" in sys.modules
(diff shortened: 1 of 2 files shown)


airbyte/_util/meta.py
```diff
@@ -30,6 +30,20 @@ def is_ci() -> bool:
...
+    """Return True if running in a Langchain environment.
+
+    This checks for the presence of the 'langchain-airbyte' package.
+
+    TODO: A more robust check would inspect the call stack or another flag to see if we are actually
+          being invoked via LangChain, vs being installed side-by-side.
+
+    This is cached for performance reasons.
+    """
+    return "langchain-airbyte" in sys.modules
```
bindipankhudi: would this also work if someone is using pip install ?

bindipankhudi: once you check-in, i can try it out.

bindipankhudi (approved):

airbyte/_util/meta.py
aaronsteers: Yes. I've just tested and confirmed. The `sys.modules` list contains all Python packages which have been imported into the Python runtime since launch. This includes `langchain_airbyte` _after_ the user has run `import langchain_airbyte` or `from langchain_airbyte import document_loaders`, etc.

PR #134, opened 2024-03-19
Feat: Always run config pre-validation

### LINKED ISSUE #126: Auto-validate source config
We have a method in PyAirbyte called Source.validate_config() but it doesn't run by default. The proposal here is that we enable that _at least_ whenever `check` is run, if not also before a `read()` operation.

Initially we left it out of the critical path because of fears it could introduce false-positive failures. What we are seeing a lot more of though, is users receiving confusing errors from the sources themselves.

False positives are still possible, but it's probably a very low-frequency occurance because the platform already requires all JSON Schema validators to pass before the user can save config in hosted versions of Airbyte.

aaronsteers:
Resolves #126 

When connectors fail due to incorrect config, they often do not do so gracefully.

This update will make pre-validation of config the default behavior, which gives us more control over the user experience.

### DIFF
--- a/airbyte/exceptions.py
+++ b/airbyte/exceptions.py
@@ -262,10 +262,10 @@ class AirbyteConnectorMissingSpecError(AirbyteConnectorError):
 
 
 class AirbyteConnectorValidationFailedError(AirbyteConnectorError):
-    """Connector check failed."""
+    """Connector config validation failed."""
 
     guidance = (
-        "Please double-check your config or review the connector's logs for more information."
+        "Please double-check your config and review the validation errors for more information."
     )
 
 
--- a/airbyte/exceptions.py
+++ b/airbyte/exceptions.py
@@ -261,6 +261,14 @@ class AirbyteConnectorMissingSpecError(AirbyteConnectorError):
     """Connector did not return a spec."""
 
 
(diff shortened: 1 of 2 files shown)


airbyte/exceptions.py
```diff
@@ -261,6 +261,14 @@ class AirbyteConnectorMissingSpecError(AirbyteConnectorError):
     """Connector did not return a spec."""
 
 
+class AirbyteConnectorValidationFailedError(AirbyteConnectorError):
+    """Connector check failed."""
+
+    guidance = (
+        "Please double-check your config or review the connector's logs for more information."
+    )
```
aaronsteers:
```suggestion
    """Connector config validation failed."""

    guidance = (
        "Please double-check your config and review the validation errors for more information."
    )
```

airbyte/sources/base.py
```diff
@@ -555,6 +566,7 @@ def read(
         streams: str | list[str] | None = None,
         write_strategy: str | WriteStrategy = WriteStrategy.AUTO,
         force_full_refresh: bool = False,
+        skip_validation: bool = True,
```
aaronsteers:
```suggestion
        skip_validation: bool = False,
```

bindipankhudi (approved):
Looks great!

PR #136, opened 2024-03-20
Feat: Allow incremental refresh with replace strategy

aaronsteers:
This is discouraged in normal situations, but there are valid use cases.

The use case brought up [in slack](https://airbytehq-team.slack.com/archives/C06FZ238P8W/p1710823746305029?thread_ts=1710320283.398979&cid=C06FZ238P8W) was an SQS queue - wherein we want to process only new records and not already-seen records.

### DIFF
--- a/airbyte/sources/base.py
+++ b/airbyte/sources/base.py
@@ -589,7 +589,6 @@ class Source:
                     "Using `REPLACE` strategy without also setting `full_refresh_mode=True` "
                     "could result in data loss. "
                     "To silence this warning, use the following: "
-                    "`import warnings; "
                     'warnings.filterwarnings("ignore", '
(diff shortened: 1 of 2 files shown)


airbyte/sources/base.py
```diff
@@ -583,12 +584,17 @@ def read(
...
-                    "write_strategy": write_strategy,
-                    "force_full_refresh": force_full_refresh,
-                },
+            warnings.warn(
+                message=(
+                    "Using `REPLACE` strategy without also setting `full_refresh_mode=True` "
+                    "could result in data loss. "
+                    "To silence this warning, use the following: "
+                    "`import warnings; "
+                    'warnings.filterwarnings("ignore", '
```
aaronsteers:
```suggestion
                    'warnings.filterwarnings("ignore", '
```

PR #139, opened 2024-03-23
Chore: Add tests to ensure replace strategy works as expected

aaronsteers:
This aims to assert more confidently that 'replace' strategy works as expected.

So far, tests are still passing after expanding the test criteria.

Here is the code in question, where `replace` write strategy directly maps to a "table swap" operation:

https://github.com/airbytehq/PyAirbyte/blob/76a29044d9fb49d6dd0ba03377fdf7144f6024b6/airbyte/_processors/sql/base.py#L685-L691


bindipankhudi (approved):
lgtm! 👍

aaronsteers:
The new tests actually went stale so I'm closing. We can re-open if this comes up again .

PR #143, opened 2024-03-26
Feat: Add Cloud interop via Airbyte API Integrations

aaronsteers:


aaronsteers:
Quick update. I've added create+delete tests for a source and destination.

Both tests are passing.

https://github.com/airbytehq/PyAirbyte/blob/68e9df1f17f17d0b95c88058a369960446f240d4/tests/integration_tests/test_api_crud.py#L61-L115

airbyte/caches/base.py
```diff
@@ -44,6 +44,11 @@ class CacheBase(BaseModel):
     table_suffix: str = ""
     """A suffix to add to all table names."""
 
+    _deployed_api_root: Optional[str] = PrivateAttr(default=None)
+    _deployed_workspace_id: Optional[str] = PrivateAttr(default=None)
+    _deployed_destination_id: Optional[str] = PrivateAttr(default=None)
+    _deployed_connection_id: Optional[str] = PrivateAttr(default=None)
```
aaronsteers: TODO: Revisit if this is needed. (@aaronsteers)

airbyte/cloud/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+"""
+
+from __future__ import annotations
+
+from airbyte.cloud._workspaces import CloudWorkspace
+
+
+__all__ = [
+    "CloudWorkspace",
+]
```
aaronsteers: There are a few more classes we want to include here. At minimum, we need to add `CloudConnection`...

airbyte/cloud/_destinations.py
```diff
@@ -0,0 +1,66 @@
...
+        DestinationSnowflake,
+    )
+    from sqlalchemy.engine import Engine
+
+    from airbyte.caches.base import CacheBase
+    from airbyte.cloud._workspaces import CloudWorkspace
+
+
+@dataclass
+class CloudDestination:
```
aaronsteers: TODO: Revisit where this class is used, and if we need it.

airbyte/cloud/_workspaces.py
```diff
@@ -0,0 +1,351 @@
...
+
+        return deployed_connection.connection_id
+
+    def delete_connection(
+        self,
+        connection_id: str | None,
+        *,
+        delete_source: bool = False,
+        delete_destination: bool = False,
+    ) -> None:
```
aaronsteers: Might be worth adding a deletion flag to this class's private attributes. Then, other methods which should not operate on a deleted object could fail in a sensible way. (Might be cleanest to implement this with a decorator on those methods.)

airbyte/datasets/_sql.py
```diff
@@ -33,16 +38,35 @@ def __init__(
         cache: CacheBase,
         stream_name: str,
         query_statement: Selectable,
+        stream_configuration: ConfiguredAirbyteStream | None | Literal[False] = None,
     ) -> None:
+        """Initialize the dataset with a cache, stream name, and query statement.
+
```
aaronsteers, line 45:
```suggestion
        """Initialize the dataset with a cache, stream name, and query statement.

        This class is not intended to be created directly. Instead, you can retrieve
        datasets from caches or Cloud connection objects, etc.

```

airbyte/secrets/base.py
```diff
@@ -0,0 +1,142 @@
...
+            return self.name == str(value)
+
+        return super().__eq__(value)
+
+
+class SecretHandle:
+    """A handle for a secret in a secret manager.
+
+    This class is used to store a reference to a secret in a secret manager.
+    The secret is not retrieved until the `get` method is called on the handle.
```
aaronsteers:
```suggestion
    The secret is not retrieved until the `get_value()` method is called on the handle.
```

airbyte/secrets/google_gsm.py
```diff
@@ -0,0 +1,209 @@
...
+        ListSecretsPager,
+    )
+
+
+class GoogleGSMSecretManager(CustomSecretManager):
+    """Secret manager that retrieves secrets from Google Secrets Manager (GSM).
+
+    This secret manager is not enabled by default. To use it, you must provide the project ID and
+    the credentials for a service account with the necessary permissions to access the secrets.
+    """
```
aaronsteers: Todo: make a note about how this implementation has additional capabilities beyond the base class.

airbyte/sources/base.py
```diff
@@ -104,6 +84,11 @@ def __init__(
         if streams is not None:
             self.select_streams(streams)
 
+        self._deployed_api_root: str | None = None
+        self._deployed_workspace_id: str | None = None
+        self._deployed_source_id: str | None = None
+        self._deployed_connection_id: str | None = None
```
aaronsteers: TODO: Consider if this is still needed.

aaronsteers (commented):
My first-pass self-review comments inline.

airbyte/secrets/google_gsm.py
```diff
@@ -0,0 +1,209 @@
...
+        """Get a named secret from Google Colab user secrets."""
+        return SecretString(
+            self.secret_client.access_secret_version(
+                name=f"projects/{self.project}/secrets/{secret_name}/versions/latest"
+            ).payload.data.decode("UTF-8")
+        )
+
+    def fetch_secrets(
+        self,
+        filter_string: str,
```
aaronsteers:
```suggestion
        self,
        *,
        filter_string: str,
```

airbyte/secrets/google_gsm.py
aaronsteers:
```suggestion
class GoogleGSMSecretManager(CustomSecretManager):
    """Secret manager that retrieves secrets from Google Secrets Manager (GSM).

    This class inherits from `CustomSecretManager` and also adds methods 
    that are specific to this implementation: `fetch_secrets()`,
    `fetch_secrets_by_label()` and `fetch_connector_secrets()`.

    This secret manager is not enabled by default. To use it, you must provide the project ID and
    the credentials for a service account with the necessary permissions to access the secrets.
    
    The `fetch_connector_secret()` method assumes a label name of `connector`
    matches the name of the connector (`source-github`, `destination-snowflake`, etc.)
    """
```

tests/integration_tests/conftest.py
aaronsteers, line 1: Note to reader: These fixtures are mostly relocated from the top-level `conftest.py` file, in order to make them sepcific to `integration_tests` tests, and not applied to unit tests which do not need (and should not try to retrieve) creds.

tests/conftest.py
aaronsteers, line 1: Note to reader: fixtures in this file that required credentials have been moved to the `integration_tests/conftest.py` file, one layer down.

pyproject.toml
```diff
@@ -41,6 +41,7 @@ ulid = "^1.1"
 
 # TODO: Remove this arbitrary python constraint once `sqlalchemy-bigquery` has done so.
 sqlalchemy-bigquery = { version = "1.9.0", python = "<3.13" }
+airbyte-api = {git = "https://github.com/airbytehq/airbyte-api-python-sdk.git", rev = "aj/manual_rename_dir"}
```
aaronsteers: TODO (`airbyte-api`):

- If possible, re-release the `airbyte-api` library so we can pin the dependency to a released version.
- Otherwise, pin to a specific commit so the reference does not break post-release of PyAirbyte.

airbyte/cloud/_connections.py
```diff
@@ -0,0 +1,206 @@
...
+        # Get the sync job by ID (lazy loaded)
+        return SyncResult(
+            workspace=self.workspace,
+            connection=self,
+            job_id=job_id,
+        )
+
+    # Deletions
+
+    def delete(
```
bindipankhudi: Are we making deletion too easy? On the UI, when I try to delete a connection, I get the following message:
Deleting a connection cannot be undone without a full re-sync. Note that:
- All past logs and configurations will be deleted
- Updates of new data will stop
- No existing data in the destination will be altered
Type GitHub → Google Sheets below to confirm this action.

aaronsteers: I think I agree...

I though that the same when designing these. I will think a bit more on this... 🤔

If we aren't sure about the API as of now, we could make private members (`_delete()`); we could also try to phase more scarily `_delete_forever()` - or maybe force the user to confirm by passing the connection ID back to itself or turning on a 'mode' that allows destructive actions.

airbyte/cloud/_sync_results.py
```diff
@@ -0,0 +1,228 @@
...
+
+        if latest_status in FAILED_STATUSES:
+            raise AirbyteConnectionSyncError(
+                workspace=self.workspace,
+                connection_id=self.connection.connection_id,
+                job_id=self.job_id,
+                job_status=self._latest_status,
+            )
+
+    def wait_for_completion(
```
bindipankhudi: Could this result in too many calls to our servers? I am guessing not many people are doing this unless they are using Airbyte APIs. It would be good to add some telemetry that indicates these calls are coming from PyAirbyte (which might already be there).

aaronsteers: By default it only pings once every 2 seconds, which is probably more patient than the user would be if they implement their own polling logic. We could expand to 3 but I think I prefer not to poll less than once per 5 seconds...

Wdyt, @bindipankhudi ?

aaronsteers:
@bindipankhudi - FYI, I just cleaned up the docs a bit more and put a link to a recent docs snapshot in the PR description.

I found it very helpful to review the docs and you might find so also (especially the `secrets` and `cloud` submodules:

[generated-docs.zip](https://github.com/airbytehq/PyAirbyte/files/14925281/generated-docs.zip)

airbyte/cloud/_workspaces.py
```diff
@@ -0,0 +1,351 @@
...
+        if isinstance(source, Source):
+            if not source._deployed_source_id:  # noqa: SLF001
+                raise ValueError("Source has not been deployed.")  # noqa: TRY003
+
+            source_id = source._deployed_source_id  # noqa: SLF001
+
+        elif isinstance(source, str):
+            source_id = source
+
+        delete_source(
```
bindipankhudi: i am guessing the logic for making sure source is not part of any active connections exists in the api code.

aaronsteers: @bindipankhudi - Good question! I assumed so, but now I'm not 100% sure... 🤔

airbyte/cloud/_workspaces.py
```diff
@@ -0,0 +1,351 @@
...
+        else:
+            raise exc.PyAirbyteInputError(
+                guidance="You must provide either a destination ID or a cache object."
+            )
+
+        assert source_id is not None
+        assert destination_id is not None
+
+        deployed_connection = create_connection(
+            name="Connection (Deployed by PyAirbyte)",
```
bindipankhudi: Currently the connection names (in cloud UI) look like "Github -> Google Sheets". should we try to stick to the same format, so something like "Github -> Google Sheets (Deployed by PyAirbyte)"

bindipankhudi (approved):
Great work! :) Looks great to me. Left some comments, nothing blocking.

airbyte/cloud/_sync_results.py
bindipankhudi: yep, 2 seconds sounds good to me!

airbyte/cloud/_connections.py
bindipankhudi: yea I like the idea of making it private. We could do the same with the delete_connection method in deploy context?

airbyte/cloud/_workspaces.py
aaronsteers: Good idea! At this point in the code though, I only have the GUID IDs. I do think it's worth a lookup operation though, so the names are readable - and the calls themselves are quick.

bindipankhudi: ah, no worries, we can address this in another iteration.

airbyte/cloud/_connections.py
aaronsteers: @bindipankhudi - Can you take another look at these delete operations?

I have renamed `delete*()` methods to `_permanently_delete*()`. I think adding "permanently_" before the method name instills the necessary sense of dread/caution. I would be okay making them public now - or leaving private.

Similarly for the `deploy*()` methods, I did not yet make them private but I think I will go ahead and do so for now.

aaronsteers: Update: `deploy*()` methods have now all been renamed to `_deploy*()`.

Issue #118, opened 2024-03-11
Ability to add missing columns


bindipankhudi:
> This will also require us to be able to add missing columns to a cache table when those columns don't already exist. 

 _Originally posted by @aaronsteers in [#14](https://github.com/airbytehq/PyAirbyte/issues/14#issuecomment-1989009046)_

aaronsteers:
Added as related to Epic:

- #111

Issue #14, opened 2024-02-07
Align column and table normalization with Dv2 destinations


flash1293:
To allow easy migration from airbyte-lib to hosted Airbyte, the logic to generate table and column names as well as column types need to be identical to the logic used in Java destinations.

This might already be the case in a lot of situations, but we need to make sure edge cases are handled properly, like weird characters in column and table names.
```[tasklist]
### Tasks
- [ ] Confirm the common special chars are handled consistently (['-', '_', '.'])
- [ ] Confirm that casing is handled consistently
- [ ] Confirm other more random special chars are handled consistently (emoji, punctuation, non-latin characters, chars with accents, etc.)
```

----

From @aaronsteers 

## Reference Pages Available

These pages can be used to confirm the Dv2-compatible table schema requirements:

- [Upgrading to Destinations V2 | Airbyte Documentation](https://docs.airbyte.com/release_notes/upgrading_to_destinations_v2)
- [Typing and Deduping | Airbyte Documentation](https://docs.airbyte.com/using-airbyte/core-concepts/typing-deduping)

Quotes:

From [Notion](https://www.notion.so/Airbyte-Data-Warehouse-Destination-V2-Philosophy-e6081edb87084999948acdc8b9607ea1?pvs=4#18fa03d3e9a7416d8049fd9ce7370958):

> ### Notes on name transformations and per-destination rules
> 
> As a rule, we try to keep schema, table, and column names as close as possible to the source, but there are sometimes exceptions:
>
> - Spaces are not allowed and should be replaced with `_`
> - Capital letters *are* maintained in table or column names
> - Some destinations have case-sensitive conventions, e.g. Snowflake prefers UPPERCASE schemas, tables, and columns.  Learn more at [[Database Facts](https://www.notion.so/Database-Facts-4a68926d3acf46d2a65890f60f767472?pvs=21)](https://www.notion.so/Database-Facts-4a68926d3acf46d2a65890f60f767472?pvs=21).

aaronsteers:
This will also require us to be able to add missing columns to a cache table when those columns don't already exist.

PR #144, opened 2024-03-27
Feat: Align internal Airbyte columns with Dv2; Feat: Auto-add missing columns to Cache tables.

aaronsteers:
Resolves: #14 
Resolves: #118 

This update does a couple things:

1. Adds these internal columns to all PyAirbyte datasets and cache tables:
   - `_airbyte_emitted_at` - The timestamp corresponding to `emitted_at` in the Airbyte Record message.~~
   - ~~`_airbyte_loaded_at` - The timestamp representing `utcnow()` at the time the record is received by PyAirbyte.~~ - removed per comments below
   - `_airbyte_meta` - For now, defaults to an empty dictionary.
   - `_airbyte_raw_id` - A unique ID per record, assigned as it is received from source. (Unlike Dv2, there is no raw table but I kept the same column name for consistency.)
2. Auto-adds any missing columns to cache tables if they are declared in the schema. This including the above-named internal columns.

As of this iteration, PyAirbyte will not...

- ...will not attempt to expand or change column types if they are incorrect or stale.
- ~~...will not attempt to align `_airbyte_loaded_at` with the batch load time.~~
  - ~~As noted above, the time is when we first start processing the record. In the future, we can try to set this during load time, but doing so is fragile and may not be possible in a universal/generic manner.~~ (Removed, so no longer relevant.)
- ...will not try to populate `_airbyte_meta` with any data.

## Tests

- [x] I've added an integration test across all 4 SQL platforms (DuckDB, Postgres, Snowflake, BigQuery) which confirms that PyAirbyte can correctly add declared columns when they are detected missing.


airbyte/_util/protocol_util.py
```diff
@@ -3,8 +3,11 @@
 """Internal utility functions, especially for dealing with Airbyte Protocol."""
 from __future__ import annotations
 
+from datetime import datetime
 from typing import TYPE_CHECKING, Any, cast
```
aaronsteers:
```suggestion
from typing import TYPE_CHECKING, Any, cast
from datetime import datetime
```

aaronsteers:
```suggestion
from typing import TYPE_CHECKING, Any, cast
from datetime import datetime
```

aaronsteers:
@jbfbell, @evantahler - Cc'ing you here for visibility.

This PR gets us closer to compatibility with the Dv2 specs, which I've documented / linked to in the related issue:

- #14 

A couple callouts/questions:

1. Should we just forget about `_airbyte_loaded_at`? We don't use a long-running raw table and Dv2 doesn't include it in its final tables. Initially I wanted to include, but I don't know that it adds enough value to warrant a split.
   - I'd love to see Dv3 adding `_airbyte_loaded_at` into the final table also, and (as discussed) consider dropping the long-running raw table, but again, probably not a high enough priority, and its easy for us to just ignore this for the time being.
2. We are lower-casing all columns and table names as of now, but according to Notion, I think you are preserving case for all except Snowflake - is that correct? (Ignoring DuckDB, the others we have covered here are BigQuery and Postgres.)

An important caveat:

- We're not expecting that the user will be able to simply "continue" a sync from PyAirbyte in Airbyte Cloud. If they want to use the exact same table names, this will probably mean a 'reset' after migrating their jobs to Cloud. The important thing is that their _downstream_ processing shouldn't have to change, or shouldn't have to change _much_.

evantahler:
Nice work AJ!  

> Should we just forget about _airbyte_loaded_at? We don't use a long-running raw table and Dv2 doesn't include it in its final tables. Initially I wanted to include, but I don't know that it adds enough value to warrant a split.

We made a funny choice with Dv2 which tried to balance "give the users lots of data about the sync" with "users seem to be mad when we make too many new columns".  That's how we ended up showing `emitted_at` and not `loaded_at` in the final table.  `emitted_at` is more logically useful as the source time matters far more for analysis - probably what you are going to use in your analysis.  Users can still get to `loaded_at` via the join to the raw table, but we figured that was rarer. 

> I'd love to see Dv3 adding _airbyte_loaded_at into the final table also, and (as discussed) consider dropping the long-running raw table, but again, probably not a high enough priority, and its easy for us to just ignore this column for the time being.

So the long-running raw table are getting **_even more_** important now for the refresh work, especially as we are merging records across generations, and to power truncating refreshes without data-towntime.  I think they are with us for the long haul now.  If none of those words made sense (because we made them up recently), check out this [draft of the updated user-facing doc](https://docs.google.com/document/d/18_xUr8lAZexDzR4KJrQJwEknsMlh0ilhrobrXTTQaxM/edit) and join `#1-point-0-refrehes`.  

> We are lower-casing all columns and table names as of now, but according to Notion, I think you are preserving case for all except Snowflake - is that correct? (Ignoring DuckDB, the others we have covered here are BigQuery and Postgres.)

Casse-sensitivity is kind of per-destination now.  I'll let @jbfbell take that one, as I don't have the latest info any more.

aaronsteers:
@evantahler - Thanks for the review and for this context.

I'll drop `_airbyte_loaded_at` from scope. 

This PR doesn't touch capitalization rules so I think we're ok taking any needed changes to those in a subsequent pass.

@jbfbell - Let me know if you have any other questions/concerns here. I think we're good to go with the addition of `_airbyte_extracted_at`, `_airbyte_meta` (always `{}` for now) and `_airbyte_raw_id`.

This PR doesn't dive into what to do with other special characters, but if you can point me at source code or a Notion/Docs page, I'll take on any other name normalization rules in a subsequent PR.

Thanks, both!

aaronsteers:
@jbfbell and @bindipankhudi - If one or both of you is available to review/approve tomorrow, I'll go ahead and merge when ready.

All tests are passing now. ✅ 😄

airbyte/_processors/sql/base.py
```diff
@@ -285,7 +300,10 @@ def _get_table_by_name(
                 message="Cannot force refresh and use shallow query at the same time."
             )
 
-        if force_refresh or table_name not in self._cached_table_definitions:
+        if force_refresh and table_name not in self._cached_table_definitions:
+            self._invalidate_table_cache(table_name)
```
jbfbell: I think I'm missing something here, won't this never do anything? `table_name not in self._cached_table_definitions` is `True` which means inside of `_invalidate_table_cache` then `table_name in self._cached_table_definitions` _must _be__ `False` ?

aaronsteers: Good catch!

Yes, I think it should be reversed. Will flip it.

aaronsteers:
```suggestion
        if force_refresh and table_name in self._cached_table_definitions:
            self._invalidate_table_cache(table_name)
```

jbfbell (approved):

PR #145, opened 2024-03-27
CI: Add 'no-creds' test in CI, fix "only on forks" condition of full pytest suite

aaronsteers:
Add a CI test to ensure that all tests that require creds are correctly marked with the `pytest.mark.requires_creds` decorator. Otherwise, those tests will fail in this new workflow.

### DIFF
--- a/.github/workflows/python_pytest.yml
+++ b/.github/workflows/python_pytest.yml
@@ -73,8 +73,10 @@ jobs:
 
   pytest:
     name: Pytest (All, Python ${{ matrix.python-version }}, ${{ matrix.os }})
-    # Don't run on forks
-    if: github.repository_owner == 'airbytehq'
+    # Don't run on forks. Run on pushes to main, and on PRs that are not from forks.
+    if: >
+      (github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+      (github.event.pull_request.head.repo.fork == false)
     strategy:
       matrix:
         python-version: [
--- a/.github/workflows/python_pytest.yml
+++ b/.github/workflows/python_pytest.yml
@@ -65,7 +65,7 @@ jobs:
       run: poetry install
 
     # Job-specific step(s):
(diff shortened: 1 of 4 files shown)


aaronsteers:
@bindipankhudi - FYI, this is low pri.

But turns out, all our tests appear to be correctly flagged, so there's no work to do to add missing markers.

.github/workflows/python_pytest.yml
```diff
@@ -42,6 +45,32 @@ jobs:
...
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.10'
+        cache: 'poetry'
+    - name: Install dependencies
+      run: poetry install
+
+    # Job-specific step(s):
+    - name: Run Pytest (Fast Tests Only)
```
aaronsteers:
```suggestion
    - name: Run Pytest (No-Creds)
```

.github/workflows/python_pytest.yml
```diff
@@ -1,7 +1,10 @@
 # This workflow will run pytest.
 #
-# There are two jobs which run in parallel:
+# There are three jobs which run in parallel:
 # 1. pytest-fast: Run fast tests only, and fail fast so the dev knows asap if they broke something.
+# 1. pytest-no-creds: Run tests only if they don't require creds. The main use case is to run tests
+#    on forks, where secrets are not available. We flush the GCP_GSM_CREDENTIALS env var to an
+#    invalid value to ensure that tests that require creds are not run.
 # 2. pytest: Run all tests, across multiple python versions.
```
aaronsteers:
```suggestion
# 2. pytest-no-creds: Run tests only if they don't require creds. The main use case is to run tests
#    on forks, where secrets are not available. We flush the GCP_GSM_CREDENTIALS env var to an
#    invalid value to ensure that tests that require creds are not run.
# 3. pytest: Run all tests, across multiple python versions.
```

Issue #160, opened 2024-04-01
🐛 Bug: Columns are incorrectly pruned due to case sensitivity (regression from `v0.8.1`)


aaronsteers:
This appears to have been created by the fix here:

- https://github.com/airbytehq/PyAirbyte/pull/131

which was released in `v0.8.1`. This issue seems to be related to case sensitivity during the "exists" check before pruning unexpected columns.

Raised in Slack: https://airbytehq-team.slack.com/archives/C06FZ238P8W/p1711987104664789

PR #161, opened 2024-04-01
Fix: Resolve issue where case-sensitive column pruning removes real columns

aaronsteers:
Resolves: #160 

- Remove unused/unneeded `protocol_utils`.
- Rename `CaseInsensitiveDict` class to `StreamRecord`.
- Refactor to all code paths use `StreamRecord` class.
- Add top-level column names are normalized before writing to dict.
- Add handling in SQL processors to normalize names to lower-case before importing from files to SQL table.


airbyte/sources/base.py
```diff
@@ -73,7 +72,7 @@ def as_temp_files(files_contents: list[dict | str]) -> Generator[list[str], Any,
     finally:
         for temp_file in temp_files:
             with suppress(Exception):
-                temp_file.unlink()
+                Path(temp_file.name).unlink()
```
aaronsteers, line 75: Drive-by fix. `temp_file` doesn't have `unlink()` method. Was being suppressed.

bindipankhudi (approved):
Looks great!

PR #163, opened 2024-04-02
Chore: Update user-facing header to be more inclusive in `analytics.yml`

colesnodgrass:
As this file is going to be used by more than simply `PyAirbyte`, replace the `PyAirbyte` reference with `Airbyte`.


aaronsteers (approved):
Nice! 👍

PR #166, opened 2024-04-03
Docs: Move `StreamRecord` class to public module, add docs

aaronsteers:
Following from:

- #14 

This PR simply moves the class to a public module and adds user-facing reference documentation.


bindipankhudi (approved):
Looks good!

PR #169, opened 2024-04-03
CI: Skip lint checks in pytest when redundant

aaronsteers:
We have lint checks in pytest so that developers can see in one place whether their code is ready to push.

However, in CI, we have a dedicated lint job and it is redundant to show these results also in the pytest jobs.

If there is a lint/format problem, but if all other tests are successful, this change will show pytest tests as 'green' and only the lint/format jobs will show as red. This helps the developer know what is needed at-a-glance, without having to jump into pytest logs when all other tests are fine.


bindipankhudi (approved):
Looks great!

Issue #167, opened 2024-04-03
Feature Request - make cache compatible with RAW Json Airbyte format


maver1ck:
Hi,
Current format of Cache is in line with Airbyte normalized v2 format.
This has some drawback like lack of compatibility between PyAirbyte and Airbyte.
Also schema evolution is not supported.

Adding Airbyte raw json format support may fix those issues.

aaronsteers:
@maver1ck - Thanks for logging this suggestion. 🙏 

We're actually launching an update today or tomorrow which will add a couple new features:

1. Missing columns will be auto-added if they are detected missing. (Very basic schema evolution.)
2. PyAirbyte will add support for the following 3 `_airbyte_*` columns: `_airbyte_raw_id`, `_airbyte_extracted_at`, and `_airbyte_meta`. This will make PyAirbyte closer to the Dv2 table conventions.

This is also in the context of all Airbyte destinations currently on their way to the "v2" standard.

What do you think of this path forward? Any concerns?

aaronsteers:
Decided to expand our docs to better cover this topic:

- https://airbytehq.github.io/PyAirbyte/airbyte/records.html

aaronsteers:
v0.9.0 is now live 🚀 : https://github.com/airbytehq/PyAirbyte/releases/tag/v0.9.0

maver1ck:
@aaronsteers 
It may help but I really want to get access to `_airbyte_data﻿` column as we currently are doing normalization by ourselves using raw data.

aaronsteers:
@maver1ck - Which destination connector(s) are you using today in Airbyte Cloud/OSS? And do you have the ability to run docker in your python runtimes?

We have a potential path forward, which would be to add support for destinations:

- https://github.com/airbytehq/PyAirbyte/issues/197

Combined with a community contribution WIP:

- #209

aaronsteers:
This came up in slack:

- https://airbytehq-team.slack.com/archives/C06FZ238P8W/p1715112018711929?thread_ts=1714742768.802949&cid=C06FZ238P8W

And I raised in that slack thread some inherent limitations with the raw json format:

> There are some other data engineering challenges that this raises:
> 1. Primary keys and cursor keys aren't indexible or optimized for querying and partition elimination.
> 1. Because of #1, deduping (merge upserts) is impossible or least less efficient.
> 1. Compression is worse because columns aren't typed, leading to more storage usage and worse read/write performance overall.
>

[...]

PR #170, opened 2024-04-03
Docs: Add schema evolution topic

aaronsteers:
Schema evolution was raised by a community member in

- #167 

This docs update explains a bit more about how schema evolution occurs in PyAirbyte.

### DIFF
--- a/airbyte/records.py
+++ b/airbyte/records.py
@@ -51,9 +51,20 @@ PyAirbyte adds the following columns to every record:
 The names of these columns are included in the `airbyte.constants` module for programmatic
 reference.
 
-## Module Overview
+## Schema Evolution
 
-This module contains the StreamRecord class, which is a case-aware, case-insensitive dictionary.
+PyAirbyte supports a very basic form of schema evolution:
+
+1. Columns are always auto-added to cache tables whenever a newly arriving properties are detected
+   as not present in the cache table.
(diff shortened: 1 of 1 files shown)


airbyte/records.py
```diff
@@ -51,9 +51,20 @@
 The names of these columns are included in the `airbyte.constants` module for programmatic
 reference.
 
-## Module Overview
+## Schema Evolution
 
-This module contains the StreamRecord class, which is a case-aware, case-insensitive dictionary.
+PyAirbyte supports a very basic form of schema evolution:
+
+1. Columns are always auto-added to cache tables whenever a newly arriving properties are detected
```
aaronsteers:
```suggestion
1. Columns are always auto-added to cache tables whenever newly arriving properties are detected
```

airbyte/records.py
```diff
@@ -51,9 +51,20 @@
...
+   as not present in the cache table.
+2. Column types will not be modified or expanded to fit changed types in the source catalog.
+   - If column types change, we recommend user to manually alter the column types.
+3. At any time, users can run a full sync with a `WriteStategy` of 'replace'. This will create a
+   fresh table from scratch and thenswap the old and new tables after table sync is complete.
```
aaronsteers:
```suggestion
   fresh table from scratch and then swap the old and new tables after table sync is complete.
```

PR #176, opened 2024-04-07
Chore(tools): Align Pyright configuration with MyPy and specific deterministic venv location for Pyright

natikgadzhi:
## What

The only meaningful change is adding `tool.pyright` config in `pyproject.toml`, so that PyRight can see the `.venv` and resolve imports.

Side-effect: formatted pyproject.toml. 

I don't have a _very_ strong opinion. This makes _my_ life easier as Zed doesn't do mypy. But, mypy and Pyright are not quite the same and have slighly different rules, and using PyRight should not be endorsed, and/or Pyright correctness is not guaranteed as far as I understand.

But, I don't think this will hurt us in any way. ;)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ name = "airbyte"
 description = "PyAirbyte"
 authors = ["Airbyte <contact@airbyte.io>"]
 readme = "README.md"
-packages = [{include = "airbyte"}]
+packages = [{ include = "airbyte" }]
 
 # This project uses dynamic versioning
 # https://github.com/mtkennerly/poetry-dynamic-versioning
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -254,11 +254,13 @@ show_error_codes = true
...
-    "airbyte_protocol.models"
-]
-ignore_missing_imports = true  # No stubs yet (😢)
+module = ["airbyte_protocol", "airbyte_protocol.models"]
+ignore_missing_imports = true                            # No stubs yet (😢)
+
+[tool.pyright]
+pythonVersion = "3.9"
+venvPath = "./"       # Assuming .venv is at the root of your project
+venv = ".venv"
```
aaronsteers, line 263: Looks good!

FYI, I have started adding https://github.com/airbytehq/PyAirbyte/blob/main/poetry.toml to my projects (including this one), as it ensures a deterministic venv directory. So, this is safe for Pyright and also other processes or documentation that requires locating the venv.

aaronsteers (approved):
Looks great!

@natikgadzhi - For my info, what is the auto-formatter used? Is this a Zed built-in? I haven't started auto-formatting toml files but I've been thinking about it...

natikgadzhi:
I thought poetry should be able to pick up that setting from
pyproject.toml, but I remember it was a PITA.

PR #178, opened 2024-04-10
Fix: Add `json` fall back for deeply nested records

SebastienN15:
`orjson.dumps(some_dict)` raises with `TypeError: Recursion limit reached` a when dumping a heavily nested dict
This PR adds a fall back to use `json.dumps(some_dict)` instead, that can support much more nested dicts.


airbyte/_processors/file/jsonl.py
```diff
@@ -37,4 +38,12 @@ def _write_record_dict(
         record_dict: StreamRecord,
         open_file_writer: gzip.GzipFile | IO[bytes],
     ) -> None:
-        open_file_writer.write(orjson.dumps(record_dict) + b"\n")
+        # If the record is too nested, orjson will fail with error `TypeError: Recursion limit reached`
+        # If such thing happens, we fall back to the slower but more robust json.dumps
```
aaronsteers:
```suggestion
        # If the record is too nested, `orjson` will fail with error `TypeError: Recursion 
        # limit reached`. If so, we fall back to the slower but more 
        # robust `json.dumps`.
```

airbyte/_processors/file/jsonl.py
```diff
@@ -37,4 +38,13 @@ def _write_record_dict(
         record_dict: StreamRecord,
         open_file_writer: gzip.GzipFile | IO[bytes],
     ) -> None:
-        open_file_writer.write(orjson.dumps(record_dict) + b"\n")
+        # If the record is too nested, `orjson` will fail with error `TypeError: Recursion
+        # limit reached`. If so, we fall back to the slower but more
+        # robust `json.dumps`.
```
aaronsteers:
```suggestion
        # limit reached`. If so, fall back to the slower `json.dumps`.
```

aaronsteers:
@SebastienN15 - We have tests and lint checks passing. In the long run, I'd like to add unit tests and logging, but as-is, this seems ready to go, no reason not to merge. :shipit: 

A great addition, and much appreciated.

Thanks again for contributing! 🙏

PR #191, opened 2024-04-12
Feat: Add singular `fetch_connector_secret()` for `GoogleGSMSecretManager`, fix GSM edge cases, and add CI tests

aaronsteers:
We were lacking integration tests for the GSM-based secret fetching (by label and connector).

This PR adds needed integration tests and fixes bugs that otherwise prevent successful retrieval.

This also adds a singular `fetch_connector_secret()` which just gives the first secret that would be returned by `fetch_connector_secrets()`.

### DIFF
--- a/airbyte/secrets/google_gsm.py
+++ b/airbyte/secrets/google_gsm.py
@@ -31,11 +31,13 @@ read_result = source.read()
 
 from __future__ import annotations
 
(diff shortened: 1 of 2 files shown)


bindipankhudi (approved):
Nice! :)

PR #192, opened 2024-04-13
Do not merge: Release experimental introp methods to stable

aaronsteers:


airbyte/cloud/connections.py
```diff
@@ -199,16 +199,16 @@ def _permanently_delete(
...
-            self.workspace._permanently_delete_source(  # noqa: SLF001  # Non-public API (for now)
+            self.workspace.permanently_delete_source(  # Non-public API (for now)
                 source=self.source_id
             )
 
         if delete_destination:
-            self.workspace._permanently_delete_destination(  # noqa: SLF001  # Non-public API
+            self.workspace.permanently_delete_destination(  # Non-public API
                 destination=self.destination_id,
             )
```
aaronsteers:
```suggestion
            self.workspace.permanently_delete_destination(  # Non-public API
                destination=self.destination_id,)
```

tests/integration_tests/cloud/test_cloud_workspaces.py
```diff
@@ -15,16 +18,57 @@ def test_deploy_source(
...
+    source_connector: cloud.CloudConnector = cloud_workspace.deploy_source(
+        source=local_source,
+        name_key="My Faker Source",  # Used in deduplication and idempotency
+        update_existing=True,        # Update existing source
+    )
+
+    # Partially update the configuration:
+    source_connector.update_configuration(
+        {"count": 300},
+        merge=True,
```
alex-gron, line 61: @aaronsteers  How do you envision `merge` being different from `update_existing`? 
In my mind, those sound like the same thing. Can we use `update_existing` in both places?

aaronsteers: Hi, @alex-gron. The "update_existing" was/is planned for a "deploy*()" verb - where a `False` value would fail if the resource already exists - rather than update it.

In this context "merge" is in the context of changes to settings - where 'merge=true' means we want to merge the provided setting with existing settign, and 'merge=false' would mean we replace all non-provided keys with the provided ones. This is useful when you want to change only one setting without modifying others. Especially useful when we don't want to re-supply creds each time we update non-secret config values.

Issue #177, opened 2024-04-10
🐛Bug: Duck DB Cache has Problems converting Big Numbers to Decimal


tinomerl:
# Problem Description

PyAirbyte Version: v0.9.0
Source Version: source-facebook-marketing 2.1.1

When trying to get Data from the Facebook Marketing Source, to be precise the "ad_account" Stream I get the following Exception.

> duckdb.duckdb.InvalidInputException: Invalid Input Error: JSON transform error in file ".cache/default_cache/ad_account_01HV39XAFK4DART550Y0KCFWS2.jsonl.gz", in line 1: Failed to cast value to decimal: 1234567890123456.0
Try setting 'auto_detect' to true, specifying 'format' or 'records' manually, or setting 'ignore_errors' to true.

(The value in the error is redacted since it's a live ID. But the number of digits is the same.)

The column which leads to the error is no particular clumn. I had the error with the following Columns

- end_advertiser
- owner
- funding_source

Interestingly the column `id` holds the same value as eg. end_advertiser minus the decimal. Making it `1234567890123456`. This column doesn't throw an Exception. 

**Additional Info**
I tested the same stream with a PostgreSQL cache and it worked just fine. The column types there are numeric.

# Sample Code

```python
import airbyte as ab

source: ab.Source = ab.get_source("source-facebook-marketing")
source.set_config({
    "access_token": "redacted",
    "start_date": "2024-03-28T00:00:00Z",
    "account_ids": ["redacted"]
})
source.get_available_streams()
source.select_streams("ad_account")
records = source.read()
```

# Error Log

<details><summary>Stack trace</summary>
<p>

```
Started `source-facebook-marketing` read operation at 08:07:42...
Failed `source-facebook-marketing` read operation at 08:08:14.
Traceback (most recent call last):
  File "/home/tino/repos/PyAirbyte/airbyte/sources/base.py", line 740, in read
    cache.processor.process_airbyte_messages(
  File "/home/tino/repos/PyAirbyte/airbyte/_processors/base.py", line 200, in process_airbyte_messages
    self.write_all_stream_data(
  File "/home/tino/repos/PyAirbyte/airbyte/_processors/base.py", line 211, in write_all_stream_data
    self.write_stream_data(stream_name, write_strategy=write_strategy)
  File "/home/tino/repos/PyAirbyte/airbyte/_processors/sql/base.py", line 555, in write_stream_data
    temp_table_name = self._write_files_to_new_table(
[...]

PR #194, opened 2024-04-13
Fix: Expand default decimal size to `(38, 9)` to resolve issues with large number values

aaronsteers:
Resolves: #177

- #177

~~Note: The choice of `DECIMAL(38, 10)` is somewhat arbitrary, but it forces use of a large internal (int) data type in DuckDB, and it should be big enough (28 digits) on the left side of the decimal while also providing a meaningful amount of precision on the right side (10 digits).~~

After some digging, I found that Java destinations prefer scale of `9`: `DECIMAL(38, 9)`.

I've updated the PR to use this size.

----

Per internal slack inquiry: https://airbytehq-team.slack.com/archives/C03C4AVJWG4/p1713209219674449?thread_ts=1713208569.977779&cid=C03C4AVJWG4

Relevant code on the Java side:

https://github.com/airbytehq/airbyte/blob/d8e8a0dab4d951fb39ab2fb7048d167f096948c2/airbyte-cdk/java/airbyte-cdk/db-destinations/src/main/kotlin/io/airbyte/cdk/integrations/destination/jdbc/typing_deduping/JdbcSqlGenerator.kt#L80

### DIFF
--- a/airbyte/types.py
+++ b/airbyte/types.py
@@ -13,7 +13,7 @@ from rich import print
 CONVERSION_MAP = {
     "string": sqlalchemy.types.VARCHAR,
     "integer": sqlalchemy.types.BIGINT,
-    "number": sqlalchemy.types.DECIMAL(38, 10),
+    "number": sqlalchemy.types.DECIMAL(38, 9),
     "boolean": sqlalchemy.types.BOOLEAN,
     "date": sqlalchemy.types.DATE,
     "timestamp_with_timezone": sqlalchemy.types.TIMESTAMP,
--- a/airbyte/types.py
+++ b/airbyte/types.py
@@ -13,7 +13,7 @@ from rich import print
 CONVERSION_MAP = {
     "string": sqlalchemy.types.VARCHAR,
     "integer": sqlalchemy.types.BIGINT,
-    "number": sqlalchemy.types.DECIMAL,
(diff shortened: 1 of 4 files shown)


airbyte/types.py
```diff
@@ -13,7 +13,7 @@
 CONVERSION_MAP = {
     "string": sqlalchemy.types.VARCHAR,
     "integer": sqlalchemy.types.BIGINT,
-    "number": sqlalchemy.types.DECIMAL,
+    "number": sqlalchemy.types.DECIMAL(38, 10),
```
aaronsteers: Aligning with Airbyte destinations default:

```suggestion
    "number": sqlalchemy.types.DECIMAL(38, 9),
```

Per internal slack inquiry: https://airbytehq-team.slack.com/archives/C03C4AVJWG4/p1713209219674449?thread_ts=1713208569.977779&cid=C03C4AVJWG4

Relevant code on the Java side:

https://github.com/airbytehq/airbyte/blob/d8e8a0dab4d951fb39ab2fb7048d167f096948c2/airbyte-cdk/java/airbyte-cdk/db-destinations/src/main/kotlin/io/airbyte/cdk/integrations/destination/jdbc/typing_deduping/JdbcSqlGenerator.kt#L80

Issue #201, opened 2024-04-22
PyAirbyte: Add new SnowflakeCortex processer which can take catalog as an argument


bindipankhudi:
PR in progress - https://github.com/airbytehq/PyAirbyte/pull/203

PR #203, opened 2024-04-23
Feat: Add Cortex-compatible Snowflake SQL processor for storing vector data; Enable native merge upsert for Snowflake caches

bindipankhudi:
Related ticket: https://github.com/airbytehq/PyAirbyte/issues/201

Things are working end to end. Test code is in `run_snowflake_cortex_test_data.py`

to-dos:
* Add integration tests (destination has enough tests) - also we will CortexProcessor into Destination once we refactor this code to use new catalog manager (third point here) 
* Add logging (not a blocker)
* Refactor catalog manager to not create state tables (this is being worked on in another PR)


airbyte/_processors/base.py
```diff
@@ -152,7 +152,6 @@ def process_record_message(
         the file processor.
         """
 
-    @final
```
bindipankhudi, line 155: overriding this in snowflakecortex processor, hence removing @final

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,257 @@
...
+
+    @overrides
+    def to_sql_type(
+        self,
+        json_schema_property_def: dict[str, str | dict | list],
+    ) -> sqlalchemy.types.TypeEngine:
+        """Convert a value to a SQL type."""
+        sql_type = super().to_sql_type(json_schema_property_def)
+        if isinstance(sql_type, sqlalchemy.types.ARRAY):
+            return f"Vector(Float, {self.vector_length})"
```
bindipankhudi: Returning custom type Vector(Float, xx), since SQLAlchemy does not have support for Snowflake vector yet.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,257 @@
...
+
+        return sql_type
+
+
+class SnowflakeCortexSqlProcessor(SnowflakeSqlProcessor):
+    """A Snowflake implementation for use with Cortex functions."""
+
+    type_converter_class: SnowflakeCortexTypeConverter
+
+    def __init__(
```
bindipankhudi, line 63: overriding the constructor since we want to do a bunch of custom stuff - initialize source, catalog manager, typecoverter, etc

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,257 @@
...
+        stream_schemas: dict[str, dict] = {}
+
+        # Process messages, writing to batches as we go
+        for message in messages:
+            if message.type is Type.RECORD:
+                record_msg = cast(AirbyteRecordMessage, message.record)
+                stream_name = record_msg.stream
+
+                if stream_name not in stream_schemas:
+                    stream_schemas[stream_name] = self._get_json_schema_from_catalog(
```
bindipankhudi: the base class is calling self.cache.processor.get_json_schema, which is null in this case, so created a local method.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,257 @@
...
+        self._connection_to_reuse: Connection | None = None
+        # call base call to do necessary initialization
+        RecordProcessor.__init__(self, cache=cache, catalog_manager=None)
+        self._ensure_schema_exists()
+        self.file_writer = file_writer or self.file_writer_class(cache)
+        self.type_converter = SnowflakeCortexTypeConverter(vector_length=vector_length)
+        self._cached_table_definitions: dict[str, sqlalchemy.Table] = {}
+
+    @overrides
+    def process_airbyte_messages(
```
bindipankhudi, line 89: most of the implementation in this method is same as base class, except for the two cases called out.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,257 @@
...
+                    stream_name = "myteststream"  # stream_state.stream_descriptor.name
+                    self._pending_state_messages[stream_name].append(state_msg)
+
+            else:
+                # Ignore unexpected or unhandled message types:
+                # Type.LOG, Type.TRACE, Type.CONTROL, etc.
+                pass
+
+        # in base class, following reads from expected streams, hence overriding
+        for stream_name in stream_schemas:
```
bindipankhudi: this relies on self.expected_streams in base class, hence calling the  write_stream_data method directly

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,258 @@
...
+    ) -> bool:
+        # to-do: SQLAlchemy is not compatible with vector type. Implemenbt using Snowflake python connector
+        return True
+
+    def _finalize_state_messages(
+        self,
+        stream_name: str,
+        state_messages: list[AirbyteStateMessage],
+    ) -> None:
+        # to-do: need to override since we are not using catalog manager. Do we want to write state messages in this scenario ?
```
bindipankhudi: State tables are not being written right now.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,258 @@
...
+        print(put_files_statements)
+        self._execute_sql(put_files_statements)
+        columns_list = [
+            self._quote_identifier(c)
+            for c in list(self._get_sql_column_definitions(stream_name).keys())
+        ]
+        files_list = ", ".join([f"'{f.name}'" for f in files])
+        columns_list_str: str = indent("\n, ".join(columns_list), " " * 12)
+
+        # to-do: move following into a separate method, then don't need to override the whole method.
```
bindipankhudi: We need to read embedding data suffixed by ::Vector(float, xx) for it to be written like that to Snowflake. This is the only change in this method.

airbyte/types.py
```diff
@@ -24,6 +24,7 @@
     # We include them here for completeness.
     "object": sqlalchemy.types.JSON,
     "array": sqlalchemy.types.JSON,
+    "vector_array": sqlalchemy.types.ARRAY,
```
bindipankhudi, line 28: vector_array is converted to SQLAlchemy Array, which gets converted into Vector data type by TypeConverter.

examples/run_snowflake_cortex_test_data.py
```diff
@@ -0,0 +1,114 @@
...
+        name="myteststream", json_schema=stream_schema, supported_sync_modes=[SyncMode.incremental, SyncMode.full_refresh]
+    ),
+    primary_key=[["int_col"]],
+    sync_mode=SyncMode.incremental,
+    destination_sync_mode=DestinationSyncMode.overwrite,
+)
+catalog = ConfiguredAirbyteCatalog(streams=[overwrite_stream])
+
+
+# create test messages
```
bindipankhudi: Here str_col & int_col are original columns. The destination will add three additional columns to the catalog and to the data: "page_content", "metadata" and "embedding" all of which we get from the vector_db_based chunking & embedding logic. PyAirbyte code does not enforce presence of any columns though.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,258 @@
...
+                )
+
+            elif message.type is Type.STATE:
+                state_msg = cast(AirbyteStateMessage, message.state)
+                if state_msg.type in [AirbyteStateType.GLOBAL, AirbyteStateType.LEGACY]:
+                    self._pending_state_messages[f"_{state_msg.type}"].append(state_msg)
+                else:
+                    stream_state = cast(AirbyteStreamState, state_msg.stream)
+                    # to-do: remove hardcoding below
+                    stream_name = "myteststream"  # stream_state.stream_descriptor.name
```
bindipankhudi: I couldn't quite figure out how to initialize the state message with a stream. This is the only change in this method. once hardcoding is fixed, can remove this method from here.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,291 @@
...
+    def _ensure_compatible_table_schema(
+        self,
+        stream_name: str,
+        *,
+        raise_on_error: bool = True,
+    ) -> bool:
+        """Read the exsting table schema using Snowflake python connector"""
+        json_schema = self.get_stream_json_schema(stream_name)
+        stream_column_names: list[str] = json_schema["properties"].keys()
+        table_column_names: list[str] = self._get_column_list_from_table(stream_name)
```
bindipankhudi, line 129: get_column_list_from_table uses a local Snowflake Python connector based implementation. SQLAlchemy does not recognize vector type, hence fetching scheme (with embeddings) fails.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,258 @@
...
+        self._execute_sql(put_files_statements)
+        columns_list = [
+            self._quote_identifier(c)
+            for c in list(self._get_sql_column_definitions(stream_name).keys())
+        ]
+        files_list = ", ".join([f"'{f.name}'" for f in files])
+        columns_list_str: str = indent("\n, ".join(columns_list), " " * 12)
+
+        # to-do: move following into a separate method,
+        # then don't need to override the whole method.
```
bindipankhudi: this is the only piece that's overridden in this method.

airbyte/caches/base.py
```diff
@@ -74,6 +74,11 @@ def get_database_name(self) -> str:
         """Return the name of the database."""
         ...
 
+    def get_database_connection_via_alternate_method(self) -> object:
```
bindipankhudi: i could not think of a better way to do this :/ we need to use Snowflake Python connector for checking destination schema.

airbyte/_processors/sql/snowflakecortex.py
aaronsteers: Nice. 👍

airbyte/caches/base.py
aaronsteers: Maybe just a rename of the method, perhaps something like 

`get_vendor_client()` or `get_custom_client()`?

airbyte/_processors/sql/snowflakecortex.py
aaronsteers: Makes sense. When we're ready to merge, I can optionally do another pass to see if there's any way to reduce duplication (maybe call `super()`?). But as it is, I think this is okay.

aaronsteers (commented):
Nice! I don't see any fatal flaws or design issues.

A couple things we can think on here:

1. Interesting to see we didn't need to modify the `SqlProcessorBase` implementation. This somewhat confirms that we can and probably should consider moving the `SqlProcessorBase` upstream into the CDK itself. (Not today, but eventually.)
2. We talked about the cumbersome design of the catalog manager. I'm curious if you've worked around that, or if you think we need to refactor at all before merging your work here.
3. The type handling looks pretty clean. 👍 I understand the temporary workarounds there.
4. The tests look good! I was happy to see those included.

airbyte/_processors/sql/snowflakecortex.py
bindipankhudi: super() does a few things we actually want to override, that's why doing the entire initialization here.

airbyte/caches/base.py
bindipankhudi: Updated name.

pyproject.toml
```diff
@@ -15,7 +15,7 @@ enable = true
 [tool.poetry.dependencies]
 python = ">=3.9,<4.0"
 
-airbyte-cdk = "^0.73.0"
+airbyte-cdk = "^0.81.6"
```
bindipankhudi, line 18: This needs to be updated to match the version used in snowflake-cortex destination, ow the older version of langchain gets used, which creates a bunch of library conflicts.

airbyte/_processors/sql/snowflake.py
```diff
@@ -51,6 +51,7 @@ class SnowflakeSqlProcessor(SqlProcessorBase):
 
     file_writer_class = JsonlWriter
     type_converter_class = SnowflakeTypeConverter
+    supports_merge_insert = True
```
bindipankhudi, line 54: @aaronsteers fyi - i updated this.

tests/integration_tests/test_source_faker_integration.py
```diff
@@ -315,3 +315,36 @@ def test_example_config_file(source_faker_seed_a: ab.Source) -> None:
             output_file=temp.name,
         )
         assert Path(temp.name).exists()
+
+
+def test_merge_insert_not_supported_for_duckdb(
```
bindipankhudi, line 320: for some reason, i was not able to use a CacheBase fixture comprising all cache types, so wrote separate tests.

airbyte/_processors/sql/snowflakecortex.py
aaronsteers: @bindipankhudi - since not all array inputs would be `VECTOR` type (in theory, at least), can we add a TODO comment here for future action?

```suggestion
            # SQLAlchemy doesn't yet support the `VECTOR` data type.
            # We may want to remove this or update once this resolves:
            # https://github.com/snowflakedb/snowflake-sqlalchemy/issues/499
            return f"VECTOR(FLOAT, {self.vector_length})"
```

(Also, more of a nit, but it is common to put types in all-caps so I applied that above also in my suggestion.)

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -0,0 +1,230 @@
...
+        )
+        self.file_writer = file_writer or self.file_writer_class(cache)
+        self.type_converter = SnowflakeCortexTypeConverter(vector_length=vector_length)
+        self._cached_table_definitions: dict[str, sqlalchemy.Table] = {}
+
+    def _get_column_list_from_table(
+        self,
+        table_name: str,
+    ) -> list[str]:
+        """Get column names for passed stream."""
```
aaronsteers: Suggesting adding context for the reader...

```suggestion
        """Get column names for passed stream.
        
        This is overridden due to lack of SQLAlchemy compatibility for the
        `VECTOR` data type. 
        """
```

tests/integration_tests/test_source_faker_integration.py
aaronsteers: Okay. I could probably help integrate with the generic tests, but what you have here seems sufficient so I don't think we need to prioritize it.

aaronsteers (approved):
Just a few small suggestions, then I think this is ready to go! 🚀

PR #205, opened 2024-04-24
Chore: Update ruff rules and fix exceptions

aaronsteers:
The interesting changed files here are:

- `pyproject.toml` - enable 'preview' mode and update lint rules
- `tests/pyproject.toml` lint exceptions/overrides for the `tests` folder

The rest of the files modified are just applying the new lint rules.

The big impactful points are:

1. We enable all the 'preview' rules that previously were skipped, even if selected.
2. We use the new style pyproject declarations `tools.ruff.lint...` and this silences prior warnings.
3. One of the newly enabled checks is the check for copyright: https://docs.astral.sh/ruff/rules/missing-copyright-notice/ (This was previously not check without `airbyte-ci`.)
4. We re-enable linting on the 'tests' folder.
5. We add a few choice exceptions so that linting on tests isn't abnoxious.


aaronsteers:
@bindipankhudi - I'm going to go ahead and merge this so it doesn't create long-term merge conflicts. Should be harmless. Just internal rules updates and correlated fixes (most are auto-fixes).

PR #209, opened 2024-04-26
Feat: Add an Executor for Docker sources

Tarekk:
## What
Adding a `DockerExecutor` class to the `get_source` factory. 

## Why
I absolutely love this library as it allows me to rapidly pull data from a variety of different sources and load them into a "local DW" in duckDB for quick analysis using SQL. All this without having to deploy or startup any servers!! However, on numerous occasions I needed to pull in data from Java based Airbyte sources such as PostgreSQL and MySQL and couldn't use PyAirbyte to do that. Therefore, it would be great if the library supported the execution of Docker images that adhere to the Airbyte protocol. This will greatly increase the number of compatible sources and allow users to develop additional sources in their language of choice.

## How
`DockerExecutor` is almost identical to `PathExecutor` but instead of receiving a path to the executable it expects a list of docker commands. These commands are passed down from the factory and include the mounting of the the `temp_dir` used for materializing the JSON config files.

```python
if docker_executable:
    version = version or "latest"

    temp_dir = tempfile.gettempdir()
    return Source(
        name=name,
        config=config,
        streams=streams,
        executor=DockerExecutor(
            name=name,
            executable=[
                "docker",
                "run",
                "--rm",
                "-i",
                "--volume",
                f"{temp_dir}:{temp_dir}",
                f"airbyte/{name}:{version}",
            ],
        ),
    )

``` 

**Example Usage**
```python
source = ab.get_source(
    "source-faker",
    docker_executable=True,
    config={
        "count": 10,
    },
)
```

Tested via `test_docker_executable.py` which is a replica of `test_source_faker_integration.py` with ```docker_executable=True```

P.S. This is my first ever open source contribution 😃

### DIFF
--- a/airbyte/_executor.py
+++ b/airbyte/_executor.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 import shlex
+import shutil
 import subprocess
 import sys
 from abc import ABC, abstractmethod
@@ -449,4 +450,56 @@ class PathExecutor(Executor):
 
     def execute(self, args: list[str]) -> Iterator[str]:
         with _stream_from_subprocess([str(self.path), *args]) as stream:
+            yield from stream
+
+
+class DockerExecutor(Executor):
+    def __init__(
+        self,
+        name: str | None = None,
+        *,
+        executable: list[str],
+        target_version: str | None = None,
+    ) -> None:
+        self.executable: list[str] = executable
+        name = name or executable[0]
+        super().__init__(name=name, target_version=target_version)
+
+    def ensure_installation(
+        self,
+        *,
+        auto_fix: bool = True,
+    ) -> None:
+        """Ensure that the connector executable can be found.
+
+        The auto_fix parameter is ignored for this executor type.
+        """
+        _ = auto_fix
+        try:
+            assert (
+                shutil.which("docker") is not None
+            ), "Docker couldn't be found on your system. Please Install it."
+            self.execute(["spec"])
+        except Exception as e:
+            # TODO: Improve error handling. We should try to distinguish between
+            #       a connector that is not installed and a connector that is not
+            #       working properly.
+            raise exc.AirbyteConnectorExecutableNotFoundError(
+                connector_name=self.name,
+            ) from e
+
+    def install(self) -> NoReturn:
+        raise exc.AirbyteConnectorInstallationError(
+            message="Connector cannot be installed because it is not managed by PyAirbyte.",
+            connector_name=self.name,
+        )
+
+    def uninstall(self) -> NoReturn:
+        raise exc.AirbyteConnectorInstallationError(
+            message="Connector cannot be uninstalled because it is not managed by PyAirbyte.",
+            connector_name=self.name,
+        )
+
+    def execute(self, args: list[str]) -> Iterator[str]:
+        with _stream_from_subprocess([*self.executable, *args]) as stream:
             yield from stream
--- a/airbyte/sources/util.py
+++ b/airbyte/sources/util.py
@@ -5,12 +5,13 @@ from __future__ import annotations
 
 import shutil
 import sys
+import tempfile
 import warnings
 from pathlib import Path
 from typing import Any
 
 from airbyte import exceptions as exc
-from airbyte._executor import PathExecutor, VenvExecutor
+from airbyte._executor import DockerExecutor, PathExecutor, VenvExecutor
 from airbyte._util.telemetry import EventState, log_install_state
 from airbyte.sources.base import Source
 from airbyte.sources.registry import ConnectorMetadata, get_connector_metadata
@@ -50,6 +51,7 @@ def get_source(
     version: str | None = None,
     pip_url: str | None = None,
     local_executable: Path | str | None = None,
+    docker_executable: bool = False,
     install_if_missing: bool = True,
 ) -> Source:
     """Get a connector by name and version.
@@ -112,6 +114,28 @@ def get_source(
             executor=PathExecutor(
                 name=name,
                 path=local_executable,
+            ),
+        )
+
+    if docker_executable:
+        version = version or "latest"
+
+        temp_dir = tempfile.gettempdir()
+        return Source(
+            name=name,
+            config=config,
+            streams=streams,
+            executor=DockerExecutor(
+                name=name,
+                executable=[
+                    "docker",
+                    "run",
+                    "--rm",
+                    "-i",
+                    "--volume",
+                    f"{temp_dir}:{temp_dir}",
+                    f"airbyte/{name}:{version}",
+                ],
             ),
         )
 
--- a/tests/integration_tests/test_docker_executable.py
+++ b/tests/integration_tests/test_docker_executable.py
@@ -1,0 +1,297 @@
+# Copyright (c) 2023 Airbyte, Inc., all rights reserved.
+
+"""Integration tests which leverage the source-faker connector to test the framework end-to-end with the Docker Executor."""
+
+from __future__ import annotations
+
+import os
+import shutil
+import tempfile
+from collections.abc import Generator
+from pathlib import Path
+
+import airbyte as ab
+import pytest
+import ulid
+from airbyte.caches.base import CacheBase
+from airbyte.caches.duckdb import DuckDBCache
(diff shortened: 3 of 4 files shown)


airbyte/sources/util.py
```diff
@@ -50,6 +51,7 @@ def get_source(
     version: str | None = None,
     pip_url: str | None = None,
     local_executable: Path | str | None = None,
+    docker_executable: bool = False,
```
aaronsteers, line 54: Wdyt of refactoring to take `docker_image` (or `docker_image:tag`) here as a string? 

This would be mutually exclusive with `local_executable` and `pip_url`. If provided, we'd invoke the named image.

Couple things to decide:

1. Should we attempt "docker pull" if the image is not present? (I think 'yes', meaning the user doesn't need to pre-pull the image.)
2. Should we automatically re-pull on every invocation? (I think 'no', meaning the user _does_ need to force a re-pull if they want the latest.)

Curious to your thoughts here. 🤔

airbyte/sources/util.py
```diff
@@ -115,6 +117,28 @@ def get_source(
...
+                executable=[
+                    "docker",
+                    "run",
+                    "--rm",
+                    "-i",
+                    "--volume",
+                    f"{temp_dir}:{temp_dir}",
+                    f"airbyte/{name}:{version}",
+                ],
+            ),
```
aaronsteers, line 139: Following from the above, I think we can take the image name (and optional version) as input, in place of `f"airbyte/{name}:{version}"`.

And then from there, I wonder if we can reduce the constructor to only require the image name? Or maybe also the local temp_dir to use for storing docker volume files?

airbyte/_executor.py
```diff
@@ -450,3 +451,55 @@ def uninstall(self) -> NoReturn:
...
+
+
+class DockerExecutor(Executor):
+    def __init__(
+        self,
+        name: str | None = None,
+        *,
+        executable: list[str],
+        target_version: str | None = None,
+    ) -> None:
```
aaronsteers, line 477: See notes in other comments re: suggestion to take in docker image (and tag?) along with optional temp dir(s).

aaronsteers (commented):
@Tarekk - Thanks so much for this contribution. We made a deliberate decision to not include Docker support in our initial launch, in order to prioritize the best-possible Python-native experience. That said, this contribution is an exciting one and would be welcomed by many users who _do_ have a Docker runtime and who would like to leverage the broadest possible set of connectors.

In terms of path forward, we will discus internally on how best to deliver/position this, but for now I've left some inline feedback re: a few small comments on the implementation details. Lmk what you think on those.

On the bigger-picture roadmap/delivery topic, I will circle back later this week. I'm optimistic we can leverage this - and will confirm once I've got confirmation either way.

Thanks so much - and congrats on your first (BIG) open source contribution! 🙌

aaronsteers:
@Tarekk - Thanks for your patience. We're going to move forward and merge this, but we'll keep this functionality (for a short while) under a new `airbyte.experimental` module. No need to make any changes here - I'll tackle that work post-merge of this PR.

I did see when invoking tests that the we're seeing failures on the Windows runners - but Ubuntu runners are passing all tests 🎉 . We already skip other docker-dependent user tests on Windows, so I would suggest the following:

1. We should try to catch docker-executor failures when on Windows is detected, we report these under an Exception class that messages that Windows isn't supported.
   - Technically it is possible with WSL, but by default, Windows docker backends will be Windows-based containers.
3. We can skip integration tests related to docker executable if we detect we are running on Windows. We have this pattern in place for other tests, so we can replicate it for this case also.

airbyte/sources/util.py
Tarekk: My thought process around adding a Boolean flag here was to make the end user experience of switching from the default venv execution to Docker seamless. I.e. user can keep all other arguments the same and simply "flip the switch". But looking at it again now I am making the [assumption](https://github.com/airbytehq/PyAirbyte/pull/209/files#diff-c3ea7142407ab026659c30b3c6e11abf4010665385b372f00a994de4c250dccaR137) that an [official Airbyte source](https://hub.docker.com/u/airbyte) will be used. So yes I think refactoring to a ```docker_image``` string will make this more flexible and compatible with any image that follows the Airbyte Protocol!

airbyte/sources/util.py
Tarekk: Yes that makes sense to me! No need to take in the entire docker run command and it's arguments as those are constants. For temp_dir I think we should keep that fixed as it needs to match the PyAirbyte temp_dir [here](https://github.com/airbytehq/PyAirbyte/blob/main/airbyte/_util/temp_files.py#L23) unless there is a way for overriding that which I am not seeing?

Tarekk:
Thank you @aaronsteers for reviewing this PR and bringing it up for review with the team, it's great to hear that it will be merged in 😃 And thank you fro the valuable feedback, I replied to your comments above and they all make sense to me!

I am not very familiar with the testing infra but sounds like we would exclude running tests on Windows somewhere [here](https://github.com/airbytehq/PyAirbyte/blob/main/tests/conftest.py#L67-L76)?

tests/integration_tests/test_docker_executable.py
```diff
@@ -0,0 +1,297 @@
...
+    ]
+
+
+def test_which_source_faker() -> None:
+    """Test that source-faker is available on PATH."""
+    assert shutil.which(
+        "source-faker"
+    ), f"Can't find source-faker on PATH: {os.environ['PATH']}"
+
+
```
aaronsteers: We can delete this test since it doesn't actually test the docker invocation path (happens to pass but for other reasons).

```suggestion
```

tests/integration_tests/test_docker_executable.py
```diff
@@ -0,0 +1,297 @@
...
+FAKER_SCALE_A = 200
+# We want this to be different from FAKER_SCALE_A.
+FAKER_SCALE_B = 300
+
+
+# Patch PATH to include the source-faker executable.
+
+
+@pytest.fixture(scope="function")  # Each test gets a fresh source-faker instance.
+def source_faker_seed_a() -> ab.Source:
```
aaronsteers: We can make this name distinct and then apply skip logic centrally.

```suggestion
def source_docker_faker_seed_a() -> ab.Source:
```

Note this will also need to be renamed below in the file, whereever the fixture is used.

tests/integration_tests/test_docker_executable.py
```diff
@@ -0,0 +1,297 @@
...
+    source.select_streams([
+        "users",
+        "products",
+        "purchases",
+    ])
+    return source
+
+
+@pytest.fixture(scope="function")  # Each test gets a fresh source-faker instance.
+def source_faker_seed_b() -> ab.Source:
```
aaronsteers: Ditto. We can make this name distinct and then apply skip logic centrally.

```suggestion
def source_docker_faker_seed_b() -> ab.Source:
```

aaronsteers (commented):
Second round of feedback, just focused on testing this time.

> I am not very familiar with the testing infra but sounds like we would exclude running tests on Windows somewhere [here](https://github.com/airbytehq/PyAirbyte/blob/main/tests/conftest.py#L67-L76)?

Yes, that's correct. There are probably many ways to do this (TIMTOWDI), but making distinct fixture names for docker-based source execution, and then riding the same rails to auto-skip those as you call out above is a good path, I think. 👍

Tarekk:
Thank you for the guidance @aaronsteers updated to add the docker fixtures to the exclusion list.

aaronsteers:
@Tarekk - I just ran the test suite and all tests are green ✅ 🎉 🙌 

We'll try to merge this in within the next 24-48 hours. As noted above, I'll take a follow-on PR to move this into an `experimental` module while we iterate and gather feedback from the community, but this won't disrupt your ability to use the functionality in the meanwhile post-release.

CC @bindipankhudi

Tarekk:
That's great to hear @aaronsteers 🥳 appreciate your help with getting this landed!!

bindipankhudi:
@aaronsteers, just took a pass at this PR. Looks great to me! No additional comments from me. The implementation feels very clean and loving the integration tests :). Nicely done @Tarekk

PR #220, opened 2024-05-08
Feat: CatalogManager refactor, decoupling sql processors, caches, and internal sql backend tables

aaronsteers:
This PR does a lot of things...

## CatalogManager refactor

The role of catalog manager is refactored into the new classes, under the role of two "backends", two "providers", and a "writer" (for state).

- `StateBackend` - Responsible for serializing and deserializing state in the internal SQL tables.
- `CatalogBackend` - Responsible for serializing and deserializing catalog info in the internal SQL tables.
- `StateProvider` - Provides state inputs to anything that needs state - namely to `Source` objects, which use the state to know where to start their sync.
- `CatalogProvider` - Provides catalog schema info to any method that needs access to catalog metadata.
- `StateWriter` - There are two implementations:
   - `StdOutStateWriter` Performs the role as a destination would - which is to simply print the state message to `STDOUT`.
   - `SqlStateWriter` - Writes state messages to the internal SQL table, which is how a Cache should behave.

With these class in mind, the "Backend" classes generally are able to create "provider" classes when they are the source of the state or catalog. Also, much of the parsing logic for how to handle catalogs was able to be moved out of the processor and cache classes, and into the `CatalogProvider` class - for instance, getting the primary keys of a stream or getting a stream's properties.

- Refactor CatalogManager to isolate StateManager functions in a new class.
- Create abstract base classes of both of the above so we can decouple them from their SQL-based serialization implementations.
- Create simple versions of these two classes that simply accept input from upstream.
- Allow creation of SQLProcessor objects with State and Catalog artifacts provided explicitly, rather than being read/written to internal SQL tables.
- Move classes that we want to move to the CDK into a new `_future_cdk` module. These would be internal classes anyway, but the goal is to move these abstract implementations upstream so destinations can depend upon them.

## Decoupling `SQLProcessor` classes from `SQLBackend` and `Cache` classes.

We want to be able to use `SQLProcessor` classes in destinations like the Cortext destination, and so this PR decouples `SQLProcessor` from `Cache` and `Backend` classes (previously the `CatalogManager`). Now, instead of passing a backend to the `SQLProcessor`, we simply pass a `CatalogProvider` and `StateWriter`.  When a state provider is not explicitly created, the `SQLProcessor` class will just create its own `StdOutStateWriter` class, and will behave like a destination. (See the Cortext sample script for an example that doesn't require a cache.)

One last change was that, since we don't want `SQLProcessor` to depend on `Cache` class, we needed a different way to provide the inputs that previously were mapped to an embedded `cache` property in the `SQLProcessor` classes. Now, we decouple the functionality of _configuration_ into a new set of `SQLConfig` classes - these are basically the _just the user inputs_ that would be provided via the `Cache`, but without the behavioral traits of a full `Cache` class. `SQLConfig` objects know their properties and they know how to create `SQLAlchemy` connections (as well as 3rd-party vendor connections, when applicable), but they are lightweight and can be handed down to the SQLProcessor classes. They can also be imported or implemented by destinations, without the need to create a full cache class.

For backwards compatibility and ease-of-use, the `Cache` classes inherit from the respective `SQLConfig` classes, so they accept the same config inputs in their constructor as they did previous to this PR (no breaking change) and they can pass themselves to the `SQLProcessor` classes as a valid subclass of `SQLConfig`. Typing prevents the processor from performing cache-related functions on the `SQLConfig` instances, even though the object may _also_ be a `Cache`. This ensures that we don't end up with any circular dependencies between the role of the `Cache` and the `SQLProcessor`, and it also ensures that destination connectors that use the `SQLProcessor` class can send the lighter-weight `SQLConfig` object to the `SQLProcessor` initializer, and they won't need to create a full `Cache` object.

## Eligible for CDK Re-Use

- `RecordProcessor`
- `SQLProcessor`
- `CatalogProvider`
- `StateProvider`
- `StateWriter`
- `StdOutStateWriter`


airbyte/_future_cdk/__init__.py
```diff
@@ -0,0 +1,16 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
+"""Module for future CDK components.
+
+Components here are planned to move to the CDK.
+
+TODO!: Add GitHub link here before merging.
+"""
+
+from __future__ import annotations
```
bindipankhudi, line 9: great idea to move the base classes into future_cdk module! will be easy to move it out later.

airbyte/_future_cdk/catalog_manager.py
```diff
@@ -0,0 +1,152 @@
...
+class CachedStream(Base):  # type: ignore[valid-type,misc]
+    __tablename__ = STREAMS_TABLE_NAME
+
+    stream_name = Column(String)
+    source_name = Column(String)
+    table_name = Column(String, primary_key=True)
+    catalog_metadata = Column(String)
+
+
+class CatalogManagerBase(abc.ABC):
```
bindipankhudi: would it make sense to name this file to be catalog_manager_base.py. I got confused earlier since there is the `airbyte/catalog_manager` that inherits from it as well. Although when we move it out to the cdk, should not be an issue.

bindipankhudi:
Finished taking a pass at the code. The refactoring is easy to follow and makes sense to me. It is def much more intuitive and cleaner to have a separate statemanager and catalogmanager. This would make the cortex processor implementation much cleaner!!

aaronsteers:
Tests are passing _except_ one new issue with Windows paths. That will be fixed shortly, but it doesn't change anything in the core code.

examples/run_snowflake_cortex_test_data.py
```diff
@@ -155,14 +156,17 @@ def _state(data: dict[str, Any]) -> AirbyteMessage:
...
 stream_names = set()
 stream_names.add("myteststream")
 processor = SnowflakeCortexSqlProcessor(
-    cache=cache,
-    catalog=catalog,
+    sql_config=cache,
     vector_length=5,
-    source_name="github",
-    stream_names=stream_names,
+    catalog_provider=catalog,
```
aaronsteers:
```suggestion
    catalog_provider=catalog_provider,
```

airbyte/_future_cdk/record_processor.py
```diff
@@ -245,25 +257,26 @@ def _setup(self) -> None:  # noqa: B027  # Intentionally empty, not abstract
...
         self,
         stream_name: str,
     ) -> dict[str, Any]:
-        """Return the column definitions for the given stream."""
-        return self._get_stream_config(stream_name).stream.json_schema
+        """Return the column definitions for the given stream.
+
+        This method is a convenience wrapper around the catalog manager implementation.
+        """
+        return self.catalog_provider.get_stream_json_schema(stream_name)
```
aaronsteers: Still needed? 🤔

airbyte/_future_cdk/sql_processor.py
```diff
@@ -74,7 +75,55 @@ class SQLRuntimeError(Exception):
...
+        raise NotImplementedError(
+            f"The type '{type(self).__name__}' does not define a custom client."
+        )
+
+
+@dataclass
+class SqlTableDomain:
+    database_name: str
+    schema_name: str
+    table_prefix: str
```
aaronsteers: Remove?

```suggestion
```

airbyte/_future_cdk/sql_processor.py
```diff
@@ -89,129 +138,84 @@ class SqlProcessorBase(RecordProcessor):
...
-            engine=self.get_sql_engine(),
-            table_name_resolver=self.get_sql_table_name,
+        if not temp_dir and not file_writer:
+            raise exc.PyAirbyteInternalError(
+                message="Either `temp_dir` or `file_writer` must be provided.",
+            )
+
+        state_writer = state_writer or StdOutStateWriter()
+
+        self._sql_config: Engine = sql_config
```
aaronsteers:
```suggestion
        self._sql_config: SqlConfig = sql_config
```

airbyte/caches/_catalog_backend.py
```diff
@@ -0,0 +1,266 @@
...
+        source_name: str,
+        incoming_source_catalog: ConfiguredAirbyteCatalog,
+        incoming_stream_names: set[str],
+    ) -> None:
+        """Register a source and its streams in the cache."""
+        self._save_catalog_info(
+            source_name=source_name,
+            incoming_source_catalog=incoming_source_catalog,
+            incoming_stream_names=incoming_stream_names,
+        )
```
aaronsteers, line 142: Double check order of operations here?

bindipankhudi (approved):
Looks great to me! Thank you for the PR walkthrough! 👍

airbyte/caches/_catalog_backend.py
aaronsteers: Order of operations was correct, but comments are needed to clarify the reason why save is happening before merge. Added comments to cover this.

airbyte/_future_cdk/record_processor.py
aaronsteers: Removed.

PR #230, opened 2024-05-14
Feat: Accept docker image names, move `DockerExecutor` support to `experimental` module

aaronsteers:
This PR moves the newly-expanded `get_source` implementation into a new `airbyte.experimental` module, and it reverts the default `get_source` implementation to not support Docker (yet).

Also, I've renamed the argument from `docker_executor` (`bool`) to `docker_image` which can be `bool` (`True`) _or_ it can be a string naming the docker image, with an optional tag.


bindipankhudi (approved):
Looks great!

PR #234, opened 2024-05-15
Chore: Bump Snowflake client and `cryptography` package

### LINKED ISSUE #221: Update snowflake-python-connector dependency to address security risk
PyAirbyte currently has a dependency on `snowflake-python-connector==3.6.0`, which in turn depends on an outdated version of the `cryptography` library (<4.2). This outdated version poses a security risk.

To mitigate this risk, it's recommended to update the `snowflake-python-connector` dependency to a version that uses a more recent `cryptography` version (>=4.2).

Please consider updating the dependency to ensure a secure environment for PyAirbyte users.

aaronsteers:
Resolves #221

This allows `cryptography>=42.0`...

> image


aaronsteers:
We're starting to see weird hanging on Snowflake client. Merging (without resolved lint checks) so we can pull into the PR that does contain the lint checks 😄

PR #236, opened 2024-05-15
CI: Add '/poetry-lock' slash command

aaronsteers:
This adds a new `/poetry-lock` slash command. It copies all code from the `/fix-pr` slash command, and adapts it slightly for this use case.

> **[Note]**
> This is in auto-merge status, ready to merge on approval.


bindipankhudi (approved):
Looks good to me! 🥳

Issue #233, opened 2024-05-14
🐛 Bug: PyAirbyte fails when field contains special characters (e.g. `discountapplied(%)`)


aaronsteers:
Reported today, if source stream has a property called `discountapplied(%)`, PyAirbyte fails with the message:

`Illegal field name: discountapplied(%)`

Issue #232, opened 2024-05-14
🐛 Bug: Unexpected behavior if stream name is non-standard


aaronsteers:
We've had a couple reports this week of unexpected behaviors from PyAirbyte when the incoming stream name has weird characters and/or weird input format.

Example stream name of "*" is one example. PyAirbyte does not complain but then is later unable to correctly insert into the correct table. (In this case was used with BigQuery cache.)

PR #239, opened 2024-05-15
Feat: Handle special chars in column and table name normalization

aaronsteers:
This addresses:

- #233 
- #232 

What this PR does, as documented in the code:

Please see the unit test for a description of the logic and examples.


aaronsteers:
@edgao - I've updated the code to reflect the logic you shared in our slack thread. Do you mind reviewing the latest and approve if everything basically matches your expectations?

(Ignore stale PR description quotes.)

airbyte/_util/name_normalizers.py
```diff
@@ -46,8 +49,43 @@ class LowerCaseNormalizer(NameNormalizerBase):
...
     @staticmethod
     def normalize(name: str) -> str:
-        """Return the normalized name."""
-        return name.lower().replace(" ", "_").replace("-", "_")
+        """Return the normalized name.
+
+        - All non-alphanumeric characters are replaced with underscores.
+        - "%" is replaced with "pct".
+        - "#" is replaced with "num".
+        - "+" is replaced with "plus".
```
aaronsteers:
```suggestion
```

bindipankhudi (approved):
lgtm! 👍

edgao (approved):

aaronsteers:
There was a very sticky bug that surfaced and delayed this merging.

Here's the finding... we previously were running an extra "normalize()" operation on Snowflake, when evaluating `$1."column_name"`, which is snowflake-speak for getting "column_name" from a variant column. By running an unnecessary normalization, we were inadvertantly normalizing at the wrong place, and converting the expression to `$1._column_name_`, which was incorrect and returning null values all around. (Previously this had no effect, because the quote characters were being left alone in the prior iteration of the normalize code.

This 👆 is fixed. So I'll merge this PR shortly.

PR #243, opened 2024-05-16
Chore: Clean up missing column additions logic

aaronsteers:
This brings together the logic from the old "ensure_compatible_schema" implementation with the newer "add_missing_columns".

When we added the "add_missing_columns" functionality, we neglected to update/incorporate the prior implementation.


bindipankhudi (approved):
makes sense to me! 🚀

PR #245, opened 2024-05-17
Chore: Configure warnings as errors in `pytest`

aaronsteers:
We haven't done any hygeine for warnings, and over time a large number of these have accumulated. When users see these, they can cause confusion - and worse, sometimes they are signs of real issues that need to be addressed.

This PR does a few things:

- Make warnings raise errors in PyTest.
- Add execptions to `pyproject.toml` for any errors we _do_ expect.
- Update specific packages in pyproject.toml which are generating deprecation warnings.


aaronsteers:
@bindipankhudi - FYI, this is a draft for now, but I hope to finish today. This came up when researching a user-facing issue.

PR #250, opened 2024-05-20
Fix: Cortex processor - normalize table names before passing in query

bindipankhudi:
Related to: https://github.com/airbytehq/airbyte/issues/38353

Note:
This branch is created from 0.10.5 tag and the changes will be patched on top of 0.10.5 release.


airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -107,9 +107,10 @@ def _get_column_list_from_table(
         This is overridden due to lack of SQLAlchemy compatibility for the
         `VECTOR` data type.
         """
+        normalized_table_name = self.get_sql_table_name(table_name)
```
aaronsteers: Technically, the get_sql_table_name() method is supposed to take in a stream name and output the table name. And this method should receive a table name also, which should have already been normalized. While it will work in most cases, the problem would be that if there's a table prefix, that could end up getting applied twice.

Is there a way to move this operation upstream in the flow and still get a valid fix? And I think there's another method to normalize without risking adding a new prefix.


I'm on my mobile right now but will be able to dive in a bit deeper shortly...

bindipankhudi: good point, i will take a look tomm.

airbyte/_processors/sql/snowflakecortex.py
```diff
@@ -107,9 +107,10 @@ def _get_column_list_from_table(
         This is overridden due to lack of SQLAlchemy compatibility for the
         `VECTOR` data type.
         """
+        normalized_table_name = self.normalizer.normalize(table_name)
```
bindipankhudi, line 110: updated to use self.normalize which replaces spaces and hyphens with underscore..

airbyte/_processors/sql/snowflakecortex.py
bindipankhudi: updated to use `self.normalizer.normalize(...)`

airbyte/_processors/sql/snowflakecortex.py
aaronsteers: Nice - yeah, this is great. I will also look to see why it was not already normalized. Whenever we pass a "table_name" arg, it should be already normalized. But no reason to block you if this is working.

aaronsteers (approved):
Approved!  ✅ 

Merge when ready. (Feel free to ignore linting issues.)

Issue #244, opened 2024-05-17
source-zendesk-support: TypeError: unhashable type: 'list'


nakamichiworks:
I got the following error when I ran `source-zendesk-support` connector.

```python
import os

import airbyte as ab

source = ab.get_source(
    "source-zendesk-support",
    config={
        "subdomain": os.getenv("ZENDESK_API_SUBDOMAIN"),
        "start_date": "2024-05-17T00:00:00Z",
        "credentials": {
            "credentials": "api_token",
            "email": os.getenv("ZENDESK_API_EMAIL"),
            "api_token": os.getenv("ZENDESK_API_TOKEN"),
        },
    },
)
source.check()
source.select_streams(["tickets"])
result = source.read()
```

```
Traceback (most recent call last):
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/sources/base.py", line 734, in read
    cache_processor.process_airbyte_messages(
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/_future_cdk/record_processor.py", line 217, in process_airbyte_messages
    self.write_all_stream_data(
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/_future_cdk/record_processor.py", line 226, in write_all_stream_data
    self.write_stream_data(
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/_future_cdk/sql_processor.py", line 508, in write_stream_data
    final_table_name = self._ensure_final_table_exists(
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/_future_cdk/sql_processor.py", line 402, in _ensure_final_table_exists
    for column_name, sql_type in self._get_sql_column_definitions(
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/_future_cdk/sql_processor.py", line 476, in _get_sql_column_definitions
    columns[clean_prop_name] = self.type_converter.to_sql_type(
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/types.py", line 123, in to_sql_type
    airbyte_type, _ = _get_airbyte_type(json_schema_property_def)
  File "/private/tmp/test-pyairbyte/.venv/lib/python3.9/site-packages/airbyte/types.py", line 67, in _get_airbyte_type
[...]

PR #251, opened 2024-05-20
Fix: Resolve error `TypeError` "unhashable type list" (2nd pass)

aaronsteers:
A fix for #244.

This is smaller scope than:

- https://github.com/airbytehq/PyAirbyte/pull/246

In that PR, we modify the failover type to be `object`/`JSON`/`VARIANT` for cases of `anyOf(string, object)`. In this case, we don't change anything except to make sure that the expected exception is raised - and this triggers the expected failsafe behavior, without changing it.

### DIFF
--- a/airbyte/types.py
+++ b/airbyte/types.py
@@ -64,8 +64,13 @@ def _get_airbyte_type(  # noqa: PLR0911  # Too many return statements
         if json_schema_format == "time":
             return "time_without_timezone", None
 
-    if json_schema_type in {"string", "number", "boolean", "integer"}:
-        return cast(str, json_schema_type), None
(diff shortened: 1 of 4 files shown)


bindipankhudi (approved):
lgtm! 👍

aaronsteers:
Noting here for follow-up, I'm not getting consistent rendering across database platforms when I try to validate records converted from object type to string type. The differences are minor, but enough to break tests significantly. 

E.g. one database will write the string `'{"nested_column": "nested_value"}'` and another will write `'{"nested_column":"nested_value"}'`. I'm simplifying test inputs so we can prioritize the known fix to type resolution itself, and we can come back to this in:

- #253

PR #259, opened 2024-05-23
Fix: Add back default schema name `'airbyte_raw'` for caches

aaronsteers:
This reverts back to `airbyte_raw` as the default schema name.

Some database types lost this default when we refactored in either `0.10.x` or `0.11.x`.

Anyone who is providing an explicit schema name will not be affected by this change.

⚠️ Set to auto-merge


bindipankhudi (approved):
Looks great! 👍

PR #269, opened 2024-06-03
Chore: Bump DuckDB from `0.9` to stable version `v1.x`

aaronsteers:
This should not be a breaking change for existing PyAirbyte users who use DuckDB as a local cache.

For reference on backwards compatibility matters:

- https://duckdb.org/2024/02/13/announcing-duckdb-0100.html
- https://duckdb.org/2024/06/03/announcing-duckdb-100.html

Optionally, users may choose to upgrade their existing cache DB files using the method described here:

- https://duckdb.org/docs/internals/storage#how-to-move-between-storage-formats

Note to MotherDuck users:

1. You will need to use this newer version in order to connect with MotherDuck accounts on post-0.9 versions of DuckDB.
2. If you have not already done so, you can upgrade your MotherDuck account by navigating to the settings page.
3. New MotherDuck accounts automatically use the newer (stable) database format, which requires this update.

Manual Testing performed:

- ✅ Prep: Updated `examples/run_faker.py` to use a persistent path location (`get_default_cache()`).
- ✅ Create a cache in 0.9.x:
  - Switch to `main` and re-run `poetry install`. Confirmed that DuckDB was downgraded.
  - Run `poetry run python ./examples/run_faker.py`. Confirm everything ran correctly.
- ✅ Write to the cache in 1.x:
  - Switch back to this branch and re-run `poetry install`. Confirmed that DuckDB was upgraded.
  - Run `poetry run python ./examples/run_faker.py`. Confirm everything ran correctly.

> ⚠️ Auto-Merge Enabled ⚠️


bindipankhudi (approved):
Looks great! 🥳

PR #296, opened 2024-07-13
Fix: Normalize schema names for correct name comparison

### LINKED ISSUE #279: Add an option not to create schema for SnowflakeCache
When I use `SnowflakeCache`, pyairbyte always tries to `create schema if not exists ...` in [SqlProcessorBase._ensure_schema_exists()](https://github.com/airbytehq/PyAirbyte/blob/37b50db7ec5a1121d3f6c51b7faef850f35317da/airbyte/_future_cdk/sql_processor.py#L303).

This behavior is, however, not desirable if the target schema already exists and I don't want the snowflake user to have privileges to create or update the schema. (In such situations, `_ensure_schema_exists()` throws an error like `"Insufficient privileges to operate on database 'your_database'"` even if the schema already exists.)

Could you add an option to avoid this error?

Currently, my workaround is to define  a wrapper class like this:

```python
from airbyte._processors.sql.snowflake import SnowflakeSqlProcessor
from airbyte.caches import SnowflakeCache
from pydantic import PrivateAttr

class PatchedSnowflakeSqlProcessor(SnowflakeSqlProcessor):
    def _ensure_schema_exists(self) -> None:
        return None


class PatchedSnowflakeCache(SnowflakeCache):
    _sql_processor_class = PrivateAttr(default=PatchedSnowflakeSqlProcessor)
```

nakamichiworks:
Fixes #279.

Since sqlalchemy always returns normalized schema names, user given schema name must be also normalized for correct comparison.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved schema name normalization in SQL processing, enhancing compatibility with various SQL engine dialects.
	- Enhanced robustness of schema existence checks by standardizing both target schema names and existing schema lists.


aaronsteers:
@coderabbitai review

aaronsteers:
@nakamichiworks - Thanks for submitting this! Can you clarify for me if the issue is related to casing or any other specific normalization issue? 

For context on casing, we have an internal `LowerCaseNormalizer` class which we apply in all cases, and for Snowflake we are using settings that are intentionally case insensitive or case-normalized.

airbyte/_future_cdk/sql_processor.py
```diff
@@ -303,8 +303,8 @@ def _get_table_by_name(
     def _ensure_schema_exists(
         self,
     ) -> None:
-        """Return a new (unique) temporary table name."""
-        schema_name = self.sql_config.schema_name
+        dialect = self.get_sql_engine().dialect
+        schema_name = dialect.normalize_name(self.sql_config.schema_name)
         if schema_name in self._get_schemas_list():
             return
```
aaronsteers: I think we should try to use the internal normalizer, just to avoid depending too much on SQLAlchemy dialects. Our normalizer class has methods for both `normalize()` and `normalize_set(Iterable[str])`.

Something like `schema_name = self.normalizer.normalize(...)`...

Or perhaps just making `if schema_name in self._get_schemas_list()` into a case-insensitive check, something like `if self.normalizer.normalizer(schema_name) in self.normalizer.normalize_set(self._get_schemas_list()`?

Wdyt?

nakamichiworks:
Thanks for your review and sorry for a late reply.

`LowerCaseNormalizer` is sufficient to resolve my original issue and I have updated the PR to use it.

However, it might be best to define and use something like `SQLAlchemyNormalizer` for consistency because schema names returned by `SqlProcessorBase._get_schemas_list()` is possibly normalized by sqlalchemy's `Inspector.get_schema_names()`, whose implementation depends on DB dialects (Snowflake implementation is [here](https://github.com/snowflakedb/snowflake-sqlalchemy/blob/main/src/snowflake/sqlalchemy/snowdialect.py#L826)).

aaronsteers:
> Thanks for your review and sorry for a late reply.
> 
> `LowerCaseNormalizer` is sufficient to resolve my original issue and I have updated the PR to use it.
> 
> However, it might be best to define and use something like `SQLAlchemyNormalizer` for consistency because schema names returned by `SqlProcessorBase._get_schemas_list()` is possibly normalized by sqlalchemy's `Inspector.get_schema_names()`, whose implementation depends on DB dialects (Snowflake implementation is [here](https://github.com/snowflakedb/snowflake-sqlalchemy/blob/main/src/snowflake/sqlalchemy/snowdialect.py#L826)).

Thank you for this suggestion. I am a bit wary to adopt Snowflake's own case normalization because some string casings are sometimes-unreachable given Snowflake's odd "case-unaware, case-sensitive" handling. Instead, we try to exclusively interface with Snowflake using the [QUOTED_IDENTIFIERS_IGNORE_CASE](https://docs.snowflake.com/en/sql-reference/parameters#quoted-identifiers-ignore-case) setting enabled.

Since MIxeD case identifiers are sometimes unreachable, we internally lower-case everything, with the expectation that Snowflake will upper-case them internally. It's not a perfect solution, per se, and we will certainly consider adapting the internal dialect version in the future. 👍

Issue #321, opened 2024-07-30
Add logic to `_ensure_compatible_table_schema()` to expand column sizes if needed.


aaronsteers:
https://github.com/airbytehq/PyAirbyte/blob/73eaa0ee5a04d2e5fac086d8662a52b612584039/airbyte/_future_cdk/sql_processor.py#L435

PR #327, opened 2024-08-06
Ensure compatible table schema by adding missing columns and expanding column sizes   #321

GovindHede:
### Summary

This pull request ensures that the table schema is compatible with the stream's schema by adding any missing columns and expanding column sizes if necessary.

### Changes Made

1. Implemented `_ensure_compatible_table_schema` method to:
   - Add any missing columns to the table.
   - Ensure column sizes are compatible with the stream's schema.
2. Added helper methods:
   - `_get_stream_schema`: Retrieve the schema for the specified stream.
   - `_get_table_schema`: Retrieve the schema of the specified table.
   - `_is_size_expansion_needed`: Check if the column size needs to be expanded.
   - `_generate_alter_column_statement`: Generate an ALTER TABLE statement for expanding column size.
   - `_execute_alterations`: Execute a list of ALTER TABLE statements.

### Issue Reference

    - 321

### DIFF
--- a/airbyte/_future_cdk/sql_processor.py
+++ b/airbyte/_future_cdk/sql_processor.py
@@ -424,17 +424,94 @@ class SqlProcessorBase(RecordProcessorBase):
         stream_name: str,
         table_name: str,
     ) -> None:
-        """Return true if the given table is compatible with the stream's schema.
+        """Ensure the given table is compatible with the stream's schema.
 
         Raises an exception if the table schema is not compatible with the schema of the
         input stream.
         """
-        # TODO: Expand this to check for column types and sizes.
-        # https://github.com/airbytehq/pyairbyte/issues/321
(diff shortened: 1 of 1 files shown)


aaronsteers:
@GovindHede - This is exciting! Thanks for contributing to PyAirbyte. 🙏 

It looks like great progress here and I'd be excited to merge this when ready.

Lmk by tagging me if you need any assistance. I applied the `/fix-pr` slash command as admin on the repo - which does the equivalent of auto-fix with `ruff` lint and format.

In case it is helpful, you can use these CI jobs to check any issues:

- `Run Tests / Pytest (Fast)` - just the fast tests, failing fast.
- `Run Tests / Pytest (No Creds)` - a bit more tests, but nothing requiring creds.
- `Run Linters / Ruff Lint Check` - lint/format checks.

PR #330, opened 2024-08-09
Fix: Add "Destinations 1.0" handling to PyAirbyte: sync IDs, generation IDs, and stream success statuses

### LINKED ISSUE #329: PyAirbyte destinations (e.g. `destination-bigquery`) fail with `You must upgrade your platform version to use this connector version.`
Related to: https://github.com/airbytehq/airbyte/pull/43417

The newest specification for destinations will fail if either `generationId` or `minGenerationId` is missing in the configured catalog.

Fix would be for PyAirbyte to inject `1` and `1` as static values.

aaronsteers:
Resolves #329

So far, this PR only gives a script to repro the issue. See #329 and https://github.com/airbytehq/airbyte/pull/43417 for more info.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new example script for integrating with Google BigQuery, demonstrating a complete data pipeline process.
	- Added a catalog validation method to ensure necessary fields are initialized, enhancing the robustness of catalog management.
	- Enhanced stream processing feedback by adding success messages upon stream completion.

- **Bug Fixes**
	- Improved catalog handling by enforcing default states for specific identifiers during initialization, preventing potential data processing issues.
	- Enhanced message handling for unclosed streams, improving resource management during processing.

### DIFF
--- a/airbyte/_future_cdk/catalog_providers.py
+++ b/airbyte/_future_cdk/catalog_providers.py
@@ -10,8 +10,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, final
 
-import ulid
-
 from airbyte_protocol.models import (
     ConfiguredAirbyteCatalog,
 )
--- a/airbyte/_future_cdk/catalog_providers.py
+++ b/airbyte/_future_cdk/catalog_providers.py
@@ -10,6 +10,8 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Any, final
 
+import ulid
+
 from airbyte_protocol.models import (
     ConfiguredAirbyteCatalog,
 )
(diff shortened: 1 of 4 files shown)


aaronsteers:
Logging here for visibility:

I resolved the three missing catalog properties and then received this issue due to the source not emitting an explicit `SUCCESS` signal:

```
2024-08-08 17:51:13 - INFO - INFO main i.a.c.i.d.a.AsyncStreamConsumer(close):207 Some streams either received an INCOMPLETE stream status, or did not receive a stream status at all: null.users, null.products, null.purchases
```

I can force-emit this message from PyAirbyte, but IMHO, this defeats the point of having a success indicator in the first place - as it destroys the confidence that a destination might have that the stream is _actually_ complete. My preference would be for destinations _not_ to require this, for the reason that forcing it be faked by PyAirbyte or the platform remove the confidence that it means anything.

I'm going to go ahead and make the change - since destinations are already in the wild with this spec - but hopefully we can consider reverting this in a future version of Java CDK.

PR #340, opened 2024-08-19
Fix: Updated handling for Manifest-only connectors; Breaks: Remove Python 3.9 support

aaronsteers:
This fixes the language check to include the new `'manifest-only'` language type, and it updates PyAirbyte to locate manifest yaml in their new directory.

This PR also requires dropping Python 3.9 support, as we've previously discussed in Slack. This is necessary in order to bump to the latest CDK version, which is necessary to install the latest manifest-only connectors.

Notes:
- PyAirbyte always called these `source_manifest` connectors, so no rename is needed. "Low-code" and "manifest-only" are both the same category from the PyAirbyte user perspective.
- The git refs can go away once we have manifests being published to GCS:
  - https://github.com/airbytehq/airbyte/pull/42584
- `source-faker` was previously a dev dependency, which created conflicts when trying to bump the CDK. This PR removes `source-faker` as a dev dependency and tests that rely on it will now auto-install the connector to complete their tests. This makes the diff much messier, but it is required in order to bump the CDK.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the configurability of the DeclarativeExecutor by adding a name parameter for instance identification.
	- Introduced a mechanism for validating manifest structures to ensure compatibility with the executor.

- **Improvements**
	- Simplified manifest retrieval logic, consolidating error handling and improving maintainability.
	- Updated integration tests to ensure the installation of sources when they are missing, enhancing test reliability.
	- Added a new label for better handling of connector metadata based on associated tags.

- **Updates**
	- Minimum Python version requirement increased to 3.10, ensuring compatibility with newer features.
	- Adjusted dependency specifications for improved version management.
	- Updated GitHub Actions workflows to support only Python versions 3.10 and 3.11.

### DIFF
--- a/airbyte/_executors/declarative.py
+++ b/airbyte/_executors/declarative.py
@@ -9,12 +9,13 @@ from pathlib import Path
 from typing import IO, TYPE_CHECKING, cast
 
 import pydantic
+import yaml
 
 from airbyte_cdk.entrypoint import AirbyteEntrypoint
 from airbyte_cdk.sources.declarative.manifest_declarative_source import ManifestDeclarativeSource
 
+from airbyte import exceptions as exc
 from airbyte._executors.base import Executor
-from airbyte.exceptions import PyAirbyteInternalError
 
 
 if TYPE_CHECKING:
@@ -40,6 +41,7 @@ class DeclarativeExecutor(Executor):
 
     def __init__(
         self,
+        name: str,
         manifest: dict | Path,
     ) -> None:
         """Initialize a declarative executor.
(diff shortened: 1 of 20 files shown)


airbyte/_executors/declarative.py
```diff
@@ -49,22 +51,36 @@ def __init__(
...
         self.declarative_source = ManifestDeclarativeSource(source_config=self._manifest_dict)
 
         # TODO: Consider adding version detection
         # https://github.com/airbytehq/airbyte/issues/318
         self.reported_version: str | None = None
 
+    def _validate_manifest(self, manifest_dict: dict) -> None:
+        """Validate the manifest."""
+        manifest_text = json.dumps(manifest_dict)
+        if "class_name:" in manifest_text:
```
maxi297, line 73: My understanding is that this means we don't support any connector with custom components. I think this is fair but there were some talks a while back to support custom components in the Connector Builder. I don't know where those talks are at today but it could be interesting for the marketplace team to know that such change will have an impact on PyAirbyte

airbyte/_executors/util.py
```diff
@@ -143,49 +195,23 @@ def get_connector_executor(  # noqa: PLR0912, PLR0913, PLR0915 # Too complex
...
-
-        # Source manifest is a dict at this point
-        return DeclarativeExecutor(
-            manifest=source_manifest,
-        )
+        if isinstance(source_manifest, (str, bool)):
+            # Source manifest is either a URL or a boolean (True)
+            source_manifest = _try_get_source_manifest(
+                source_name=name,
+                manifest_url=None if source_manifest is True else source_manifest,
```
maxi297, line 208: I'm trying to figure out a bit more about PyAirbyte ; in which case `get_connector_executor` is called with `source_manifest` being a URL?

maxi297 (commented):
More a review for me to learn PyAirbyte than for confirming the change

Issue #181, opened 2024-04-11
🐛 Bug: Google Ads fails due to nested primary keys


aaronsteers:
The Google Ads source is failing due to nested primary keys. Logged here in slack: https://airbytehq-team.slack.com/archives/C06FZ238P8W/p1712842623104989?thread_ts=1712842217.390879&cid=C06FZ238P8W

Example error message:

- `NotImplementedError: Nested primary keys are not yet supported. Found: ad_group.id`

Source code:

- https://github.com/airbytehq/PyAirbyte/blob/ce20b56c708d8814e775ce5391ce9ba295497be1/airbyte/_future_cdk/sql_processor.py#L802-L806

aaronsteers:
As a first step to resolving, we should try to find out (1) which stream is causing this error and (2) if that stream is standard or dynamic from the user workspace and/or user config.

sukantaroy01:
Facing this error, @aaronsteers do we have any RCA on this after the issue was initially raised.

sukantaroy01:
Sample code I am running:

```python

source_gads = ab.get_source('source-google-ads', install_if_missing=True, config=google_payload.get('configuration'))
source_gads.select_streams(['ad_group'])

result = source_gads.read()
```


Response:
```txt
Read Progress
Started reading at 14:08:55.

Read 1,562 records over 7 seconds (223.1 records / second).

Wrote 1,562 records over 1 batches.

Finished reading at 14:09:03.

Started finalizing streams at 14:09:03.

Finalized 0 batches over 0 seconds.

Failed `source-google-ads` read operation at 19:39:03.
```


Error stacktrace:

```

NotImplementedError                       Traceback (most recent call last)
File ~/pixis/pyAirbyte-prefect-POC/myenv/lib/python3.11/site-packages/airbyte/sources/base.py:804, in Source.read(self, cache, streams, write_strategy, force_full_refresh, skip_validation)
    [803](https://file+.vscode-resource.vscode-cdn.net/Users/sukantaroy/pixis/pyAirbyte-prefect-POC/~/pixis/pyAirbyte-prefect-POC/myenv/lib/python3.11/site-packages/airbyte/sources/base.py:803) try:
[...]

PR #343, opened 2024-08-20
Fix: Resolve issue where PyAirbyte would fail if property names contain the dot character (`'.'`), e.g. with `source-google-ads`

sukantaroy01:
Fixes: #181

During a proof of concept (POC) with the Google Ads source, I encountered an issue where primary keys containing periods (e.g., campaign.id, adset.id) were not being correctly handled. The root cause was that get_primary_keys did not normalize column names before checking for nested keys.

To address this, I implemented normalization of primary keys prior to the nested key check. This resolved the issue, allowing successful data collection from the Google Ads platform.

## Summary by CodeRabbit

- **New Features**
	- Enhanced primary key processing by introducing normalization for consistency and compatibility.
	- Added support for a new stream type with a primary key containing a dot, expanding integration capabilities.
- **Bug Fixes**
	- Improved validation for primary keys to prevent inconsistencies related to case sensitivity.
- **Chores**
	- Updated the `.gitignore` file to exclude PyCharm project files for a cleaner repository.


sukantaroy01:
@aaronsteers Can you please look into this PR once you have some bandwidth?

aaronsteers:
@Udit107710 and @sukantaroy01 - Thanks for your work on this. Do you have the ability to enable commits from maintainers/reviewers? (Looks like it is not enabled as of now.)

tests/integration_tests/fixtures/source-test/source_test/run.py
```diff
@@ -65,6 +65,23 @@
...
+                "supported_sync_modes": ["full_refresh"],
+                "json_schema": {
+                    "$schema": "http://json-schema.org/draft-07/schema#",
+                    "type": "object",
+                    "properties": {
+                        "table1.Column1": {"type": "string"},
+                        "table1.Column2": {"type": "number"},
+                        "table1.empty_column": {"type": "string"},
+                        "table1.big_number": {"type": "number"},
+                    },
```
aaronsteers, line 82: Thanks for adding this test! 🙏 

I think the schema we are trying to handle is slightly different...

~~In this adaptation, the "." delimiter is a means of navigating to the subnode, but not actually part of the property names.~~

aaronsteers: My comment above is incorrect! Please disregard. I misunderstood the root cause. After digging into the source records, I found that the property name just contains the "." character, and so it is not "nested" in the way that I originally interpreted it - and as the guard statement in PyAirbyte was trying to prevent.

image

aaronsteers: Given my improved understanding now, it seems what is needed is just to escape the "." when we are normalizing the name, which I think would cause `ad_group.id` to become `ad_group_id`.

aaronsteers: I've opened a second PR targeting this one:

- https://github.com/pixisai/PyAirbyte/pull/9

aaronsteers:
Opened second PR targeting this one: 

- https://github.com/pixisai/PyAirbyte/pull/9

Udit107710:
Thanks @aaronsteers Merged your PR!

tests/unit_tests/test_text_normalization.py
```diff
@@ -211,6 +211,7 @@ def test_case_insensitive_w_pretty_keys(
         ("", "", True),
         ("*", "", True),
         ("!@$", "", True),
+        ("some.col", "some_col", False),
```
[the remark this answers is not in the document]
aaronsteers: @coderabbitai - I think you misread the test case. `False` is the answer to `should_raise`, meaning whether the normalization should fail with an exception. 

Hence, I think it is correct as-is.

aaronsteers:
@sukantaroy01 - This looks good and I really appreciate your investment here. When I ran on the other PR, there were still some failing tests, mostly related to listing the stream names in the source-test fixture, and similar trivial changes that broke existing tests.

That said, this PR can't accept my contributions so I will go ahead and merge, to resolve the small test failures in a subsequent PR which I can tackle fairly quickly.

Thanks very much for contributing to PyAirbyte! 🎉 ❤️

aaronsteers (approved):
Amazing work here. Thanks so much! ✅

PR #545, opened 2024-12-02
Fix: Added regex to sub special characters

### LINKED ISSUE #544: 🐛Bug: Problem with special Characters in Stream Names
Preface: This isn't directly a PyAirbyte Bug and more of an OS Bug. Im using Ubuntu 22.04 on WSL.

# Description
In some systems custom Tables or the like are possible. If one of these custom tables has a special character in it's name the jsonl.gz file, used to import everything into the DuckDB cache, cannot be created. Therefore the whole sync fails. We have a case where a custom table in a customer's system which contains a backward slash `\`.  Since we are using a custom connector for this customer and it's the system of said customer, i'm not able to share logs. maybe at a later stage we can release the connector to the main airbyte repo.

# Possible Fix

In `file_writers.py` the method [_get_new_cache_file_path](https://github.com/tinomerl/PyAirbyte/blob/main/airbyte/_writers/file_writers.py#L55) generates a unique name by adding a uuid / batch_id. By my understanding it should be possible to add a line which would remove special characters in the jsonl.gz file.

# Willing to create PR
Yes. Eager to try something out.

tinomerl:
Fixes #544.

I've tested the fix with three different sources.

<details><summary>Source File</summary>

```python

import airbyte as ab

file_content = "col_1,col_2\nfoo,bar\n"
with open("test \ special chars.csv", "w") as f:
    f.write(file_content)

source: ab.Source = ab.get_source("source-file")
config = {
    "dataset_name": "test \ special chars",
    "format": "csv",
    "url": "test \ special chars.csv",
    "provider": {"storage": "local"}
}

source.set_config(config)
source.select_all_streams()
records = source.read()
records.streams.get("test \\ special chars").to_pandas()
```

</details>

<details><summary>PokeApi</summary>
<p>

```python

import airbyte as ab

source: ab.Source = ab.get_source(    "source-pokeapi",
    config={"pokemon_name": "bulbasaur"},
    source_manifest=True,
)

source.check()
source.get_available_streams()
source.select_all_streams()
records = source.read()

```

</p>
</details> 

It also worked on the custom Connector mentioned in the Bug Issue.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved file path generation by cleaning stream names to remove special characters, ensuring compatibility with operating system restrictions.

### DIFF
--- a/airbyte/_writers/file_writers.py
+++ b/airbyte/_writers/file_writers.py
@@ -65,7 +65,7 @@ class FileWriterBase(AirbyteWriterInterface):
         # If a stream contains a special Character, the temporary jsonl.gz
         # file can't be created, because of OS restrictions. Therefore, we
         # remove the special characters.
-        cleaned_stream_name = re.sub(r"[^a-zA-Z0-9\s]", "", stream_name)
+        cleaned_stream_name = re.sub(r'[<>:"/\\|?*\x00-\x1F]', "", stream_name)
         return target_dir / f"{cleaned_stream_name}_{batch_id}{self.default_cache_file_suffix}"
 
     def _open_new_file(
--- a/airbyte/_writers/file_writers.py
+++ b/airbyte/_writers/file_writers.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import abc
+import re
 from collections import defaultdict
 from pathlib import Path
 from typing import IO, TYPE_CHECKING, final
@@ -61,7 +62,8 @@ class FileWriterBase(AirbyteWriterInterface):
         batch_id = batch_id or str(ulid.ULID())
         target_dir = Path(self._cache_dir)
         target_dir.mkdir(parents=True, exist_ok=True)
-        return target_dir / f"{stream_name}_{batch_id}{self.default_cache_file_suffix}"
+        cleaned_stream_name = re.sub(r"[^a-zA-Z0-9\s]", "", stream_name)
+        return target_dir / f"{cleaned_stream_name}_{batch_id}{self.default_cache_file_suffix}"
 
     def _open_new_file(
         self,
--- a/airbyte/_writers/file_writers.py
+++ b/airbyte/_writers/file_writers.py
@@ -62,6 +62,9 @@ class FileWriterBase(AirbyteWriterInterface):
         batch_id = batch_id or str(ulid.ULID())
         target_dir = Path(self._cache_dir)
         target_dir.mkdir(parents=True, exist_ok=True)
+        # If a stream contains a special Character, the temporary jsonl.gz
+        # file can't be created, because of OS restrictions. Therefore, we
+        # remove the special characters.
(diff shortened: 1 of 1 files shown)


airbyte/_writers/file_writers.py
```diff
@@ -61,7 +62,11 @@ def _get_new_cache_file_path(
         batch_id = batch_id or str(ulid.ULID())
         target_dir = Path(self._cache_dir)
         target_dir.mkdir(parents=True, exist_ok=True)
-        return target_dir / f"{stream_name}_{batch_id}{self.default_cache_file_suffix}"
+        # If a stream contains a special Character, the temporary jsonl.gz
+        # file can't be created, because of OS restrictions. Therefore, we
+        # remove the special characters.
```
aaronsteers:
```suggestion
        # If a stream contains a special Character, the temporary jsonl.gz
        # file can't be created, because of OS restrictions. Therefore, we
        # remove the special characters.
        # Specifically: we remove any of these characters: `<>:"/\|?*`
        # and we remove characters in the ASCII range from 0 to 31.
```

aaronsteers (commented):
@tinomerl - This is a valuable update - thank you!

Should we include similar logic also when designating a table name?

tinomerl:
Hey @aaronsteers,

thanks, that's great to hear. I think there is already some kind of logic implemented for table names. Just looked at the table name for my file test and it looks like this, in the default cache:

image

Maybe i should adapt the same replacement pattern for filenames. So instead of removing any of the characters with nothing, using the underscore could be better. I think that would also cover some edge cases better. Like for example if we have something like this:

- `myStreamDataFromA\B`
- `myStreamDataFromAB`

They would currently be named the same. But with the underscores they would differ. What do you think?

aaronsteers:
@tinomerl - thanks for confirming that the table names are working well already. The method used for table name normalization is in the LowerCaseNormalizer class. You may be able to use the same here for normalizing the stream name prior to creating the path.

I'm ok with either approach but if the lower case normalizer class works well, that might be slightly easier to maintain, since the same code would be applied for both cases. 

Want to try and let me know what you find? If for any reason that doesn't work well, I think your approach here is perfectly sound.

tinomerl:
@aaronsteers - thanks for the Infos. I'll try it out tomorrow and let you know the results.

tinomerl:
@aaronsteers - So i just changed it to use the `LowerCaseNormalizer` instead of just the regex pattern. And it works just fine. I like it this way more because, like you said, it can now be maintained at  just one spot.

aaronsteers (approved):

aaronsteers:
@tinomerl - This is great. I'll run a few more tests and merge/release later today if all is passing! Thanks again for this contribution.  🙏

aaronsteers:
I meant to merge earlier but ran into some test failures. All is looking good now so I'll go ahead and merge+release.

PR #614, opened 2025-02-14
feat: add "runas_host_user" param to get_source()

nicob3y:
Docker in Docker context.

I need to execute a containerized source with same UID as the  containerized host executing script in order to simplify file permissions management.
"config" and "catalog" files generated thanks to "tempfile" library have 600 POSIX perms and the "runas_host_user" approach helped me to allow source container to read this ones. 

It works like the `use_host_network` param, simply set `runas_host_user` to `True` while calling `get_source()` method.

## Summary by CodeRabbit

- **New Features**
  - Connectors now offer an option to execute with host user privileges, improving file permission handling and resource access.
  
- **Deprecations**
  - An older connector method has been deprecated in favor of the updated implementation.

### DIFF
--- a/airbyte/_executors/util.py
+++ b/airbyte/_executors/util.py
@@ -1,6 +1,7 @@
 # Copyright (c) 2023 Airbyte, Inc., all rights reserved.
 from __future__ import annotations
 
+import os
 import tempfile
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


nicob3y:
resolved by #637

PR #629, opened 2025-03-19
Permit SQL based caches to drop dependent objects when swapping temp tables with final ones

nicob3y:
In the case we use the cache as a built-in destination we might expect that table-dependent objects  would be droppped during a refresh.
Use case is also described on the Postgres destination connector (https://docs.airbyte.com/integrations/destinations/postgres#creating-dependent-objects)


aaronsteers:
@nicob3y - Thank you for this contribution! The change looks good to me except I don't think every SQL dialect supports CASCADE. 

Could we implement this on the specific cache implementations(s) where needed?

PR #668, opened 2025-05-15
feat: add ability to override pk and cursor key values

aaronsteers:
## Summary by CodeRabbit

- **New Features**
  - Added the ability to override cursor keys and primary keys for individual streams using new configuration options.
  - Introduced methods to set these overrides either individually or in bulk for greater flexibility.

- **Improvements**
  - Enhanced catalog generation to automatically apply any configured cursor key or primary key overrides when streams are set up.

- **Tests**
  - Added comprehensive unit tests to verify correct handling of cursor key and primary key overrides across multiple scenarios.

- **Chores**
  - Updated GitHub Actions workflows to use Poetry version 2.0.1 for dependency management.
  - Skipped flaky integration tests related to cloud SQL reads pending fixes.

### DIFF
--- a/airbyte/sources/base.py
+++ b/airbyte/sources/base.py
@@ -185,9 +185,9 @@ class Source(ConnectorBase):  # noqa: PLR0904
         - Stream names are case sensitive and are not validated by PyAirbyte. If the stream name
           does not exist in the catalog, the override may be ignored.
         """
-        self._primary_key_overrides.update({
-            k: v if isinstance(v, list) else [v] for k, v in kwargs.items()
-        })
+        self._primary_key_overrides.update(
+            {k: v if isinstance(v, list) else [v] for k, v in kwargs.items()}
+        )
 
     def _log_warning_preselected_stream(self, streams: str | list[str]) -> None:
         """Logs a warning message indicating stream selection which are not selected yet."""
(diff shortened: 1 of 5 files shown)


airbyte/sources/base.py
```diff
@@ -97,6 +105,90 @@ def set_streams(self, streams: list[str]) -> None:
         )
         self.select_streams(streams)
 
+    def set_cursor_key(
+        self,
+        stream_name: str,
+        cursor_key: str,
+    ) -> None:
+        """Set the cursor for a single stream.
```
aaronsteers, line 118: After merge, these docstrings will render into API docs here:

- https://airbytehq.github.io/PyAirbyte/airbyte/sources/base.html

marcosmarxm (commented):
@aaronsteers are we considering cases users can override to use a non-existing field for pk or cursor?

aaronsteers:
@marcosmarxm - re:

> @aaronsteers are we considering cases users can override to use a non-existing field for pk or cursor?

As of now, we don't validate the field names provided by the user. However, the source will probably (and should ideally) fail if it receives a field reference that does not exist. More validation could be added in the future, but the main reason we don't pre-validate is that we (probably) don't yet know the discovered catalog at the time that the user is providing the overrides.

The main use case for this is for db-type sources, where no source-defined primary key or cursor key might exist. So, in docs, I document that if the source doesn't accept the overrides (like if they try to override a cursor key in stripe) the input might be ignored (whether a valid valid or not).

Does that make sense?

Issue #707, opened 2025-07-02
source-mixpanel raises PyAirbyteNameNormalizationError


nickolasclarke:
Opening a ticket per request from @aaronsteers in this [community slack thread](https://airbytehq.slack.com/archives/C06FZ238P8W/p1751321461271109)

I'm running into a PyAirbyteNameNormalizationError when I attempt to run the simple example below.

Example config.json
```json
{
  "credentials": {
    "project_id": 123,
    "secret": "my_secret",
    "username": "my_service_account"
  },
  "start_date": "2025-06-30T21:00:00Z",
  "region": "US" #aside: this appears to be required, despite what https://docs.airbyte.com/integrations/sources/mixpanel suggests
}
```

Run this simple example:
```python

import airbyte as ab
from utils import load_json_config

mp_config = load_json_config("starbyte/sources/mixpanel/mixpanel_config.json")
mixpanel_source = ab.get_source(
    name="source-mixpanel", config=mp_config, streams=["export"]
)
res = mixpanel_source.read(cache=ab.get_default_cache())

for name, records in res.streams.items():
    print(f"Stream {name}: {len(records)} records")
```

Output
```
#From STDOUT
Started reading from source at 15:07:31:

 • Read 68,074 records over 6.1 seconds (11,097.3 records/s, 8.33 MB/s).
 • Received records for 1 streams:
    • 68,074 export
```

```
#FROM LOG
tail -f /tmp/airbyte/logs/source-mixpanel/source-mixpanel-log-JZ1E2HRF0.log
2025-06-30 15:07:32 - INFO - Starting syncing SourceMixpanel
2025-06-30 15:07:32 - INFO - Marking stream export as STARTED
2025-06-30 15:07:32 - INFO - Setting state of SourceMixpanel stream to {}
2025-06-30 15:07:32 - INFO - Syncing stream: export
2025-06-30 15:07:43 - INFO - Marking stream export as RUNNING
```

This raises the following: See the full trace after this block.
```
Exception has occurred: PyAirbyteNameNormalizationError
Name cannot be empty after normalization. (PyAirbyteNameNormalizationError)
------------------------------------------------------------
PyAirbyteNameNormalizationError: Name cannot be empty after normalization.
    Raw Name: ''
    Normalization Result: ''
KeyError: ''

During handling of the above exception, another exception occurred:

  File ".../simple_example.py", line 8, in <module>
    res = mixpanel_source.read(cache=ab.get_default_cache())
[...]

PR #709, opened 2025-07-03
#7 Handle undefined keys in rows

nickolasclarke:
This should resolve https://github.com/airbytehq/PyAirbyte/issues/707 by handling record field keys that are undefined. This is technically valid json, but undesirable. Not sure if this is the desired approach, however, so I'll keep this in draft and hold off on tests until I hear more.

### DIFF
--- a/airbyte/_util/name_normalizers.py
+++ b/airbyte/_util/name_normalizers.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 import abc
 import functools
 import re
+import uuid
 from typing import TYPE_CHECKING
 
 from airbyte import exceptions as exc
@@ -68,6 +69,10 @@ class LowerCaseNormalizer(NameNormalizerBase):
         - "-1" -> "_1"
         """
         result = name
+        if not result:
+            # Use a short hash of the original name to avoid collisions.
+            uuid_suffix = uuid.uuid1().hex[:4]
+            return f"undefined_key_{uuid_suffix}"
 
         # Replace all non-alphanumeric characters with underscores.
         result = re.sub(r"[^A-Za-z0-9]", "_", result.lower())


airbyte/_util/name_normalizers.py
```diff
@@ -68,6 +69,10 @@ def normalize(name: str) -> str:
         - "-1" -> "_1"
         """
         result = name
+        if not result:
+            # Use a short hash of the original name to avoid collisions.
+            uuid_suffix = uuid.uuid1().hex[:4]
```
nickolasclarke: 4 characters should be more than enough, but this can always be expanded.

airbyte/_util/name_normalizers.py
```diff
@@ -68,6 +69,10 @@ def normalize(name: str) -> str:
         - "-1" -> "_1"
         """
         result = name
+        if not result:
```
nickolasclarke: @aaronsteers this should probably log out a WARNING, but I dont immediately see the pattern for doing so in this repo.

airbyte/_util/name_normalizers.py
```diff
@@ -68,6 +69,10 @@ def normalize(name: str) -> str:
...
+            # Use a short hash of the original name to avoid collisions.
+            uuid_suffix = uuid.uuid1().hex[:4]
+            return f"undefined_key_{uuid_suffix}"
```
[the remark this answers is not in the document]
nickolasclarke: I believe that cache busting here would be fine because we do not want to get key collisions.

airbyte/_util/name_normalizers.py
aaronsteers: I think `warn_once` may be a good fit for this.

https://github.com/airbytehq/PyAirbyte/blob/679498794a1619906146c5d31390be9ef02f3015/airbyte/logs.py#L48-L58

airbyte/_util/name_normalizers.py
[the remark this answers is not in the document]
aaronsteers: I think the challenge with uuid here is that re-running the sync will end up with a new column name on subsequent executions. Some other implementations which need to solve a similar problem apply an ordinal id or letter like `_undefined_key_1`, `_undefined_key_2`, etc.

Another option is to simply treat unnamed properties as ignored. Did you find the key node that was having a null/empty field name? Or is the root cause something different?

I'm not sure what the best path is here. The AI had a good instinct, but bad suggestions. For example: if "result" is an empty string, then hashing it (`hashlib.md5(str(result).encode())`) won't be helpful for uniqueness if other field names could also be blank/empty.

yohannj: Considering this method is used to normalize names, this behavior feels very unnatural to me.

Rather than this implementation, what about skipping columns with an empty key completely? 
https://github.com/airbytehq/PyAirbyte/blob/679498794a1619906146c5d31390be9ef02f3015/airbyte/records.py#L209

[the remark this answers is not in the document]
nickolasclarke: Yes, the problematic data in question, was JSONL results that contained an empty key with a value `"":my_still_valuable_value"`, which is technically valid JSON. This was a bug on our side, but is still present in the historical records. 
 
 An ordinal ID is probably the best solution here if you want to maintain the KV pair, but it would require passing additional state into this function, which feels wrong to me as well. I think warning in the logs skipping the row is probably best.

nickolasclarke: yes I think I agree now. See my reply here: https://github.com/airbytehq/PyAirbyte/pull/709/files#r2192886996

It seems that if we do want to skip, this has to be done before the StreamRecordHandler is instantiated on [L215](https://github.com/airbytehq/PyAirbyte/blob/679498794a1619906146c5d31390be9ef02f3015/airbyte/records.py#L215) correct @aaronsteers?

nickolasclarke: Actually, I think we may be overthinking this. This edge case will only occur when there is a single falsy key, you cannot have a dict with multiple falsy keys. So perhaps my original implementation without the hash and throwing a warning would be sufficient. Python would throw when casting the json to a dict. 

Otherwise I could add something like 
```python
        for k, v in from_dict.items():
            if not k:
                warn_once(
                    f"Empty key found in StreamRecord initialization with value: {v}. Ignoring.",
                )
                from_dict = {k: v for k, v in from_dict.items() if k}
                break
```
at [L215](https://github.com/airbytehq/PyAirbyte/blob/679498794a1619906146c5d31390be9ef02f3015/airbyte/records.py#L215
)

I'm happy with either implementation.

yohannj: Yes, something like that.
It could also be a variable like `invalid_keys = ['']` that we use to remove some keys.
It could also be:
```python
if '' in from_dict:
  warn_once(f"Empty key found in StreamRecord. Ignoring.")
  del from_dict['']
```

airbyte/records.py
```diff
@@ -215,6 +216,11 @@ def __init__(
             extracted_at: The time the record was extracted. If not provided, the current time will
                 be used.
         """
+        # warn if any keys in from_dict are empty and remove them.
+        if "" in from_dict:
+            warn_once(f"Empty key found in StreamRecord. Ignoring.", with_stack=False)
+            del from_dict[""]
```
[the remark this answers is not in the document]
nickolasclarke: I dont know that this has any real impact in practice, but I do generally avoid the use of `del` and this is how I had originally done this. So..sure!

nickolasclarke:
@aaronsteers there does not seem to be unit tests against the `StreamRecord` class directly. I could put in a unit test in the normalization testers, but that feels out place. Let me know where or if I should put tests in for this, and I'll do so.

PR #731, opened 2025-07-31
feat: Add Python version compatibility checking during connector installation

aaronsteers:
# feat: Add Python version compatibility checking during connector installation

## Summary

This PR adds Python version compatibility checking to PyAirbyte's connector installation process. When installing a connector, PyAirbyte now:

1. **Fetches version requirements from PyPI** - Queries the PyPI JSON API to get the `requires_python` field for the connector package
2. **Validates current Python version** - Compares the current Python interpreter version against the connector's requirements using the `packaging` library
3. **Issues warnings for incompatibilities** - Uses `warn_once()` to display user-friendly warnings when the current Python version may not be compatible
4. **Fails silently when appropriate** - Skips version checking entirely when `AIRBYTE_OFFLINE_MODE` is enabled or when PyPI is unreachable

**Example behavior:** When installing `airbyte-source-hubspot` (which requires `"<3.12,>=3.10"`) on Python 3.12, users will see a warning but installation proceeds.

- [ ] **Test with actual version-constrained connector** - Install `airbyte-source-hubspot` on Python 3.12+ to verify warning appears correctly
- [ ] **Verify no performance regression** - Measure installation time impact for normal connector installations
- [ ] **Test offline mode behavior** - Confirm no network hangs when `AIRBYTE_OFFLINE_MODE=1` is set
- [ ] **Validate warning behavior** - Ensure warnings appear only once per session and don't break installation flow
- [ ] **Check existing workflows** - Verify that current connector installation patterns still work normally

**Recommended test plan:**
1. Install a connector with Python version constraints on an incompatible Python version
2. Install the same connector in offline mode to ensure no network calls
3. Install a normal connector to verify no regression in performance or functionality

---

### Diagram

```mermaid
%%{ init : { "theme" : "default" }}%%
graph TD
    subgraph "VenvExecutor.install()"
        A["VenvExecutor.install()"]:::major-edit
        B["_get_pypi_python_requirements()"]:::major-edit
        C["_check_python_version_compatibility()"]:::major-edit
        D["Create venv + pip install"]:::context
    end
    
    subgraph "External Dependencies"
        E["PyPI JSON API<br/>pypi.org/pypi/{package}/json"]:::context
        F["packaging library<br/>SpecifierSet, Version"]:::context
        G["warn_once()<br/>from airbyte.logs"]:::context
    end
    
    subgraph "Configuration"
        H["AIRBYTE_OFFLINE_MODE<br/>from airbyte.constants"]:::context
        I["ConnectorMetadata<br/>pypi_package_name"]:::context
    end
    
    A --> B
    B --> E
    B --> H
    A --> C
    C --> F
    C --> G
    A --> D
    B --> I
    
    subgraph Legend
        L1[Major Edit]:::major-edit
        L2[Minor Edit]:::minor-edit  
        L3[Context/No Edit]:::context
    end

classDef major-edit fill:#90EE90
classDef minor-edit fill:#87CEEB
classDef context fill:#FFFFFF
```

### Notes

- **Implementation follows existing patterns** - Uses same PyPI API approach as other PyAirbyte utilities, respects existing offline mode handling, and leverages established warning infrastructure
- **Silent failure by design** - All error scenarios (network failures, parsing errors, missing metadata) fail silently to avoid breaking existing installations
- **Performance consideration** - PyPI API call adds ~100-500ms to installation time but only occurs during actual installations, not version checks
- **Session info** - Implemented by @aaronsteers in Devin session: https://app.devin.ai/sessions/01b58a09a41d49789670f0586fcd98cf


aaronsteers:
FYI - Ignore these unrelated errors:

`E           AssertionError: Expected 'source-mailchimp' init success but got 'ValidationError'.`

airbyte/_executors/python.py
```diff
@@ -193,6 +207,70 @@ def get_installed_version(
 
             return None
 
+    def _get_pypi_python_requirements(self, package_name: str) -> str | None:
```
aaronsteers: Add lru_cache here.

airbyte/_executors/python.py
```diff
@@ -193,6 +208,71 @@ def get_installed_version(
...
+                headers={"User-Agent": f"PyAirbyte/{get_version()}"},
+                timeout=10,
+            )
+            response.raise_for_status()
+
+            data = response.json()
+            return data.get("info", {}).get("requires_python")
+
+        except Exception:
+            return None
```
aaronsteers: This needs a clear comment to explain why we are being so dismissive of exception handling.

airbyte/_executors/python.py
```diff
@@ -193,6 +208,71 @@ def get_installed_version(
...
+                warn_once(
+                    f"Python version compatibility warning for '{package_name}': "
+                    f"Current Python {current_version} may not be compatible with "
+                    f"package requirement '{requires_python}'. "
+                    f"Installation will proceed but may fail.",
+                    with_stack=False,
+                )
+
+        except Exception:
+            pass
```
aaronsteers: Ditto. This at least needs a comment.

airbyte/_executors/python.py
```diff
@@ -193,6 +208,71 @@ def get_installed_version(
...
+
+        Example:
+            For airbyte-source-hubspot, returns "<3.12,>=3.10"
+        """
+        if AIRBYTE_OFFLINE_MODE:
+            return None
+
+        try:
+            url = f"https://pypi.org/pypi/{package_name}/json"
+            response = requests.get(
```
aaronsteers: Handle none here.

```suggestion
            version = get_version()
            response = requests.get(
                url=url,
                headers={"User-Agent": f"PyAirbyte/{version}" if version else "PyAirbyte"},
```

airbyte/_executors/python.py
```diff
@@ -5,24 +5,71 @@
...
+
+    try:
+        url = f"https://pypi.org/pypi/{package_name}/json"
+        version = get_version()
+        response = requests.get(
+            url=url,
+            headers={"User-Agent": f"PyAirbyte/{version}" if version else "PyAirbyte"},
+            timeout=10,
+        )
+        response.raise_for_status()
```
aaronsteers: Instead of raise_for_status, why not just check if success and exit early if not. Saves the messy exception handling.

airbyte/_executors/python.py
```diff
@@ -193,6 +248,49 @@ def get_installed_version(
...
+        """Check if current Python version is compatible with package requirements.
+
+        Args:
+            package_name: Name of the package being checked
+            requires_python: The requires_python constraint from PyPI (e.g., "<3.12,>=3.10")
+        """
+        if not requires_python:
+            return
+
+        try:
```
aaronsteers: Let's remove this try/catch with throwaway exception. Instead, rely on the called function to resolve into None if something goes awry.

airbyte/_executors/python.py
```diff
@@ -193,6 +248,49 @@ def get_installed_version(
 
             return None
 
+    def _get_pypi_python_requirements(self, package_name: str) -> str | None:
+        """Get the requires_python field from PyPI for a package."""
+        return _get_pypi_python_requirements_cached(package_name)
```
aaronsteers: We don't need this method anymore. Just have the caller call the cached function directly.

airbyte/_executors/python.py
```diff
@@ -5,24 +5,69 @@
...
+
+    try:
+        data = response.json()
+        return data.get("info", {}).get("requires_python")
+    except Exception:
+        # Intentionally broad exception handling to ensure silent failure in all scenarios:
+        # - JSON parsing errors (json.JSONDecodeError)
+        # - Any other unexpected errors
+        # This ensures connector installation never fails due to version checking issues.
+        return None
```
aaronsteers: We've narrowed this enough, I think, that we should just use the supress() context manager with the exception(s) we want to ignore.

```
data: dict | None = None
with supress(SomeException):
    data = response.json()
if not data:
    return None
return data.get("info", {}).get("requires_python")
```

airbyte/_executors/python.py
```diff
@@ -193,6 +243,39 @@ def get_installed_version(
...
             return None
 
+    def _check_python_version_compatibility(
+        self,
+        package_name: str,
+        requires_python: str | None,
+    ) -> None:
+        """Check if current Python version is compatible with package requirements.
+
+        Args:
```
aaronsteers: Update the implementation as follows:

```suggestion
    def _check_python_version_compatibility(
        self,
        package_name: str,
        requires_python: str | None,
    ) -> bool | None:
        """Check if current Python version is compatible with package requirements.
        
        Returns True if confirmed, False if incompatible, or None if no determination can be made.

        Args:
```

airbyte/_executors/python.py
```diff
@@ -193,6 +243,44 @@ def get_installed_version(
...
+        Returns True if confirmed, False if incompatible, or None if no determination can be made.
+
+        Args:
+            package_name: Name of the package being checked
+            requires_python: The requires_python constraint from PyPI (e.g., "<3.12,>=3.10")
+        """
+        if not requires_python:
+            return None
+
+        try:
```
aaronsteers: Again, please remove the try/catch/else block. Nothing here should fail.

airbyte/_executors/python.py
```diff
@@ -106,6 +148,14 @@ def install(self) -> None:
 
         After installation, the installed version will be stored in self.reported_version.
         """
+        package_name = (
+            self.metadata.pypi_package_name
+            if self.metadata and self.metadata.pypi_package_name
+            else f"airbyte-{self.name}"
+        )
+        requires_python = _get_pypi_python_requirements_cached(package_name)
+        self._check_python_version_compatibility(package_name, requires_python)
```
aaronsteers: Let's make it explicit we are no-op-ing the return value on purpose.
I'll apply this and you can pull the latest.

```suggestion
         _ = self._check_python_version_compatibility(package_name, requires_python)
```

airbyte/_executors/python.py
```diff
@@ -5,24 +5,66 @@
...
+    Returns:
+        The requires_python string from PyPI, or None if unavailable
+
+    Example:
+        For airbyte-source-hubspot, returns "<3.12,>=3.10"
+    """
+    if AIRBYTE_OFFLINE_MODE:
+        return None
+
+    url = f"https://pypi.org/pypi/{package_name}/json"
```
aaronsteers, line 47: For my info, is there a versioned endpoint of this?

airbyte/_executors/python.py
```diff
@@ -5,24 +5,66 @@
...
+from airbyte.logs import warn_once
+from airbyte.version import get_version
 
 
 if TYPE_CHECKING:
     from airbyte.sources.registry import ConnectorMetadata
 
 
+@lru_cache(maxsize=128)
+def _get_pypi_python_requirements_cached(package_name: str) -> str | None:
```
aaronsteers, line 32: Add a test for this function.

airbyte/_executors/python.py
```diff
@@ -5,24 +5,68 @@
...
+    Example:
+        For airbyte-source-hubspot, returns "<3.12,>=3.10"
+    """
+    if AIRBYTE_OFFLINE_MODE:
+        return None
+
+    url = f"https://pypi.org/pypi/{package_name}/json"
+    version = get_version()
+
+    with suppress(Exception):
```
aaronsteers: I thought we removed this 'suppress' - either revert or tell me why it's needed. And if you revert, report back on first failure.

airbyte/_executors/python.py
```diff
@@ -193,6 +245,40 @@ def get_installed_version(
 
             return None
 
+    def _check_python_version_compatibility(
+        self,
+        package_name: str,
+        requires_python: str | None,
+    ) -> bool | None:
```
aaronsteers: This seems like a waste of a class method. Move to a util module, maybe in a new '_semver' submodule, and then simplify by calling the semver helper upstream. Doesn't seem like it belongs here.

pyproject.toml
```diff
@@ -29,6 +29,7 @@ google-cloud-secret-manager = "^2.17.0"
 jsonschema = ">=3.2.0,<5.0"
 orjson = "^3.10"
 overrides = "^7.4.0"
+packaging = "^23.0"
```
aaronsteers: Common module. Probably safer to just:


```suggestion
packaging = ">=23.0"
```

aaronsteers:
/poetry-lock
> `poetry lock` job started... [Check job output.][1]

[1]: https://github.com/airbytehq/PyAirbyte/actions/runs/16666820107
> ✅ `poetry lock` applied successfully.

aaronsteers:
/poetry-lock
> `poetry lock` job started... [Check job output.][1]

[1]: https://github.com/airbytehq/PyAirbyte/actions/runs/16667650655
> ✅ `poetry lock` applied successfully.

PR #838, opened 2025-10-23
feat(airbyte-internal): Add methods and MCP tools for connector version pinning 

aaronsteers:
# Add MCP tools and PyAirbyte methods for connector version pinning

## Summary
Adds MCP tools and PyAirbyte class methods to pin/unpin connector versions in Airbyte Cloud using the internal Config API's scoped configuration system. This allows users to override the default connector version for specific source/destination instances.

**Key implementation details:**
- Uses `/v1/scoped_configuration/get_context` endpoint to find existing overrides (avoiding the non-scalable `/list` endpoint)
- Pins versions at the ACTOR scope (individual connector instances)
- Requires admin access via environment variables (`AIRBYTE_INTERNAL_ADMIN_FLAG=airbyte.io` and `AIRBYTE_INTERNAL_ADMIN_USER`)
- The `AIRBYTE_INTERNAL_ADMIN_USER` env var must be an email that contains the text `@airbye.io`.
- Tracks override reason (min 10 chars) and optional reference URL for audit purposes
- Looks up user UUID via `/v1/users/list_instance_admin` endpoint

**⚠️ CRITICAL: This PR has NOT been tested end-to-end.** The implementation was developed through API investigation but never validated against a live environment. Previous attempts encountered 500 errors, leading to multiple iterations on the approach.

- [ ] **Test complete workflow**: Deploy test source → get version → set override → verify pinned → unset → verify unpinned
- [ ] **Verify API primitives are correct**: The `resource_type` (set to connector_type: "source"/"destination"), `resource_id` (set to connector_id), and scope configuration may be incorrect. Early testing returned 500 errors with different primitive combinations.
- [ ] **Confirm `/get_context` endpoint works**: This endpoint was chosen to avoid the non-scalable `/list` endpoint, but its actual behavior hasn't been validated
- [ ] **Test admin access flow**: Verify user lookup via `/users/list_instance_admin` works and returns the correct user UUID
- [ ] **Test error cases**: Non-existent versions, clearing when no override exists, invalid semver strings, etc.

### Test Plan Recommendation
1. Set up admin env vars locally
2. Deploy a test source-faker connector
3. Use MCP tools or Python API to get current version, set override to a different version, verify it's applied, then clear it
4. Check that version changes are actually reflected in the UI/API responses
5. Test with both sources and destinations

### Notes
- Implementation evolved through multiple iterations after discovering the `/list` endpoint doesn't scale well
- Field name bugs were discovered during development (e.g., `versionId` vs `actorDefinitionVersionId`)
- The scoped configuration system's primitives (resource_type, scope_type, etc.) were inferred from OpenAPI specs and platform code but never validated with actual API calls

**Session**: https://app.devin.ai/sessions/877297054b1443b59f8ab1b42ce9e699  
**Requested by**: AJ Steers (@aaronsteers)

## Summary by CodeRabbit

* **New Features**
  * View current connector versions for cloud sources and destinations.
  * Pin or clear connector versions with enforced minimum reason length and optional reference URL.
  * Admin-only cloud tools to set/clear overrides (gated by internal-admin flags and configured admin user).
  * User lookup by email/ID to support scoped override assignment.
  * Built-in permission validation enforcing business rules for override actions.

* **Chores**
  * Expanded Ruff lint ignore list.

### DIFF
--- a/airbyte/mcp/_tool_utils.py
+++ b/airbyte/mcp/_tool_utils.py
@@ -25,9 +25,10 @@ F = TypeVar("F", bound=Callable[..., Any])
 AIRBYTE_CLOUD_MCP_READONLY_MODE = (
     os.environ.get("AIRBYTE_CLOUD_MCP_READONLY_MODE", "").strip() == "1"
 )
-AIRBYTE_CLOUD_MCP_SAFE_MODE = os.environ.get("AIRBYTE_CLOUD_MCP_SAFE_MODE", "").strip() == "1"
+AIRBYTE_CLOUD_MCP_SAFE_MODE = os.environ.get("AIRBYTE_CLOUD_MCP_SAFE_MODE", "1").strip() != "0"
 
 _REGISTERED_TOOLS: list[tuple[Callable[..., Any], dict[str, Any]]] = []
+_GUIDS_CREATED_IN_SESSION: set[str] = set()
 
 
 class SafeModeError(Exception):
@@ -36,8 +37,34 @@ class SafeModeError(Exception):
     pass
 
 
+def register_guid_created_in_session(guid: str) -> None:
+    """Register a GUID as created in this session.
+
+    Args:
+        guid: The GUID to register
+    """
+    _GUIDS_CREATED_IN_SESSION.add(guid)
+
+
+def check_guid_created_in_session(guid: str) -> None:
+    """Check if a GUID was created in this session.
+
+    Raises SafeModeError if the GUID was not created in this session and
+    AIRBYTE_CLOUD_MCP_SAFE_MODE is set to 1.
+
+    Args:
+        guid: The GUID to check
+    """
+    if AIRBYTE_CLOUD_MCP_SAFE_MODE and guid not in _GUIDS_CREATED_IN_SESSION:
+        raise SafeModeError(
+            f"Cannot perform destructive operation on '{guid}': "
+            f"Object was not created in this session. "
+            f"AIRBYTE_CLOUD_MCP_SAFE_MODE is set to '1'."
+        )
+
+
 def should_register_tool(annotations: dict[str, Any]) -> bool:
-    """Check if a tool should be registered based on safe mode settings.
+    """Check if a tool should be registered based on mode settings.
 
     Args:
         annotations: Tool annotations dict containing domain, readOnlyHint, destructiveHint,
@@ -55,17 +82,9 @@ def should_register_tool(annotations: dict[str, Any]) -> bool:
     if annotations.get("domain") != "cloud":
         return True
 
-    if not AIRBYTE_CLOUD_MCP_READONLY_MODE and not AIRBYTE_CLOUD_MCP_SAFE_MODE:
-        return True
-
     if AIRBYTE_CLOUD_MCP_READONLY_MODE:
         is_readonly = annotations.get(READ_ONLY_HINT, False)
         if not is_readonly:
-            return False
-
-    if AIRBYTE_CLOUD_MCP_SAFE_MODE:
-        is_destructive = annotations.get(DESTRUCTIVE_HINT, True)  # Default is True per FastMCP
-        if is_destructive:
             return False
 
     return True
--- a/airbyte/mcp/cloud_ops.py
+++ b/airbyte/mcp/cloud_ops.py
@@ -19,7 +19,12 @@ from airbyte.cloud.connections import CloudConnection
 from airbyte.cloud.connectors import CloudDestination, CloudSource, CustomCloudSourceDefinition
 from airbyte.cloud.workspaces import CloudWorkspace
 from airbyte.destinations.util import get_noop_destination
-from airbyte.mcp._tool_utils import mcp_tool, register_tools
+from airbyte.mcp._tool_utils import (
+    check_guid_created_in_session,
+    mcp_tool,
+    register_guid_created_in_session,
+    register_tools,
+)
 from airbyte.mcp._util import resolve_config, resolve_list_of_strings
 
 
@@ -116,14 +121,14 @@ def deploy_source_to_cloud(
     try:
         source = get_source(
             source_connector_name,
-            install_if_missing=False,
+            no_executor=True,
         )
         config_dict = resolve_config(
             config=config,
             config_secret_name=config_secret_name,
             config_spec_jsonschema=source.config_spec,
         )
-        source.set_config(config_dict)
+        source.set_config(config_dict, validate=True)
 
         workspace: CloudWorkspace = _get_cloud_workspace()
         deployed_source = workspace.deploy_source(
@@ -135,6 +140,7 @@ def deploy_source_to_cloud(
     except Exception as ex:
         return f"Failed to deploy source '{source_name}': {ex}"
     else:
+        register_guid_created_in_session(deployed_source.connector_id)
         return (
             f"Successfully deployed source '{source_name}' with ID '{deployed_source.connector_id}'"
             f" and URL: {deployed_source.connector_url}"
@@ -186,14 +192,14 @@ def deploy_destination_to_cloud(
     try:
         destination = get_destination(
             destination_connector_name,
-            install_if_missing=False,
+            no_executor=True,
         )
         config_dict = resolve_config(
             config=config,
             config_secret_name=config_secret_name,
             config_spec_jsonschema=destination.config_spec,
         )
-        destination.set_config(config_dict)
+        destination.set_config(config_dict, validate=True)
 
         workspace: CloudWorkspace = _get_cloud_workspace()
         deployed_destination = workspace.deploy_destination(
@@ -205,6 +211,7 @@ def deploy_destination_to_cloud(
     except Exception as ex:
         return f"Failed to deploy destination '{destination_name}': {ex}"
     else:
+        register_guid_created_in_session(deployed_destination.connector_id)
         return (
             f"Successfully deployed destination '{destination_name}' "
             f"with ID: {deployed_destination.connector_id}"
@@ -266,6 +273,7 @@ def create_connection_on_cloud(
     except Exception as ex:
         return f"Failed to create connection '{connection_name}': {ex}"
     else:
+        register_guid_created_in_session(deployed_connection.connection_id)
         return (
             f"Successfully created connection '{connection_name}' "
             f"with ID '{deployed_connection.connection_id}' and "
@@ -381,6 +389,7 @@ def deploy_noop_destination_to_cloud(
     except Exception as ex:
         return f"Failed to deploy No-op Destination: {ex}"
     else:
+        register_guid_created_in_session(deployed_destination.connector_id)
         return (
             f"Successfully deployed No-op Destination "
             f"with ID '{deployed_destination.connector_id}' and "
@@ -656,6 +665,7 @@ def publish_custom_source_definition(
     except Exception as ex:
         return f"Failed to publish custom source definition '{name}': {ex}"
     else:
+        register_guid_created_in_session(custom_source.definition_id)
         return (
             "Successfully published custom YAML source definition:\n"
             + _get_custom_source_definition_description(
@@ -723,6 +733,7 @@ def update_custom_source_definition(
     Note: Only YAML (declarative) connectors are currently supported.
     Docker-based custom sources are not yet available.
     """
+    check_guid_created_in_session(definition_id)
     try:
         processed_manifest = manifest_yaml
         if isinstance(manifest_yaml, str) and "\n" not in manifest_yaml:
@@ -772,6 +783,7 @@ def permanently_delete_custom_source_definition(
     Note: Only YAML (declarative) connectors are currently supported.
     Docker-based custom sources are not yet available.
     """
+    check_guid_created_in_session(definition_id)
     workspace: CloudWorkspace = _get_cloud_workspace()
     definition = workspace.get_custom_source_definition(
         definition_id=definition_id,
@@ -1061,6 +1073,7 @@ def update_cloud_source_config(
     and `AIRBYTE_API_ROOT` environment variables will be used to authenticate with the
     Airbyte Cloud API.
     """
+    check_guid_created_in_session(source_id)
     workspace: CloudWorkspace = _get_cloud_workspace()
     source = workspace.get_source(source_id=source_id)
 
@@ -1136,6 +1149,7 @@ def update_cloud_destination_config(
     and `AIRBYTE_API_ROOT` environment variables will be used to authenticate with the
     Airbyte Cloud API.
     """
+    check_guid_created_in_session(destination_id)
     workspace: CloudWorkspace = _get_cloud_workspace()
     destination = workspace.get_destination(destination_id=destination_id)
 
@@ -1182,6 +1196,7 @@ def rename_cloud_connection(
 
 @mcp_tool(
     domain="cloud",
+    destructive=True,
     open_world=True,
 )
 def set_cloud_connection_table_prefix(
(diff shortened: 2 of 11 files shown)


airbyte/mcp/cloud_ops.py
```diff
@@ -681,6 +681,153 @@ def permanently_delete_custom_source_definition(
...
+) -> str:
+    """Set or clear a version override for a deployed source connector.
+
+    You must specify EXACTLY ONE of version OR clear_override=True, but not both.
+
+    By default, the `AIRBYTE_CLIENT_ID`, `AIRBYTE_CLIENT_SECRET`, `AIRBYTE_WORKSPACE_ID`,
+    and `AIRBYTE_API_ROOT` environment variables will be used to authenticate with the
+    Airbyte Cloud API.
+    """
+    try:
```
aaronsteers: Devin, don't use blanket try/catch. (This pattern is used elsewhere, I know, but I would rather fail hard - except for specific _handled_ or expected exceptions.)

Apply this feedback to all the new MCP tools in this PR, but don't apply it to other tools as of now.

airbyte/cloud/connectors.py
```diff
@@ -257,6 +351,100 @@ def _from_destination_response(
...
 
+    def get_connector_version(self) -> dict[str, Any]:
+        """Get the current version information for this destination.
+
+        Returns:
+            A dictionary containing version information including:
+            - dockerImageTag: The version string (e.g., "0.1.0")
+            - actorDefinitionId: The connector definition ID
+            - actorDefinitionVersionId: The specific version ID
+            - isOverrideApplied: Whether a version override is active
```
aaronsteers: Devin, instead of passing dict objects around, let's return an instance of a new basic Pydantic model `ConnectorVersionInfo` with just two elements:

1. version (str)
2. is_version_pinned (bool)

This keeps the surface area of what we need to support small, while giving a type-save means of interacting with the returned result.

This model can be declared here in this file. (It will not be known to the api_utils modules, which is focused on interfacing only with the internal API.

airbyte/_util/api_util.py
```diff
@@ -1244,3 +1244,243 @@ def get_connector_builder_project_for_definition_id(
...
+            "resource_id": actor_definition_id,
+            "scope_type": "actor",
+            "scope_id": connector_id,
+            "origin": "user",
+            "origin_type": "user",
+        },
+        api_root=api_root,
+        client_id=client_id,
+        client_secret=client_secret,
+    )
```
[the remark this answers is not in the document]
aaronsteers: Devin, please address this feedback. If we want to keep things symetrical, you can keep `connector_type` in the function signature and just add this to the top of the method implementation to quiet the linter:

```
_ = connector_type  # not used currently
```

airbyte/cloud/connectors.py
```diff
@@ -216,6 +216,100 @@ def _from_source_response(
...
+            api_root=self.workspace.api_root,
+            client_id=self.workspace.client_id,
+            client_secret=self.workspace.client_secret,
+        )
+
+    def set_connector_version_override(
+        self,
+        version: str | None = None,
+        *,
+        clear_override: bool = False,
```
aaronsteers: Devin maybe `unset` is a better bool arg input. E.g. set_connector_version_override(unset=True)

airbyte/cloud/connectors.py
```diff
@@ -257,6 +351,100 @@ def _from_destination_response(
...
+    ) -> dict[str, Any] | bool:
+        """Set or clear a version override for this destination.
+
+        You must specify EXACTLY ONE of version OR clear_override=True, but not both.
```
aaronsteers: Whenever referencing code snippets, args, class names, etc. in docstrings (except the 'args list'), you should be fencing them with backticks. This improves readability when docs are rendered by pdoc.

airbyte/cloud/connectors.py
```diff
@@ -84,6 +85,16 @@ def __repr__(self) -> str:
...
 
 
+class ConnectorVersionInfo(BaseModel):
+    """Information about a connector's version."""
+
+    version: str
+    """The version string (e.g., '0.1.0')."""
+
+    is_version_pinned: bool
+    """Whether a version override is active for this connector."""
```
aaronsteers: Devin, this looks great. Let's also add a `__str__` implementation that returns either `{version}` or `{version} (pinned)`. This can be used when messaging it to the user.

Also, let's just be really clear we're talking about cloud assets here - rename the class to `CloudConnectorVersionInfo`

airbyte/cloud/connectors.py
```diff
@@ -84,6 +85,22 @@ def __repr__(self) -> str:
...
+
+    def __str__(self) -> str:
+        """Return a string representation of the version."""
+        if self.is_version_pinned:
+            return f"{self.version} (pinned)"
+        return self.version
```
aaronsteers: Devin, ternary is more readable in this case.
I'll apply and you can pull for the latest.

```suggestion
        return self.version if not self.is_version_pinned else f"{self.version} (pinned)"
```

airbyte/mcp/cloud_ops.py
```diff
@@ -681,6 +681,146 @@ def permanently_delete_custom_source_definition(
...
+
+    By default, the `AIRBYTE_CLIENT_ID`, `AIRBYTE_CLIENT_SECRET`, `AIRBYTE_WORKSPACE_ID`,
+    and `AIRBYTE_API_ROOT` environment variables will be used to authenticate with the
+    Airbyte Cloud API.
+    """
+    workspace: CloudWorkspace = _get_cloud_workspace()
+    source = workspace.get_source(source_id=source_id)
+    version_info = source.get_connector_version()
+    return {
+        "version": version_info.version,
```
aaronsteers: Devin, just return str(version_info) or return version_info directly.

airbyte/mcp/cloud_ops.py
```diff
@@ -681,6 +681,146 @@ def permanently_delete_custom_source_definition(
...
+    and `AIRBYTE_API_ROOT` environment variables will be used to authenticate with the
+    Airbyte Cloud API.
+    """
+    workspace: CloudWorkspace = _get_cloud_workspace()
+    destination = workspace.get_destination(destination_id=destination_id)
+    version_info = destination.get_connector_version()
+    return {
+        "version": version_info.version,
+        "is_version_pinned": version_info.is_version_pinned,
+    }
```
aaronsteers: Ditto. Just return the object or stringify it.

airbyte/mcp/cloud_ops.py
```diff
@@ -26,6 +28,37 @@
 from airbyte.mcp._util import resolve_config, resolve_list_of_strings
 
 
+def _check_admin_access() -> str:
```
aaronsteers: Devin, rename '_check_admin_access' to '_check_internal_admin_flag'. Technically it is not a permission check - just a flag check.

airbyte/mcp/cloud_ops.py
```diff
@@ -26,6 +28,37 @@
 from airbyte.mcp._util import resolve_config, resolve_list_of_strings
 
 
+def _check_internal_admin_flag() -> str:
```
aaronsteers: Let's just return tuple[bool, str] and fail gracefully with a warning print and returning `false`.

aaronsteers: I mean, return tuple[bool, str | None]

aaronsteers: Actually no. Just decouple this. The check operation can return true or false after comparing both values, optionally printing warnings for invalid input.

But a separte internal helper function can return the email - calling the check function first and then independently reading the email var if 'bool' is returned.

airbyte/mcp/cloud_ops.py
```diff
@@ -686,6 +719,192 @@ def permanently_delete_custom_source_definition(
...
+        str | None,
+        Field(
+            description="Optional URL with more context (e.g., issue link).",
+            default=None,
+        ),
+    ] = None,
+) -> str:
+    """Set or clear a version override for a deployed source connector.
+
+    You must specify EXACTLY ONE of version OR unset=True, but not both.
```
aaronsteers: Devin, I thought we talked about this. Code snippets and refs should be in backticks. I'll apply the below but please check your docstrings for any other issues.

```suggestion
    You must specify EXACTLY ONE of `version` OR `unset=True`, but not both.
```

airbyte/mcp/cloud_ops.py
```diff
@@ -26,6 +27,53 @@
...
+        print(
+            "Warning: Admin user not configured. "
+            "Set AIRBYTE_INTERNAL_ADMIN_USER to your email to enable version override tools."
+        )
+        return False
+
+    return True
+
+
+def _get_admin_user_email() -> str:
```
aaronsteers:
```suggestion
def _get_admin_user_email() -> str | None:
```

Fail gracefully and return None. There's no failure condition not already caught in the check function.

airbyte/cloud/connectors.py
```diff
@@ -216,6 +233,129 @@ def _from_source_response(
...
+        )
+
+    def set_connector_version_override(
+        self,
+        version: str | None = None,
+        *,
+        unset: bool = False,
+        override_reason: str | None = None,
+        override_reason_reference_url: str | None = None,
+        user_email: str | None = None,
```
aaronsteers: Don't allow users to provide email here as input. It must be set in env var. Also, make this private ("_") prefix with a note that it is only for internal Airbyte admin use only. (Endpoints not enabled for end user access.)

You'll get linter warnings when calling it but just noqa them with a note "accessing non-public API".

airbyte/cloud/connectors.py
```diff
@@ -257,6 +397,129 @@ def _from_destination_response(
...
+    ) -> bool:
```
aaronsteers: Ditto.

airbyte/mcp/cloud_ops.py
```diff
@@ -22,6 +23,44 @@
...
 
 
+def _check_internal_admin_flag() -> bool:
+    """Check if internal admin flag is properly configured.
+
+    Returns:
+        True if both AIRBYTE_INTERNAL_ADMIN_FLAG and AIRBYTE_INTERNAL_ADMIN_USER are set correctly.
+    """
+    admin_flag = os.environ.get("AIRBYTE_INTERNAL_ADMIN_FLAG")
+    admin_user = os.environ.get("AIRBYTE_INTERNAL_ADMIN_USER")
```
aaronsteers, line 38: Add another internal guard statement here so that the admin flag much also be contained within the admin user string. Indirectly, this will prevent non-airbyte.io emails from leveraging this feature (unless they fork and/or change the code, of course).

airbyte/mcp/cloud_ops.py
```diff
@@ -747,6 +786,226 @@ def permanently_delete_custom_source_definition(
...
+        return f"No version override was set for source '{source_id}'"
+    return f"Successfully set version override to '{version}' for source '{source_id}'"
+
+
+@mcp_tool(
+    domain="cloud",
+    destructive=True,
+    idempotent=True,
+    open_world=True,
+)
```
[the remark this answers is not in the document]
aaronsteers: Devin, coderabbit is right. This isn't idempotent.

airbyte/mcp/_tool_utils.py
```diff
@@ -117,6 +127,7 @@ def list_sources():
         DESTRUCTIVE_HINT: destructive,
         IDEMPOTENT_HINT: idempotent,
         OPEN_WORLD_HINT: open_world,
+        "airbyte_internal": airbyte_internal,
```
aaronsteers: Make this string a constant also.

airbyte/mcp/cloud_ops.py
```diff
@@ -22,6 +23,44 @@
...
+        True if both AIRBYTE_INTERNAL_ADMIN_FLAG and AIRBYTE_INTERNAL_ADMIN_USER are set correctly.
+    """
+    admin_flag = os.environ.get("AIRBYTE_INTERNAL_ADMIN_FLAG")
+    admin_user = os.environ.get("AIRBYTE_INTERNAL_ADMIN_USER")
+
+    if admin_flag != "airbyte.io":
+        print(
+            "Warning: Admin access not enabled. "
+            "Set AIRBYTE_INTERNAL_ADMIN_FLAG=airbyte.io to enable version override tools."
+        )
```
aaronsteers, line 44: Your help text here and below is too transparent. We're not trying to guide users how to remedy the issue. We're trying to say there's an invalid value they should remove.

airbyte/mcp/_annotations.py
```diff
@@ -50,3 +50,13 @@
 
 FastMCP default if not specified: True
 """
+
+AIRBYTE_INTERNAL_HINT = "airbyte_internal"
```
aaronsteers: Rename to `AIRBYTE_INTERNAL`.

airbyte/mcp/_tool_utils.py
```diff
@@ -100,6 +109,8 @@ def mcp_tool(
         destructive: If True, tool modifies/deletes existing data (default: False)
         idempotent: If True, repeated calls have same effect (default: False)
         open_world: If True, tool interacts with external systems (default: False)
+        airbyte_internal: If True, tool is only for internal Airbyte admin use and requires
+            AIRBYTE_INTERNAL_ADMIN_FLAG and AIRBYTE_INTERNAL_ADMIN_USER env vars (default: False)
```
aaronsteers: Devin, fewer suggestions = more helpful. Internal means internal.

```suggestion
        airbyte_internal: If True, tool is only for internal Airbyte admin use.
```

aaronsteers:
Devin, resolve conflicts and update with latest design patterns from main.

aaronsteers:
@clnoll, @katmarkham - I feel good about the additional protections added here since we first discussed.

@clnoll - Can you add additional context here regarding what additional business logic protections should be added? You mentioned logic around RC logic and (I think) not overriding other pinned sets. Can you say more about what business logic we should apply?

(Devin replies disabled so we can talk without spammy interjections.)

Thanks!

cc @frifriSF59

clnoll:
@aaronsteers the main logic that I'm concerned with is making sure that we can't unpin or pin over a pin that already exists without "good reason". I'm not sure exactly how you envision this being used, but if it's for pinning pre-release connector versions that would mean only allowing it to pin/unpin versions with the `-dev` suffix, and even then maybe only if the user doing the pinning/unpinning was the one that created it (although that might be too strict for `-dev` pins). And likewise if it's for pinning release candidates, making sure that it only pins/unpins connectors with the `-rc` suffix. But most importantly, we should never pin/unpin a prod version of a connector unless we know for sure that's what we want to do, either because a customer requested it or support needs to in order to investigate/fix something.

airbyte/cloud/connectors.py
```diff
@@ -84,6 +87,116 @@ def __repr__(self) -> str:
...
+
+def _validate_version_override_permission(
+    *,
+    current_version: str,
+    new_version: str | None,
+    is_setting_override: bool,
+    override_reason: str,
+    user_email: str,
+    existing_override_creator_email: str | None = None,
+) -> tuple[bool, str]:
```
aaronsteers: Let's move this into a new submodule: `airbyte._util.biz_logic`. That is private already by being under _util, and it can be a public helper function (no prefix) from that module.

airbyte/_util/biz_logic.py
```diff
@@ -0,0 +1,104 @@
...
+def validate_version_override_permission(
+    *,
+    current_version: str,
+    new_version: str | None,
+    is_setting_override: bool,
+    override_reason: str,
+    user_email: str,
+    existing_override_creator_email: str | None = None,
+) -> tuple[bool, str]:
+    """Validate whether a version override operation should be permitted.
```
aaronsteers, line 20: @clnoll - Not sure if this is _actually_ the right logic but I've created this helper function in a new `_util.biz_logic` module, so we can iterate here on protection rules without needing to be bogged down with the technical implementation and business logic being intermixed.

aaronsteers: Re feedback from [this comment](https://github.com/airbytehq/PyAirbyte/pull/838#issuecomment-3442643047):

> @aaronsteers the main logic that I'm concerned with is making sure that we can't unpin or pin over a pin that already exists without "good reason". I'm not sure exactly how you envision this being used, but if it's for pinning pre-release connector versions that would mean only allowing it to pin/unpin versions with the `-dev` suffix, and even then maybe only if the user doing the pinning/unpinning was the one that created it (although that might be too strict for `-dev` pins). And likewise if it's for pinning release candidates, making sure that it only pins/unpins connectors with the `-rc` suffix. But most importantly, we should never pin/unpin a prod version of a connector unless we know for sure that's what we want to do, either because a customer requested it or support needs to in order to investigate/fix something.

aaronsteers: In regards to the use cases, I'm actually not the best person to speak to what those are - since I rarely work in this space myself.

aaronsteers:
Moved to internal `airbyte-ops-mcp`

Issue #850, opened 2025-11-04
🐛 Bug: Coral MCP tries to install connectors before deploying a coral cloud destination


aaronsteers:
Coral MCP tries to install connectors before deploying a Cloud destination. This can fail if the connector can't be installed, for instance if the Python version is a mismatch with the connector, or if PyAirbyte doesn't have a writeable directory.

## Workarounds

First workaround is to ensure that the `AIRBYTE_PROJECT_DIR` env var is set as described here: https://docs.airbyte.com/ai-agents/pyairbyte-mcp#step-1-generate-a-dotenv-secrets-file

Second workaround is to use a CLI arg like `--python=3.11` to ensure a compatible Python version is being used. Example here: https://docs.airbyte.com/ai-agents/pyairbyte-mcp#step-2-registering-the-mcp-server

Third workaround is to try to force Docker mode instead of Python. (Not well documented.)

## Proposed Fixes:

The best fix is to skip installation when we are only creating the source or destination in order to deploy it to Cloud. (This _might_ be implemented already for sources, and perhaps just destinations are not patched to leverage this pattern?)

Another fix option is to force Docker mode when we are running as MCP. I had thought this was already implemented, but the setting may not be correctly applied across both sources and destination use cases.

Docs: https://airbytehq.github.io/PyAirbyte/airbyte/mcp.html

PR #851, opened 2025-11-04
fix: Skip local validation when deploying connectors to Cloud via MCP

aaronsteers:
# fix: Skip local validation when deploying connectors to Cloud via MCP

## Summary
Fixes [#850](https://github.com/airbytehq/PyAirbyte/issues/850) - MCP was trying to install connectors locally before deploying them to Cloud, which failed when Python/Docker wasn't available or the connector couldn't be installed.

The root cause: `deploy_source_to_cloud()` and `deploy_destination_to_cloud()` were calling `.config_spec` which triggers local connector execution via `_execute()`, requiring the connector to be installed locally. 

The fix skips local validation entirely by:
1. Passing `config_spec_jsonschema=None` to `resolve_config()` - skips hardcoded secrets detection
2. Passing `validate=False` to `set_config()` - skips config schema validation

Since `CloudWorkspace.deploy_source/destination()` only accesses `.name` and `._hydrated_config`, no local execution is needed. All validation now happens server-side via the Cloud API.

- [ ] **End-to-end test**: Deploy a source and destination to Cloud via MCP in an environment without Python/Docker to verify the original issue is resolved
- [ ] **Validation fallback**: Confirm that invalid configs are properly caught and reported by Cloud API validation (since local validation is now skipped)
- [ ] **Edge case review**: Check if any other MCP functions might have similar local execution requirements that weren't addressed

### Test Plan
1. Set up MCP environment without local Python/Docker or where connectors can't be installed
2. Use the `deploy_source_to_cloud` and `deploy_destination_to_cloud` MCP tools to deploy connectors
3. Verify deployment succeeds without installation errors
4. Try with both valid and invalid configs to confirm Cloud validation works

### Notes
- This PR intentionally shifts validation from local (client-side) to Cloud (server-side)
- The `install_if_missing=False` parameter was already present but insufficient because `.config_spec` still triggered local execution
- Alternative approaches considered but rejected: NoOp executor (would require broader changes), propagating `install_if_missing` through all executors (too broad for this fix)

---
**Link to Devin run**: https://app.devin.ai/sessions/d1818e9eff6f4b0ebd7b0cb0b9b85865  
**Requested by**: AJ Steers (@aaronsteers)

## Summary by CodeRabbit

* **New Features**
  * Validate connector configs against registry specs.
  * "No-executor" mode to operate/inspect connectors without installing them.
  * Registry-based connector spec fetching utility.

* **Bug Fixes**
  * Cloud deployments now validate configs against registry specs to catch invalid configs earlier.

* **Changes**
  * Connector listings now default to showing Docker-installable connectors.

* **Tests**
  * Added integration tests for registry spec fetching and config validation.

### DIFF
--- a/tests/integration_tests/test_registry_spec.py
+++ b/tests/integration_tests/test_registry_spec.py
@@ -12,18 +12,25 @@ from airbyte._util.registry_spec import (
 
 
 @pytest.mark.parametrize(
-    "connector_name,platform",
+    "connector_name,platform,version",
     [
-        ("source-faker", "oss"),
-        ("source-faker", "cloud"),
-        ("destination-duckdb", "oss"),
+        ("source-faker", "oss", None),
+        ("source-faker", "cloud", None),
+        ("destination-duckdb", "oss", None),
+        ("source-faker", "oss", "6.2.0"),
+        ("source-faker", "cloud", "6.2.0"),
     ],
 )
-def test_get_connector_spec_from_registry(connector_name: str, platform: str) -> None:
+def test_get_connector_spec_from_registry(
+    connector_name: str,
+    platform: str,
+    version: str | None,
+) -> None:
     """Test fetching connector specs from the registry."""
     spec = get_connector_spec_from_registry(
         connector_name,
         platform=platform,
+        version=version,
     )
 
     assert spec is not None


airbyte/mcp/connector_registry.py
```diff
@@ -70,7 +70,8 @@ def list_connectors(
     Returns:
         List of connector names.
     """
-    connectors: list[str] = get_available_connectors()
+    # Start with the full list of known connectors (all support Docker):
+    connectors: list[str] = get_available_connectors(install_type=InstallType.DOCKER)
```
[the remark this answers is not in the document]
aaronsteers: Disagreed with path forward and business logic. Caller can ask for a specific install type - if and only if they actually want to install it - but the other use case is to deploy to Cloud. We don't want users without docker to be constrained on what they can deploy to Cloud, since Cloud supports all of them. "Docker" is indeed used as a standin for "all connectors".

airbyte/destinations/util.py
```diff
@@ -84,7 +94,13 @@ def get_destination(  # noqa: PLR0913 # Too many arguments
             use_host_network=use_host_network,
             install_if_missing=install_if_missing,
             install_root=install_root,
-        ),
+        )
```
aaronsteers: Devin, Move this logic to be internal to get_connector_executor() so get_connector_executor() is also capable of returning a NoOpExecutor.

airbyte/sources/util.py
```diff
@@ -111,13 +114,19 @@ def get_source(  # noqa: PLR0913 # Too many arguments
...
-        executor=get_connector_executor(
+    if no_executor:
+        metadata = get_connector_metadata(name)
+        executor = NoOpExecutor(
+            name=name,
+            metadata=metadata,
+            target_version=version,
+        )
+    else:
+        executor = get_connector_executor(
```
aaronsteers: Ditto. This logic can live inside get_connector_executor()

airbyte/_util/registry_spec.py
aaronsteers, line 1: Add very few, very simple integration tests for this module's helper functions. Use pytest.parametrize to get multiple test cases per test.

airbyte/_executors/util.py
```diff
@@ -167,11 +167,24 @@ def get_connector_executor(  # noqa: PLR0912, PLR0913, PLR0914, PLR0915, C901 #
...
+    no_executor: bool = False,
 ) -> Executor:
     """This factory function creates an executor for a connector.
 
     For documentation of each arg, see the function `airbyte.sources.util.get_source()`.
     """
+    if no_executor:
+        from airbyte._executors.noop import (  # noqa: PLC0415 # Local import to avoid cycles
+            NoOpExecutor,
+        )
```
aaronsteers: Don't ever do this unless you are given explicit permission. Always put imports to the top of the file.

airbyte/_executors/util.py
aaronsteers, line 185: Devin, you should include `no_executor` in this evaluator for "no more than one install method count". Which means moving your handling of the no op case down below to be inline with the other methods.

airbyte/destinations/util.py
```diff
@@ -69,22 +70,28 @@ def get_destination(  # noqa: PLR0913 # Too many arguments
...
+        version=version,
+        use_python=use_python,
+        pip_url=pip_url,
+        local_executable=local_executable,
+        docker_image=docker_image,
+        use_host_network=use_host_network,
+        install_if_missing=install_if_missing,
+        install_root=install_root,
+        no_executor=no_executor,
+    )
```
aaronsteers, line 88: Devin, you can move this back inline below as it was before.

airbyte/sources/util.py
```diff
@@ -111,24 +112,30 @@ def get_source(  # noqa: PLR0913 # Too many arguments
...
+        use_python=use_python,
+        pip_url=pip_url,
+        local_executable=local_executable,
+        docker_image=docker_image,
+        use_host_network=use_host_network,
+        source_manifest=source_manifest,
+        install_if_missing=install_if_missing,
+        install_root=install_root,
+        no_executor=no_executor,
+    )
```
aaronsteers, line 131: Ditto. You can move this expression back inline as it was before.

tests/integration_tests/test_registry_spec.py
```diff
@@ -0,0 +1,81 @@
...
+
+@pytest.mark.parametrize(
+    "connector_name,platform",
+    [
+        ("source-faker", "oss"),
+        ("source-faker", "cloud"),
+        ("destination-duckdb", "oss"),
+    ],
+)
+def test_get_connector_spec_from_registry(connector_name: str, platform: str) -> None:
```
aaronsteers: Add some instances here that leverage an optional version (instead of defaulting to latest). If needed, add the necessary logic upstream to make version an optional input.

tests/integration_tests/test_registry_spec.py
```diff
@@ -0,0 +1,81 @@
...
+            },
+            False,
+        ),
+    ],
+)
+def test_validate_connector_config_from_registry(
+    connector_name: str,
+    config: dict,
+    expected_valid: bool,
+) -> None:
```
aaronsteers, line 74: Ditto. Add one or two entries that use an explicit version instead of the default 'latest' version.

airbyte/_executors/noop.py
```diff
@@ -0,0 +1,136 @@
...
+                "use a Docker or Python executor. "
+                "For MCP cloud deployments, set pre_validate_config=False "
+                "to skip local validation."
+            ),
+        )
+
+    def ensure_installation(self, *, auto_fix: bool = True) -> None:
+        """No-op: NoOpExecutor doesn't require installation."""
+        _ = auto_fix  # Unused
+        pass
```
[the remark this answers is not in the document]
aaronsteers: aside (Devin-ignore)

I disagree with this. Removing the `pass` makes it look like the code has a bug. Removing the prev `_ = ` assignment likewise would likely (?) trigger the linters.

PR #858, opened 2025-11-11
feat(mcp): Add get_api_docs_urls tool with data.externalDocumentationUrls support

aaronsteers:
# refactor(registry): Move manifest parsing logic from MCP to registry module

## Summary
Refactored the manifest parsing and API docs URL extraction logic from the MCP module (`airbyte/mcp/connector_registry.py`) into the core registry module (`airbyte/registry.py`) to improve separation of concerns and code organization.

**Key changes:**
- Moved `ApiDocsUrl` class, helper functions (`_manifest_url_for`, `_fetch_manifest_dict`, `_extract_docs_from_registry`), and created new public function `get_connector_api_docs_urls` in `airbyte/registry.py`
- Simplified MCP `get_api_docs_urls` tool to be a thin wrapper that calls `get_connector_api_docs_urls` from registry
- Updated `_extract_docs_from_registry` to use `_get_registry_url()` for consistency and include `documentationUrl` from registry metadata
- Added `_DEFAULT_MANIFEST_URL` constant to `registry.py` to avoid circular import with `_executors.util`
- Updated unit tests to import from registry module with proper mocking

## Review & Testing Checklist for Human
- [ ] **Circular import fix**: Verify that duplicating the `_DEFAULT_MANIFEST_URL` constant in `registry.py` (instead of importing from `_executors.util`) is acceptable. This was necessary to avoid a circular import since `_executors.util` imports from `registry`.
- [ ] **End-to-end testing**: Test the MCP `get_api_docs_urls` tool with a real connector (e.g., `source-faker`) to ensure the refactored code works correctly and returns the expected documentation URLs.
- [ ] **API design**: Verify that importing the private constant `_DEFAULT_MANIFEST_URL` from `registry.py` in the MCP module is acceptable, or if it should be made public.
- [ ] **Deduplication logic**: Confirm that URL deduplication still works correctly after the refactoring (test with a connector that has duplicate URLs across registry and manifest sources).

### Notes
- All 8 unit tests passing locally
- All lint checks passing (`poe fix-and-check`)
- The refactoring maintains backward compatibility for the MCP tool API - it still returns the same data structure
- CI is currently running on the latest commit

**Requested by:** AJ Steers (aj@airbyte.io) / @aaronsteers  
**Session:** https://app.devin.ai/sessions/02ae774e5a3a4052b5260e136bec5ea0

## Summary by CodeRabbit

* **New Features**
[...]

PR #863, opened 2025-11-12
feat(registry): Add KeyError validation for externalDocumentationUrls

aaronsteers:
# feat(registry): Add KeyError validation for externalDocumentationUrls

## Summary
Adds validation to raise clear `PyAirbyteInputError` exceptions when `externalDocumentationUrls` entries are missing required `title` or `url` fields, instead of allowing KeyError to propagate. This applies to both manifest.yaml parsing and registry metadata parsing.

**Changes:**
- Modified `ApiDocsUrl.from_manifest_dict()` to wrap dict access in try-except and raise `PyAirbyteInputError` with clear message
- Modified `_extract_docs_from_registry()` with same validation pattern
- Added 2 unit tests verifying exception behavior for missing fields
- Updated docstrings to document the new exception

**Related PRs:**
- airbytehq/PyAirbyte#858 (merged) - Initial implementation of get_api_docs_urls tool

This is a **yellow risk** change (behavior change with potential production impact):

- [ ] **Verify this won't break existing connectors**: Check if any connectors in production have malformed `externalDocumentationUrls` metadata (missing title/url fields). This change will cause those to fail instead of being silently skipped.
- [ ] **Test with source-faker**: Manually test `get_api_docs_urls("source-faker")` to ensure it still works correctly with valid metadata
- [ ] **Consider error message clarity**: The error messages are simple ("missing required field 'title'") - verify this provides enough context for debugging. May want to include connector name or manifest URL in future iteration.

### Notes
- This is a follow-up to PR #858 based on CodeRabbit's suggestion and user feedback
- The change converts silent failures (KeyError) into explicit exceptions with clear messages
- Test coverage: Added tests for manifest parsing path, but registry parsing path uses same pattern and isn't explicitly tested for error cases (relies on same code path)
- Devin session: https://app.devin.ai/sessions/02ae774e5a3a4052b5260e136bec5ea0

## Summary by CodeRabbit

* **New Features**
  * Documentation ingestion now reads connector metadata (metadata.yaml) in addition to registry entries.

* **Bug Fixes**
  * Stronger validation with clearer errors for missing documentation fields.
  * Connector presence is validated before fetching docs; registry and metadata docs are returned together (duplicates may appear).

* **Tests**
  * Added tests ensuring errors for incomplete or malformed metadata.

* **Chores**
  * Renamed public function for retrieving connector documentation URLs for clearer intent.

### DIFF
--- a/airbyte/registry.py
+++ b/airbyte/registry.py
@@ -40,8 +40,8 @@ _MANIFEST_ONLY_LANGUAGE = "manifest-only"
 _PYTHON_LANGUAGE_TAG = f"language:{_PYTHON_LANGUAGE}"
 _MANIFEST_ONLY_TAG = f"language:{_MANIFEST_ONLY_LANGUAGE}"
 
-_DEFAULT_MANIFEST_URL = (
-    "https://connectors.airbyte.com/files/metadata/airbyte/{source_name}/{version}/manifest.yaml"
+_DEFAULT_METADATA_URL = (
+    "https://connectors.airbyte.com/files/metadata/airbyte/{source_name}/{version}/metadata.yaml"
 )
 
 
@@ -312,13 +312,15 @@ class ApiDocsUrl(BaseModel):
     model_config = {"populate_by_name": True}
 
     @classmethod
-    def from_docs_list(cls, docs: list[dict[str, Any]], *, source: str, context: str) -> list[Self]:
-        """Extract documentation URLs from a list of documentation dictionaries.
+    def from_metadata_docs_list(
+        cls, docs: list[dict[str, Any]], *, source: str, context: str
+    ) -> list[Self]:
+        """Extract documentation URLs from a list of metadata documentation dictionaries.
 
         Args:
             docs: List of documentation dictionaries with 'title' and 'url' fields
             source: The source identifier for these documentation URLs
-            context: Context string for error messages (e.g., "Manifest", "Registry")
+            context: Context string for error messages (e.g., "Metadata", "Registry")
 
         Returns:
             List of ApiDocsUrl objects extracted from the docs list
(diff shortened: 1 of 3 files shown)


airbyte/registry.py
```diff
@@ -320,23 +320,31 @@ def from_manifest_dict(cls, manifest_data: dict[str, Any]) -> list[Self]:
...
+                                title=doc["title"],
+                                url=doc["url"],
+                                source="data_external_docs",
+                                doc_type=doc.get("type", "other"),
+                                requires_login=doc.get("requiresLogin", False),
+                            )
+                        )
+                    except KeyError as e:
+                        raise exc.PyAirbyteInputError(
+                            message=f"Manifest parsing error: missing required field {e}"
```
aaronsteers: Devin, let's update to something like `missing required field in {doc}: {e}`

so the user can see the dict that failed validation

airbyte/registry.py
```diff
@@ -386,6 +394,9 @@ def _extract_docs_from_registry(connector_name: str) -> list[ApiDocsUrl]:
 
     Returns:
         List of ApiDocsUrl objects extracted from the registry
+
+    Raises:
+        PyAirbyteInputError: If a documentation entry is missing required 'title' or 'url' field
```
aaronsteers: Devin, this is the wrong exception. Input error is if the user provides a bad input. This can be PyAirbyteInternalError or something else - which signified an issue the user will not be able to resolve themselves.

airbyte/registry.py
```diff
@@ -407,29 +418,33 @@ def _extract_docs_from_registry(connector_name: str) -> list[ApiDocsUrl]:
...
+                try:
+                    docs_urls.append(
+                        ApiDocsUrl(
+                            title=doc["title"],
+                            url=doc["url"],
+                            source="registry_external_docs",
+                            doc_type=doc.get("type", "other"),
+                            requires_login=doc.get("requiresLogin", False),
+                        )
                     )
```
aaronsteers: Devin, this isn't DRY. You should be calling the list factor class member on the class.

airbyte/registry.py
```diff
@@ -312,34 +312,62 @@ class ApiDocsUrl(BaseModel):
     model_config = {"populate_by_name": True}
 
     @classmethod
-    def from_manifest_dict(cls, manifest_data: dict[str, Any]) -> list[Self]:
-        """Extract documentation URLs from parsed manifest data.
+    def from_docs_list(cls, docs: list[dict[str, Any]], *, source: str, context: str) -> list[Self]:
```
aaronsteers: Rename to _from_manifest_docs_list()

PR #883, opened 2025-12-01
feat(mcp): add list_cloud_workspaces and describe_cloud_organization tools

aaronsteers:
# feat(mcp): add list_cloud_workspaces and describe_cloud_organization tools

## Summary

Adds two new MCP tools for organization-scoped workspace discovery, addressing the safety concern that LLM agents could accidentally access workspaces across organizations when using instance admin credentials.

**New MCP Tools:**
- `list_cloud_workspaces`: Lists workspaces within a specific organization, requiring either `organization_id` or `organization_name` (exact match). Supports `name_contains` filtering and `max_items_limit`.
- `describe_cloud_organization`: Gets details about a specific organization, useful for looking up org ID from name or vice versa.

**New api_util functions:**
- `list_organizations_for_user`: Uses public API `GET /organizations` to list orgs accessible to current user
- `list_workspaces_in_organization`: Uses internal Config API `POST /v1/workspaces/list_by_organization_id` with server-side name filtering and pagination

**Design decisions:**
- No `list_organizations` tool is exposed to prevent agents from enumerating orgs and picking "close matches"
- Organization name matching is case-sensitive and exact-match only
- Both tools require explicit organization context (ID or exact name)

- [ ] **Verify Config API endpoint**: The workspace listing uses internal Config API (`/v1/workspaces/list_by_organization_id`). Confirm this endpoint exists and the request/response format matches the implementation.
- [ ] **Test with real credentials**: These tools have not been tested against a real Airbyte Cloud instance. Test both tools with valid credentials to verify they work end-to-end.
- [ ] **Verify pagination**: Test with an organization that has many workspaces (>100) to confirm pagination works correctly.
- [ ] **Check workspace response fields**: The code accesses `workspaceId`, `name`, `organizationId` from the API response via `.get()`. Verify these field names match the actual API response.

**Recommended test plan:**
1. Set up MCP server with valid Airbyte Cloud credentials
2. Call `describe_cloud_organization` with a known org name → verify it returns correct org ID
3. Call `list_cloud_workspaces` with that org ID → verify it returns expected workspaces
4. Call `list_cloud_workspaces` with `name_contains` filter → verify filtering works
5. Test error cases: invalid org name, missing parameters, etc.

### Notes

- Link to Devin run: https://app.devin.ai/sessions/d5a1a4bb82b14064b29b4223d16668df
- Context: Discussion about preventing MCP agents from "going crazy" listing workspaces when using instance admin credentials

## Summary by CodeRabbit

* **New Features**
  * Add API utilities to list a user’s organizations and to enumerate workspaces within an organization, with optional name filtering, pagination and item limits.
  * Add cloud organization lookup by ID or exact name with clear not-found and ambiguity handling.
  * Expose cloud workspace listing for organizations to support discovery and tooling integration.

### DIFF
--- a/airbyte/_util/api_util.py
+++ b/airbyte/_util/api_util.py
@@ -1643,8 +1643,7 @@ def list_workspaces_in_organization(
         payload["nameContains"] = name_contains
 
     # Fetch pages until we have all results or reach the limit
-    has_more_pages = True
-    while has_more_pages:
+    while True:
         json_result = _make_config_api_request(
             path="/workspaces/list_by_organization_id",
             json=payload,
@@ -1654,14 +1653,20 @@ def list_workspaces_in_organization(
         )
 
         workspaces = json_result.get("workspaces", [])
+
+        # If no results returned, we've exhausted all pages
+        if not workspaces:
+            break
+
         result.extend(workspaces)
 
         # Check if we've reached the limit
         if max_items_limit is not None and len(result) >= max_items_limit:
             return result[:max_items_limit]
 
-        # Check if there are more pages
-        has_more_pages = len(workspaces) >= page_size
+        # If we got fewer results than page_size, this was the last page
+        if len(workspaces) < page_size:
+            break
 
         # Bump offset for next iteration
         payload["pagination"]["rowOffset"] += page_size


airbyte/_util/api_util.py
```diff
@@ -1562,3 +1562,107 @@ def get_connector_builder_project_for_definition_id(
...
+    if status_ok(response.status_code) and response.organizations_response:
+        return response.organizations_response.data
+
+    raise AirbyteError(
+        message="Failed to list organizations for user.",
+        context={
+            "status_code": response.status_code,
+            "response": response,
+        },
+    )
```
aaronsteers, line 1604: Devin, we need pagination handling, yes?

airbyte/_util/api_util.py
```diff
@@ -1562,3 +1562,107 @@ def get_connector_builder_project_for_definition_id(
...
+        max_items_limit: Optional maximum number of workspaces to return
+
+    Returns:
+        List of workspace dictionaries containing workspaceId, organizationId, name, slug, etc.
+    """
+    result: list[dict[str, Any]] = []
+    page_size = 100
+    offset = 0
+
+    while True:
```
aaronsteers: Can you write this to not use "while True"?

You can have a base payload which you set upfront and bump rowOffset at the bottom/end of each iteration.

airbyte/mcp/cloud_ops.py
```diff
@@ -852,6 +876,227 @@ def list_deployed_cloud_connections(
...
+                    "message": f"No organization found with exact name '{organization_name}' "
+                    "for the current user.",
+                },
+            )
+        if len(matching_orgs) > 1:
+            raise PyAirbyteInputError(
+                message=f"Multiple organizations found with name '{organization_name}'. "
+                "Please use 'organization_id' instead to specify the exact organization."
+            )
+        org = matching_orgs[0]
```
aaronsteers: Shouldn't this just call _resolve_organization_id()?

airbyte/_util/api_util.py
```diff
@@ -1562,3 +1562,108 @@ def get_connector_builder_project_for_definition_id(
...
+
+        workspaces = json_result.get("workspaces", [])
+        result.extend(workspaces)
+
+        # Check if we've reached the limit
+        if max_items_limit is not None and len(result) >= max_items_limit:
+            return result[:max_items_limit]
+
+        # Check if there are more pages
+        has_more_pages = len(workspaces) >= page_size
```
aaronsteers: Devin, I think there's a 'next' or something. Attempting to pull a page after end of records (if num records is a factor of page size) may result in an error. Check other API examples and/or the OpenAPI spec for the API itself in order to understand what the proper signal is.

PR #898, opened 2025-12-09
fix(cloud): Handle corrupted resources gracefully in list operations

aaronsteers:
## Summary

When a workspace contains corrupted resources (e.g., destinations with references to deleted secrets), the Airbyte API returns a 500 error during list operations. Previously, this caused `list_connections`, `list_sources`, and `list_destinations` to fail entirely, breaking higher-level operations like `deploy_destination` (which uses `list_destinations` for uniqueness checks).

This PR adds targeted error handling that detects the specific "Secret reference does not exist but is referenced in the config" error pattern and returns an empty list with a warning instead of raising an exception. Other API errors are still raised normally.

**Context:** This was discovered during CI investigation where `test_deploy_destination` failed intermittently due to a corrupted destination in the test workspace.

- [ ] **Verify the empty list behavior is acceptable** - When corrupted resources exist, callers will receive an empty list. For `deploy_destination`, this means the uniqueness check will pass (potentially allowing duplicates). Consider if this is the desired behavior or if a different approach is needed.
- [ ] **Verify error pattern matching** - The detection relies on substring matching (`"Secret reference"` AND `"does not exist but is referenced in the config"`). Confirm this matches the actual error format from the API.
- [ ] **Consider adding unit tests** - No unit tests were added for the new error handling. Consider whether tests should be added to lock in this behavior.

**Suggested test plan:**
1. Run the integration tests to verify existing functionality still works
2. If possible, manually test with a workspace containing a corrupted resource to verify the warning is printed and an empty list is returned

### Notes

- Link to Devin run: https://app.devin.ai/sessions/75875e7062834b71ade40f34d4722d2e
- Requested by: AJ Steers (aj@airbyte.io) / @aaronsteers

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace listing operations now detect corrupted workspace resources and, when encountered, log a warning and return an empty list instead of causing the listing call to fail.

* **Tests**
  * Added unit tests covering detection of corrupted-resource errors and verifying listing behavior for both corrupted and unrelated errors.

### DIFF
--- a/tests/unit_tests/test_cloud_workspace_error_handling.py
+++ b/tests/unit_tests/test_cloud_workspace_error_handling.py
@@ -1,0 +1,154 @@
+# Copyright (c) 2024 Airbyte, Inc., all rights reserved.
+"""Unit tests for CloudWorkspace error handling in list operations."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from airbyte_api.errors import SDKError
+
+from airbyte.cloud.workspaces import _is_corrupted_resource_error
+
+
+@pytest.mark.parametrize(
+    "error_body,expected",
+    [
+        pytest.param(
+            '{"message":"Secret reference 0c646a6d-8cfa-4c97-9615-86d8ad3bbaf8 '
+            'does not exist but is referenced in the config"}',
+            True,
+            id="secret_reference_error",
+        ),
+        pytest.param(
+            '{"message":"Internal server error"}',
+            False,
+            id="unrelated_500_error",
+        ),
+        pytest.param(
+            '{"message":"Secret reference abc123"}',
+            False,
+            id="partial_match_secret_only",
+        ),
+        pytest.param(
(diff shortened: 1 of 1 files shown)


airbyte/cloud/workspaces.py
```diff
@@ -454,16 +475,30 @@ def list_connections(
...
+        try:
+            connections = api_util.list_connections(
+                api_root=self.api_root,
+                workspace_id=self.workspace_id,
+                name=name,
+                name_filter=name_filter,
+                client_id=self.client_id,
+                client_secret=self.client_secret,
+            )
+        except SDKError as ex:
```
aaronsteers, line 492: Wrong. Needs to be handled on the inner loop so all valid objects are still returned. It is not sufficient to explain why we are failing. We need to not fail. Try again from scratch.

aaronsteers:
Closing. No good solution available. If thesis is correct that the 'list' endpoint implodes if anything within it is corrupted, then an upstream fix is needed.

PR #905, opened 2025-12-10
fix(mcp): The check workspace tool should not fail when missing organization permissions

aaronsteers:
## Summary

Fixes a regression introduced in v0.33.7 where `check_airbyte_cloud_workspace` fails with a 403 error for users with workspace-scoped credentials that lack ORGANIZATION_READER permissions.

**Approach:** Per AJ's request, this refactors the organization info handling upstream to `CloudWorkspace` rather than just catching exceptions in the MCP tool:

1. **New `CloudOrganization` dataclass** - Minimal value object with `organization_id` and `organization_name`
2. **New `get_organization(raise_on_error: bool = True)` method** - With overloads for proper type inference:
   - `get_organization()` → `CloudOrganization` (non-nullable)
   - `get_organization(raise_on_error=False)` → `CloudOrganization | None`
3. **Removed `organization_id` and `organization_name` properties** - ⚠️ Breaking change
4. **Validation** - Both `organization_id` and `organization_name` must be non-null and non-empty; otherwise treated as an error

**Users affected by the original issue can pin to `v0.33.6` as a workaround.**

- [ ] **Breaking change impact**: Verify no external code depends on `CloudWorkspace.organization_id` or `CloudWorkspace.organization_name` properties
- [ ] **Exception type**: Confirm `AirbyteError` is raised for 403 errors (verified at `airbyte/_util/api_util.py` line ~1206, but worth double-checking)
- [ ] **Validation strictness**: Confirm requiring BOTH `organization_id` AND `organization_name` to be non-empty is the desired behavior
- [ ] **Test with limited permissions**: Use a workspace with workspace-scoped credentials (no ORGANIZATION_READER) to confirm the tool returns the placeholder message instead of failing

**Recommended test plan:**
1. Create/use an API key with only workspace-level permissions (no org-level access)
2. Call `check_airbyte_cloud_workspace` and verify it returns successfully with `organization_id = "[unavailable - requires ORGANIZATION_READER permission]"`

### Notes

- Requested by AJ Steers (@aaronsteers) via Slack thread in #hydra-pyairbyte-mcp
- Link to Devin run: https://app.devin.ai/sessions/232375763cf84cb588a415821bbb9739
- Original issue reported by Ryan Waskewich

## Summary by CodeRabbit

* **New Features**
  * Added a new organization lookup API and value object to expose organization ID and optional name via a single call.

* **Bug Fixes**
  * Workspace verification now tolerates missing organization info (e.g., insufficient permissions) and returns clear placeholders instead of failing.

* **Breaking Changes**
  * Direct organization properties replaced by the new organization lookup API; callers should use the new API to obtain organization details.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

### DIFF
--- a/airbyte/cloud/workspaces.py
+++ b/airbyte/cloud/workspaces.py
@@ -150,7 +150,7 @@ class CloudWorkspace:
 
         Raises:
             AirbyteError: If raise_on_error=True and the organization info cannot be fetched
-                (e.g., due to insufficient permissions).
+                (e.g., due to insufficient permissions or missing data).
         """
         try:
             info = self._organization_info
@@ -159,9 +159,24 @@ class CloudWorkspace:
                 raise
             return None
 
+        organization_id = info.get("organizationId")
+        organization_name = info.get("organizationName")
+
+        # Validate that both organization_id and organization_name are non-null and non-empty
+        if not organization_id or not organization_name:
+            if raise_on_error:
+                raise AirbyteError(
+                    message="Organization info is incomplete.",
+                    context={
+                        "organization_id": organization_id,
+                        "organization_name": organization_name,
+                    },
+                )
+            return None
+
         return CloudOrganization(
-            organization_id=str(info.get("organizationId", "")),
-            organization_name=info.get("organizationName"),
+            organization_id=organization_id,
+            organization_name=organization_name,
         )
 
     # Test connection and creds
--- a/airbyte/mcp/cloud_ops.py
+++ b/airbyte/mcp/cloud_ops.py
@@ -159,9 +159,9 @@ class CloudWorkspaceResult(BaseModel):
     workspace_url: str | None = None
     """URL to access the workspace in Airbyte Cloud."""
     organization_id: str
-    """ID of the organization this workspace belongs to."""
+    """ID of the organization (requires ORGANIZATION_READER permission)."""
     organization_name: str | None = None
-    """Name of the organization this workspace belongs to."""
+    """Name of the organization (requires ORGANIZATION_READER permission)."""
 
 
 class LogReadResult(BaseModel):


airbyte/mcp/cloud_ops.py
```diff
@@ -485,12 +485,26 @@ def check_airbyte_cloud_workspace(
...
     )
 
+    # Try to get organization info, but fail gracefully if we don't have permissions.
+    # Fetching organization info requires ORGANIZATION_READER permissions on the organization,
+    # which may not be available with workspace-scoped credentials.
+    organization_id: str | None = None
+    organization_name: str | None = None
+    try:
+        organization_id = workspace.organization_id
+        organization_name = workspace.organization_name
```
aaronsteers: Devin, we need to move this upstream. Let's delete `organization_id` and `organization_name` (sadly, yes it's a breaking change). And instead, refactor to use `CloudWorkspace.get_organization(raise_on_error: bool = True) -> CloudOrganization | None`

Add overloads so the caller knows they get a non-nullable result in exchange for handling the exception if it occurs. If raise on error is false, we simply return None (suppressing all Exceptions).

The `CloudOrganization` can be a minimal class, following from other classes' patterns. We'll keep the caching logic internal to CloudWorkspace for now, so the CloudOrganization class can simply accept its args from CloudWorkspace method as it is constructed.

airbyte/cloud/workspaces.py
```diff
@@ -97,21 +113,56 @@ def _organization_info(self) -> dict[str, Any]:
...
+            or None if raise_on_error=False and an error occurred.
+
+        Raises:
+            AirbyteError: If raise_on_error=True and the organization info cannot be fetched
+                (e.g., due to insufficient permissions).
         """
-        return self._organization_info.get("organizationName")
+        try:
+            info = self._organization_info
+        except AirbyteError:
```
aaronsteers, line 157: Devin, are you positive the exception will be subclassed from `AirbyteError`?

airbyte/cloud/workspaces.py
```diff
@@ -97,21 +113,56 @@ def _organization_info(self) -> dict[str, Any]:
...
+            if raise_on_error:
+                raise
+            return None
+
+        return CloudOrganization(
+            organization_id=str(info.get("organizationId", "")),
+            organization_name=info.get("organizationName"),
+        )
```
aaronsteers, line 180: Let's assume missing an organization ID or organization Name is fatal and will trigger our behavior to either raise or suppress according to the caller preference. Then, we can assert that ID and Name are both non-zero lenght strings and not null.

airbyte/mcp/cloud_ops.py
```diff
@@ -485,12 +485,21 @@ def check_airbyte_cloud_workspace(
...
         workspace_id=workspace_response.workspace_id,
         workspace_name=workspace_response.name,
         workspace_url=workspace.workspace_url,
-        organization_id=workspace.organization_id or "[error: organization ID not discovered]",
-        organization_name=workspace.organization_name,
+        organization_id=(
+            organization.organization_id
+            if organization
+            else "[unavailable - requires ORGANIZATION_READER permission]"
+        ),
```
aaronsteers, line 501: Per my comment above, this handling can go away by enforcing upstream that we expect non-null values for name and org ID.

aaronsteers: Actually, no, you have it fine here. I was thinking about the previous coalesc logic.

airbyte/cloud/workspaces.py
aaronsteers: Great. That's fine if you are sure.

PR #911, opened 2025-12-14
fix: Add required permissions for prerelease workflow

aaronsteers:
## Summary

Fixes the `/prerelease` slash command workflow which was failing with:
```
Error calling workflow 'airbytehq/PyAirbyte/.github/workflows/pypi_publish.yml'. 
The nested job 'publish' is requesting 'contents: write, id-token: write', 
but is only allowed 'contents: read, id-token: none'.
```

The fix adds the required permissions to `prerelease-command.yml` so it can call the nested `pypi_publish.yml` workflow:
- `contents: write` - Required by the publish job for GitHub release uploads (only used on tag pushes)
- `id-token: write` - Required for OIDC trusted publishing to PyPI

- [ ] **Test the `/prerelease` command end-to-end** by commenting `/prerelease` on a test PR after merging. This is the only way to verify the fix works since the workflow requires the slash command dispatch infrastructure.
- [ ] **Verify OIDC publishing succeeds** - If PyPI's trusted publisher is configured to expect a different workflow filename (e.g., `publish.yml` instead of `pypi_publish.yml`), publishing will fail with an explicit error. Check the workflow logs if it fails.

### Notes

- The `contents: write` permission is broader than strictly needed for prereleases (which don't upload to GitHub releases), but is required because the nested `publish` job declares it at the job level.
- This workflow allows prereleases from fork PRs (per previous discussion), which combined with these elevated permissions is a security consideration - only maintainers can trigger the command.

**Link to Devin run:** https://app.devin.ai/sessions/c86d36be59664129af00617d0e66bc4d
**Requested by:** AJ Steers (@aaronsteers)

## Summary by CodeRabbit

* **Chores**
  * Updated GitHub Actions workflow permissions to enhance prerelease automation capabilities and security.

### DIFF
--- a/.github/workflows/prerelease-command.yml
+++ b/.github/workflows/prerelease-command.yml
@@ -16,8 +16,8 @@ env:
   AIRBYTE_ANALYTICS_ID: ${{ vars.AIRBYTE_ANALYTICS_ID }}
 
 permissions:
-  contents: write  # Required by nested pypi_publish.yml workflow (for GitHub release uploads on tags)
-  id-token: write  # Required by nested pypi_publish.yml workflow (for OIDC PyPI publishing)
+  contents: write
+  id-token: write
   pull-requests: write
   issues: write
 


.github/workflows/prerelease-command.yml
```diff
@@ -16,7 +16,8 @@ env:
   AIRBYTE_ANALYTICS_ID: ${{ vars.AIRBYTE_ANALYTICS_ID }}
 
 permissions:
-  contents: read
+  contents: write  # Required by nested pypi_publish.yml workflow (for GitHub release uploads on tags)
+  id-token: write  # Required by nested pypi_publish.yml workflow (for OIDC PyPI publishing)
```
aaronsteers: aside - 

```suggestion
  contents: write
  id-token: write
```

PR #918, opened 2025-12-17
fix: DeclarativeExecutor passing down empty configs

### LINKED ISSUE #868: DeclarativeExecutor passing down empty configs
## Steps to reproduce

```
import airbyte as ab
source = ab.get_source(
    "example-test",
    config={"hello": "world"},
    source_manifest=Path(os.getcwd()) / "mymanifest.yaml",
    streams="*"
)
source.check()
```

### Effects
The YAML is built with {{ config['hello'] }}  as an empty string.

## Root cause

https://github.com/airbytehq/PyAirbyte/blob/7eb746bda4a5e72b4643b6af7692d6a736080e60/airbyte/_executors/declarative.py#L125

`self.declarative_source` is accessing `self._config_dict` that is always empty, apart from `components_py`. Since the arg parsing is performed later in the `execute()` before calling the `source_entrypoint.run(args)`

```
    def declarative_source(self) -> ConcurrentDeclarativeSource:
        """Get the declarative source object.

        Notes:
        1. Since Sep 2025, the declarative source class used is `ConcurrentDeclarativeSource`.
        2. The `ConcurrentDeclarativeSource` object sometimes doesn't want to be read from twice,
           likely due to threads being already shut down after a successful read.
        3. Rather than cache the source object, we recreate it each time we need it, to
           avoid any issues with re-using the same object.
        """
        return ConcurrentDeclarativeSource(
            config=self._config_dict,
            source_config=self._manifest_dict,
        )
``` 

## Proposed fix
Promote config_dict in the initialization params of DeclarativeExecutor and use it instead of initializing it as an empty dict.

aaronsteers:
## Summary

Fixes #868 - Declarative YAML connectors were not receiving user config, causing template expressions like `{{ config['hello'] }}` to resolve to empty values.

This PR threads the `config` parameter from `get_source()` through `get_connector_executor()` to `DeclarativeExecutor`, so declarative manifests can access user-provided configuration.

**Key changes:**
- Add `config` parameter to `DeclarativeExecutor.__init__()` 
- Add `config` parameter to `get_connector_executor()` factory
- Pass `config` from `get_source()` through the call chain
- Use `config.copy() if config is not None else {}` to avoid mutable default argument issues (B006 lint error)

This supersedes PR #869 from @adileo, incorporating their fix plus addressing the mutable default argument issues flagged by CodeRabbit.

- [ ] **Test with a declarative manifest** - Create a source with `source_manifest=True` or a custom manifest dict, pass a config with values, and verify the connector receives the config (e.g., template `{{ config['api_key'] }}` resolves correctly)
- [ ] **Verify no mutation of caller's config** - The `.copy()` call should prevent the executor from mutating the original config dict passed by the caller
- [ ] **Check all DeclarativeExecutor instantiation sites** - Confirm the two call sites in `util.py` (lines 335 and 351) are the only places that need updating

### Notes

- The original issue reporter (@adileo) confirmed the bug: `_config_dict` was initialized as empty `{}` and never received the user's config
- Link to Devin run: https://app.devin.ai/sessions/5094f54055bd4fc68a357aa705357fce

## Release Notes

* **New Features**
  * Added optional configuration parameter support during executor initialization, enabling more flexible configuration management across the initialization chain while ensuring configuration objects remain isolated and unmodified during setup.

### DIFF
--- a/airbyte/_executors/declarative.py
+++ b/airbyte/_executors/declarative.py
@@ -45,7 +45,7 @@ class DeclarativeExecutor(Executor):
         self,
         name: str,
         manifest: dict | Path,
-        config: dict[str, Any] = {},
+        config: dict[str, Any] | None = None,
         components_py: str | Path | None = None,
         components_py_checksum: str | None = None,
     ) -> None:
@@ -67,7 +67,7 @@ class DeclarativeExecutor(Executor):
         elif isinstance(manifest, dict):
             self._manifest_dict = manifest
 
-        config_dict: dict[str, Any] = config
+        config_dict: dict[str, Any] = config.copy() if config is not None else {}
         if components_py:
             if isinstance(components_py, Path):
                 components_py = components_py.read_text()
--- a/airbyte/_executors/util.py
+++ b/airbyte/_executors/util.py
@@ -5,7 +5,7 @@ import hashlib
 import sys
 import tempfile
 from pathlib import Path
-from typing import TYPE_CHECKING, Literal, cast
+from typing import TYPE_CHECKING, Any, Literal, cast
 
 import requests
 import yaml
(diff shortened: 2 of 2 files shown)


mpearmain:
is this going to be merged soon? i have the exact same issue

PR #975, opened 2026-02-06
feat(cloud): add connection state and catalog management to CloudConnection

aaronsteers:
## Summary

Adds connection state and catalog dump/import operations to `CloudConnection`, extracted from [airbytehq/airbyte-ops-mcp#338](https://github.com/airbytehq/airbyte-ops-mcp/pull/338) as reusable core library functionality. This enables both MCP tools and CLI commands to share the same implementation via PyAirbyte.

**Private module `airbyte/cloud/_connection_state.py`:**
- Pydantic models (`ConnectionStateResponse`, `StreamState`, `GlobalState`, `StreamDescriptor`) with camelCase aliases — used internally only, not exported in the public API
- Helper functions `_match_stream()` and `_get_stream_list()` for stream-level filtering

**New `CloudConnection` methods (all inputs/outputs are raw dicts):**

| Method | Purpose |
|--------|---------|
| `dump_raw_state()` | Returns the full raw state envelope as a dict (backup use case) |
| `import_raw_state(connection_state_dict)` | Replaces the entire connection state (restore use case). Always overrides `connectionId` for cross-connection portability. Safe endpoint rejects updates during active syncs (HTTP 423). |
| `get_stream_state(stream_name)` | Returns just the inner state blob for one stream |
| `set_stream_state(stream_name, state_blob_dict)` | Fetch-modify-push for a single stream's state; fails on legacy/not_set states |
| `dump_raw_catalog()` | Returns the configured catalog (`syncCatalog`) as a dict |
| `import_raw_catalog(catalog)` | Replaces the entire configured catalog via PATCH-style update |

**API layer (`api_util.py`):**
- `replace_connection_state()` — calls `POST /v1/state/create_or_update_safe`, catches HTTP 423 → `AirbyteConnectionSyncActiveError`
- `replace_connection_catalog()` — calls `POST /v1/web_backend/connections/update` with `skipReset: False`

**Other changes:**
- `get_state_artifacts()` and `get_catalog_artifact()` deprecated via `@deprecated` decorator
- New `AirbyteConnectionSyncActiveError` exception for HTTP 423
- MCP `get_connection_artifact` tool updated to use `dump_raw_state()` / `dump_raw_catalog()`
- `set_stream_state` preserves raw dicts for non-matching streams (avoids Pydantic round-trip data loss)

- [ ] **`set_stream_state` fetch-modify-push correctness (HIGH RISK)**: This method uses `zip(raw_streams, parsed_streams)` to pair raw API dicts with Pydantic-parsed models for matching. Non-matching streams are preserved as raw dicts (no Pydantic round-trip), but the zip pairing assumes the raw and parsed lists are in the same order and same length. Must be tested against a real connection with both `stream` and `global` state types.
- [ ] **`import_raw_catalog` with `skipReset: False`**: Catalog replacement triggers destination table resets. For backup/restore workflows where the catalog hasn't actually changed, this could unnecessarily reset destination tables. Evaluate whether `skipReset` should be a caller-configurable parameter.
- [ ] **`connectionId` override behavior**: `replace_connection_state` does `{**connection_state_dict, "connectionId": connection_id}`. Verify the API accepts this and doesn't require `connectionId` in a nested location.
- [ ] **Namespace matching semantics**: `_match_stream()` treats `None` and `""` as equivalent (both mean "no namespace"). There is no wildcard behavior — caller must match the actual namespace. Verify this is acceptable.

**Suggested test plan:**
```python
from airbyte.cloud import CloudWorkspace

workspace = CloudWorkspace(workspace_id="...", client_id="...", client_secret="...")
connection = workspace.get_connection(connection_id="<real-connection-id>")

# Full state dump/import round-trip
state = connection.dump_raw_state()
print(state)
# connection.import_raw_state(state)  # only when safe to test

# Cross-connection portability test
# other_connection = workspace.get_connection(connection_id="<different-connection-id>")
# other_connection.import_raw_state(state)  # should work — connectionId is overridden

# Stream-level get/set
stream_state = connection.get_stream_state("some_stream_name")
print(stream_state)
# connection.set_stream_state("some_stream", {"cursor": "2024-01-01"})

# Catalog dump/import
catalog = connection.dump_raw_catalog()
print(catalog)
# connection.import_raw_catalog(catalog)  # triggers reset!
```

### Notes

- Unit tests cover models and helper functions (15 tests, all passing)
- No integration tests for the `CloudConnection` methods — they require real API credentials
- This PR is a prerequisite for refactoring airbyte-ops-mcp#338 to use PyAirbyte as a dependency

---
Link to Devin run: https://app.devin.ai/sessions/5ce605c67c5f436d94d5d76a2c6b95fc
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Full connection state and catalog management: dump/import/replace entire connection state and sync catalog; per-stream read/update helpers; structured connection state model.

* **Bug Fixes / Reliability**
  * New specific error surfaced when attempting state updates during an active sync (HTTP 423).

* **Deprecated**
  * Legacy state/catalog artifact retrieval methods marked deprecated; use the new dump/import APIs.

* **Tests**
  * Unit tests covering connection state models, (de)serialization, and per-stream/global behaviors.

* **API**
  * Connection artifact responses now consistently return a single object rather than mixed types.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

### DIFF
--- a/airbyte/_util/api_util.py
+++ b/airbyte/_util/api_util.py
@@ -2224,7 +2224,7 @@ def replace_connection_catalog(
         json={
             "connectionId": connection_id,
             "syncCatalog": configured_catalog_dict,
-            # Resets are destructive and cause customer-side data outage. 
+            # Resets are destructive and cause customer-side data outage.
             # If a reset is desired, caller will need to decide & manage.
             "skipReset": True,
         },


airbyte/cloud/connections.py
```diff
@@ -389,6 +394,194 @@ def get_state_artifacts(self) -> list[dict[str, Any]] | None:
...
+            available = [s.stream_descriptor.name for s in streams]
+            raise PyAirbyteInputError(
+                message=f"Stream '{stream_name}' not found in connection state.",
+                context={
+                    "connection_id": self.connection_id,
+                    "stream_name": stream_name,
+                    "stream_namespace": stream_namespace,
+                    "available_streams": available,
+                },
+            )
```
aaronsteers: For a handled case of "no result found" (meaning no other random exceptions), we can simply log a warning and return None.

airbyte/_util/api_util.py
```diff
@@ -2092,6 +2092,49 @@ def get_connection_state(
...
+            - connectionId: Must match the connection_id parameter
+            - One of: state (legacy), streamState (stream), globalState (global)
+        api_root: The API root URL
+        client_id: OAuth client ID
+        client_secret: OAuth client secret
+        bearer_token: Bearer token for authentication (alternative to client credentials).
+
+    Returns:
+        Dictionary containing the updated ConnectionState object.
+    """
```
aaronsteers: I can't tell from this (or the below API call) if the stream name object within connection_state is going to be properly deduped in the backend - meaning that calling with only 1 out of 10 stream states would correctly update (or append) 1 stream state without affected the other nine. Please confirm, and update the docstring to be explicit.

airbyte/cloud/connections.py
```diff
@@ -389,6 +398,197 @@ def get_state_artifacts(self) -> list[dict[str, Any]] | None:
             return None
         return state_response.get("streamState", [])
 
+    def get_state(
+        self,
+        *,
+        stream_name: str | None = None,
+        stream_namespace: str | None = None,
+    ) -> ConnectionStateResponse | None:
```
aaronsteers: I think we should only return the JSON glob body (dict). And likewise in the "set" operation, only accepting "dict" as input and optional StateTypeEnum that defaults to STREAM.

airbyte/cloud/_connection_state.py
```diff
@@ -0,0 +1,131 @@
...
+    stream: StreamState,
+    stream_name: str,
+    stream_namespace: str | None = None,
+) -> bool:
+    """Check if a StreamState matches the given name and optional namespace."""
+    if stream.stream_descriptor.name != stream_name:
+        return False
+    if stream_namespace is not None:
+        return stream.stream_descriptor.namespace == stream_namespace
+    return True
```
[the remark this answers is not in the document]
aaronsteers: Devin, code rabbit has a point. I thin it is worth consider if blank (None or "") namespaces should only match each other, and not to other namespaces not set. What do you think? Don't immediately update anything - just lmk in Slack what you think.

airbyte/cloud/connections.py
aaronsteers: No, I don't want the caller to know or care how our API performs its work. I want the caller to think that "state" is equivalent with the json payload attached to that stream - not a series of convuluted nested classes that require training to traverse. So, get/set operate on the JSON dict payload that _is_ the state of hte stream, decoupled from our many envelopes around it. Makes sense?

airbyte/_util/api_util.py
```diff
@@ -2092,6 +2092,56 @@ def get_connection_state(
     )
 
 
+def create_or_update_connection_state_safe(
```
aaronsteers: Remove _safe suffix - it implies other things that aren't implemented (such as deduping/merging against already existing). Since we clobber, I don't think "safe" is appropriate in the name on our side.

airbyte/_util/api_util.py
```diff
@@ -2092,6 +2092,56 @@ def get_connection_state(
...
+        bearer_token: Bearer token for authentication (alternative to client credentials).
+
+    Returns:
+        Dictionary containing the updated ConnectionState object.
+    """
+    return _make_config_api_request(
+        path="/state/create_or_update_safe",
+        json={
+            "connectionId": connection_id,
+            "connectionState": connection_state,
```
aaronsteers: Do we also need to inject the provided connectionId also on the inner connection_state payload?

airbyte/_util/api_util.py
```diff
@@ -2092,6 +2092,56 @@ def get_connection_state(
     )
 
 
+def create_or_update_connection_state(
```
aaronsteers: I think this name is clearer: `replace_connection_state`

airbyte/cloud/connections.py
```diff
@@ -368,26 +378,191 @@ def get_sync_result(
...
+        Args:
+            connection_state: The full connection state to import. Must include:
+                - stateType: "global", "stream", or "legacy"
+                - One of: state (legacy), streamState (stream), globalState (global)
+                - connectionId is optional and will be overridden.
+
+        Returns:
+            The updated connection state as a dictionary.
+        """
+        state_to_send = {**connection_state, "connectionId": self.connection_id}
```
aaronsteers: Move this to the inner api_util helper.

airbyte/cloud/connections.py
```diff
@@ -368,26 +378,189 @@ def get_sync_result(
...
 
     def get_state_artifacts(self) -> list[dict[str, Any]] | None:
-        """Get the connection state artifacts.
+        """Deprecated. Use `dump_raw_state()` instead."""
+        warnings.warn(
+            "The 'get_state_artifacts' method is deprecated and will be removed"
+            " in a future version. Use 'dump_raw_state()' instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
aaronsteers: Isn't there a python decorator we can just put around the method?

airbyte/cloud/connections.py
```diff
@@ -367,27 +378,185 @@ def get_sync_result(
...
+            The updated connection state as a dictionary.
+        """
+        return api_util.replace_connection_state(
+            connection_id=self.connection_id,
+            connection_state=connection_state,
+            api_root=self.workspace.api_root,
+            client_id=self.workspace.client_id,
+            client_secret=self.workspace.client_secret,
+            bearer_token=self.workspace.bearer_token,
+        )
```
aaronsteers, line 450: Should we catch the expected API error code and raise a native error for readability.

aaronsteers: Actually - that should probably be wrapped inside api_util function - so this function won't need to understand the API internals.

airbyte/cloud/connections.py
```diff
@@ -367,27 +378,185 @@ def get_sync_result(
...
             return None
-        return state_response.get("streamState", [])
+
+        return matching[0].stream_state
+
+    def set_stream_state(
+        self,
+        stream_name: str,
+        stream_state: dict[str, Any],
+        stream_namespace: str | None = None,
```
aaronsteers, line 499: Isn't namespace a property of the connection? Do we know our own namespace and if so, can we remove from the function signature?

airbyte/cloud/connections.py
```diff
@@ -367,27 +378,185 @@ def get_sync_result(
...
+        The ``connectionId`` in the blob is always overridden with this connection's
+        ID, making state blobs portable across connections.
+
+        Args:
+            connection_state: The full connection state to import. Must include:
+                - stateType: "global", "stream", or "legacy"
+                - One of: state (legacy), streamState (stream), globalState (global)
+
+        Returns:
+            The updated connection state as a dictionary.
```
aaronsteers, line 437: Document for our users here the friendly exception that our inner function now raises.

airbyte/cloud/connections.py
aaronsteers: That makes sense on the source side.

But how does this work if the user toggles the setting on the connection - under "advanced settings", they can override the namespace.

airbyte/cloud/_connection_state.py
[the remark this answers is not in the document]
aaronsteers: It feels like a bad design choice if the platform has to forcibly alter connection state if/when users set an override namespace at the connection level. More likely, the platform handles this as an add-on, and (perhaps?) layers it into the destination without altering anything in regards to the connection state. If that is the case, then the best behavior is probably "null==''" and no wildcard behaviors at all.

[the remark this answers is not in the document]
aaronsteers: I see you answered the same in the other thread. Just seeing that now. I think this confirms it.

airbyte/cloud/connections.py
aaronsteers: Got it! Maybe worth a short inline comment in the docstring to say the input namespace is not respective of connection-level advanced overrides which might be set by the user.

airbyte/cloud/connections.py
```diff
@@ -367,28 +378,197 @@ def get_sync_result(
...
         )
-        if state_response.get("stateType") == "not_set":
+
+    def import_raw_state(
+        self,
+        connection_state: dict[str, Any],
+    ) -> dict[str, Any]:
+        """Import (restore) the full raw state for this connection.
+
+        Replaces the entire connection state with the provided state blob.
```
aaronsteers, line 424: Insert a user-facing warning (markdown formatted). Something like:


```suggestion

        > **WARNING:** Modifying the state directly is not recommended and
        > could result in broken connections, and/or incorrect sync behavior.

        Replaces the entire connection state with the provided state blob.
```


And a similar inline warning for the "import catalog" public method.

aaronsteers: Actuall - add a unicode "⚠️" emoji for quick visual communication.

airbyte/cloud/connections.py
```diff
@@ -367,28 +378,209 @@ def get_sync_result(
...
         )
-        if state_response.get("stateType") == "not_set":
+
+    def import_raw_state(
+        self,
+        connection_state: dict[str, Any],
+    ) -> dict[str, Any]:
+        """Import (restore) the full raw state for this connection.
+
+        > \u26a0\ufe0f **WARNING:** Modifying the state directly is not recommended and
```
aaronsteers: aside - 

```suggestion
        > ⚠️ **WARNING:** Modifying the state directly is not recommended and
```

airbyte/cloud/connections.py
```diff
@@ -409,6 +615,31 @@ def get_catalog_artifact(self) -> dict[str, Any] | None:
         )
         return connection_response.get("syncCatalog")
 
+    def import_raw_catalog(self, catalog: dict[str, Any]) -> None:
+        """Replace the configured catalog for this connection.
+
+        > \u26a0\ufe0f **WARNING:** Modifying the catalog directly is not recommended and
```
aaronsteers: aside - 


```suggestion
        > ⚠️ **WARNING:** Modifying the catalog directly is not recommended and
```

airbyte/cloud/connections.py
```diff
@@ -367,28 +378,209 @@ def get_sync_result(
...
             api_root=self.workspace.api_root,
             client_id=self.workspace.client_id,
             client_secret=self.workspace.client_secret,
             bearer_token=self.workspace.bearer_token,
         )
-        if state_response.get("stateType") == "not_set":
+
+    def import_raw_state(
+        self,
+        connection_state: dict[str, Any],
```
aaronsteers: Rename to match the other naming convention: `connection_state_dict`

airbyte/cloud/connections.py
```diff
@@ -367,28 +378,209 @@ def get_sync_result(
...
+            )
             return None
-        return state_response.get("streamState", [])
 
+        return matching[0].stream_state
+
+    def set_stream_state(
+        self,
+        stream_name: str,
+        stream_state: dict[str, Any],
```
aaronsteers: `stream_state_dict`

aaronsteers: Actually, `state_blob_dict`

airbyte/cloud/connections.py
```diff
@@ -367,28 +378,209 @@ def get_sync_result(
...
+    def get_stream_state(
+        self,
+        stream_name: str,
+        stream_namespace: str | None = None,
+    ) -> dict[str, Any] | None:
+        """Get the state blob for a single stream within this connection.
+
+        Returns just the stream's state dictionary (e.g., {"cursor": "2024-01-01"}),
+        not the full connection state envelope.
+
```
aaronsteers, line 461: Add a note here and in set_stream_state - mentioning that this is compatible with STREAM state and stream-level entries within a GLOBAL state, but it does not compatible with LEGACY state.

We should also mention that `dump_raw_state` and `import_raw_state` can be used to get/set the entire connection-level state artifact. (Use backticks as I have done to ensure pdoc links these refs as hyperlinks.)

airbyte/cloud/connections.py
```diff
@@ -367,28 +378,218 @@ def get_sync_result(
...
+        This is compatible with ``stream``-type state and stream-level entries
+        within a ``global``-type state. It is not compatible with ``legacy`` state.
+        To get or set the entire connection-level state artifact, use
+        `dump_raw_state` and `import_raw_state` instead.
```
aaronsteers: Double-backticks are not a thing. Don't do this. Just use single-backticks.

airbyte/_util/api_util.py
```diff
@@ -2127,6 +2192,47 @@ def get_connection_catalog(
...
+
+    Returns:
+        Dictionary containing the updated WebBackendConnectionRead response.
+    """
+    return _make_config_api_request(
+        path="/web_backend/connections/update",
+        json={
+            "connectionId": connection_id,
+            "syncCatalog": configured_catalog_dict,
+            "skipReset": True,
```
aaronsteers, line 2229: aside -

```suggestion
            # Resets are destructive and cause customer-side data outage. 
            # If a reset is desired, caller will need to decide & manage.
            "skipReset": True,
```

PR #969, opened 2026-01-29
feat(connectors): add universal source and destination using PyAirbyte

aaronsteers:
## Summary

Adds universal source and destination connectors that leverage PyAirbyte as a backend. These connectors are named `destination-pyairbyte` and `source-pyairbyte` internally, but will be published to the Airbyte connector registry as `destination-smoke-test` and `source-smoke-test` (via a companion PR in airbytehq/airbyte).

**Universal Destination** (`destination-pyairbyte`): Writes to any PyAirbyte-supported cache backend (DuckDB, Postgres, Snowflake, BigQuery, MotherDuck). Implements the Airbyte CDK `Destination` interface with `spec()`, `check()`, `write()`, and support for `overwrite`/`append` sync modes.

**Universal Source** (`source-pyairbyte`): Wraps any PyAirbyte-supported source connector, acting as a proxy that delegates to the underlying source based on configuration. Implements the CDK `Source` interface with `spec()`, `check()`, `discover()`, and `read()` methods.

**Smoke Test Source** (`source-smoke-test`): A dedicated test source with predefined scenarios for destination regression testing. Includes scenarios for basic types, timestamps, large decimals, nested JSON, null handling, column naming edge cases, and more.

All connectors are registered as CLI entrypoints and include a Dockerfile for containerization.

## Updates since last revision

- Merged `source-smoke-test` connector from PR #982 (predefined scenarios for destination regression testing)
- Reverted CLI entrypoints back to `destination-pyairbyte`/`source-pyairbyte` (internal PyAirbyte naming)
- The airbyte connector registry will use `destination-smoke-test`/`source-smoke-test` names (handled in companion PR)
- Added TK-TODO comment for STATE message handling (tracks tech debt for incremental sync support)
- Imported TK-TODO CI check workflow from airbyte-ops-mcp repo

## Review & Testing Checklist for Human

- [ ] **State handling limitation (source)**: The `state` parameter in `read()` is ignored and STATE messages are skipped. Incremental syncs will not work - only full refresh is supported. This is tracked with a TK-TODO marker.
- [ ] **SQL injection risk (destination)**: Schema and table names are interpolated into SQL strings without identifier quoting. Verify this is acceptable given these values come from configuration/catalog.
- [ ] **Private API usage (source)**: Uses PyAirbyte's internal `_read_with_catalog()` method which may change without notice.
- [ ] **Exception messages in check()**: The source `check()` method returns raw exception text which could potentially expose config details. Consider if this needs sanitization.
[...]

PR #982, opened 2026-02-12
feat(connectors): add smoke test source for destination regression testing

### LINKED ISSUE #981: feat: PyAirbyte-powered Smoke Test Source for Destination Regression Testing
## Summary

Create a **PyAirbyte-powered Smoke Test Source** connector designed specifically for destination regression testing. This should be implemented as a stacked PR on top of #969 (universal source/destination).

## Context

We want to test destinations the same way we test sources today -- in a regression test framework that has **no side effects and no customer impact**. The Smoke Test Source is the first building block of this framework.

<details>
<summary>Full Destination Regression Test Plan</summary>

#### 1. Smoke Test Source (This Issue)

A purpose-built source connector that generates test data for destinations:

- **Predefined scenarios:** Ships with a curated set of data patterns known to commonly cause destination failures (edge-case types, schema variations, etc.)
- **Dynamic injection:** Supports dynamically adding new failure conditions discovered from production issues. When AI Triage identifies a customer-facing failure, an **anonymized failure condition** is generated from the error log and added to the smoke test source's scenario set.

#### 2. Test Harness

Orchestrates the end-to-end regression test:

1. **Record Injection** -- Feeds test scenarios from the Smoke Test Source into the destination under test.
2. **Destination Write** -- The destination connector processes the injected records.
3. **Destination Describe** -- Reads back the destination's state/schema. *(Today, destinations cannot describe their own data. The code for this exists in integration tests but must be promoted into the connector's main dependencies so it ships with the connector.)*
4. **Data Dumper** -- Extracts the written data from destination storage for comparison. *(Same as Describe -- must be moved from integration test code into the shipping connector.)*

#### 3. Control Test (Two-Version Comparison)

Every regression test runs against **two versions** of the destination connector (current release vs. candidate):

- Catches regressions and unexpected behavioral differences even when no known failure exists.
- Produces a **Diff Report** comparing the two runs.

#### 4. Diff Report (Per Table)

| Check | Description |
|---|---|
| **Dest Schema / Types** | Schema and column type correctness |
| **Record Count** | Expected vs. actual row counts |
| **Record Data / Null Handling** | Data integrity, especially null value handling |
| **Table Name** | Correct table naming |
| **Column Name + Casing** | Column names and case-sensitivity preservation |
| **Spec Output Diff** | Differences in connector spec output between versions |

#### 5. TDD Workflow (AI-Assisted)

1. **AI Triage** identifies a destination failure from production error logs.
2. An **anonymized failure condition** is generated and injected into the Smoke Test Source.
3. The regression test now has a **failing test case** for the specific destination.
4. A PR is opened, and **TDD** proceeds -- iterating until the test passes.
5. (`ai-prove-fix`) The fix is validated by the regression framework before merge.

#### Out of Scope

- Speed Mode (performance optimization of the test harness).

</details>

## Scope of This Issue

Build the Smoke Test Source as a PyAirbyte-powered connector, stacking on the universal source/destination work in #969. Specifically:

### Requirements

1. **Predefined test scenarios** -- A curated set of data patterns that commonly break destinations:
   - Various column types (including edge cases like timestamps, large decimals, nested JSON, etc.)
   - Null handling edge cases
   - Table/column naming edge cases (special characters, casing, long names, reserved words)
   - Schema variations and evolution scenarios
   - Large and small record batches

2. **Dynamic scenario injection** -- Ability to add new failure scenarios at runtime (e.g., from a config file or catalog entry), so that newly discovered production failures can be reproduced without code changes.

3. **Standard Airbyte Source interface** -- Implements `spec()`, `check()`, `discover()`, `read()` per the CDK, so it can be used with any destination via the standard Airbyte protocol.

4. **Anonymized failure reproduction** -- Support for defining scenarios based on anonymized production error patterns (no real customer data).

### Relationship to #969

This should be built on top of the universal connector infrastructure from #969, reusing its patterns for CLI entrypoints, Dockerfile, and PyAirbyte integration.

aaronsteers:
This PR targets the following PR:
- #969

---

## Summary

Adds a **Smoke Test Source** connector (`source-smoke-test`) for destination regression testing, as described in #981. This source generates synthetic data across 15 predefined scenarios designed to exercise common destination failure patterns:

- **Type coverage**: basic types, timestamps, large decimals, nested JSON/arrays
- **Null handling**: nullable columns across all types, always-null columns, null-vs-empty-vs-zero
- **Naming edge cases**: reserved SQL words, CamelCase, dots/dashes/spaces in column names, long column names, mixed-case stream names
- **Schema variations**: wide table (50 columns), no-primary-key stream, empty stream, single-record stream
- **Batch sizes**: configurable large batch (default 1000 records)
- **Unicode/special strings**: international characters, escape sequences

The source also supports **dynamic scenario injection** via a `custom_scenarios` config field, enabling new failure conditions (e.g., anonymized from production errors) to be added at runtime without code changes.

**New files:**
- `airbyte/cli/universal_connector/smoke_test_source.py` — `SourceSmokeTest` class + `PREDEFINED_SCENARIOS` data
- `airbyte/cli/universal_connector/run_smoke_test.py` — CLI entry point
- Updated `__init__.py` exports and `pyproject.toml` entry point registration

## Local Testing Results

All CLI commands verified locally:
- `spec` — outputs valid JSON schema
- `check --config '{}'` — passes with 15 scenarios
- `discover --config '{}'` — returns all 15 streams
- `read` with full catalog — emits **1028 records** across 14 streams (empty_stream correctly emits 0)
- `scenario_filter` — correctly filters discover and read to only named streams
- `large_batch_record_count=5` — correctly limits large_batch_stream to 5 records
- `large_batch_record_count=0` — correctly emits 0 records (stream still appears in catalog)
- `custom_scenarios` — successfully injects and reads custom scenario data

- [ ] **No automated tests are included.** The source has not been verified with any unit or integration tests. Consider whether tests should be required before merge or tracked as follow-up.
- [ ] **`wide_table_50_columns` schema mismatch**: The second record sets all 49 columns to `None`, but the schema declares them as `"type": "string"` (not nullable). This may cause issues with strict schema validation in some destinations.
- [ ] **`large_batch_record_count=0` behavior**: The stream still appears in `discover()` but emits zero records. Verify whether this is the desired behavior or if the stream should be excluded from the catalog entirely.
- [ ] **End-to-end test**: Pipe output into `destination-pyairbyte` (from #969) to verify compatibility.

**Recommended test plan:**
1. Run `source-smoke-test spec` and verify the output JSON schema
2. Run `source-smoke-test check --config '{}'` and confirm success
3. Run `source-smoke-test discover --config '{}'` and verify all 15 streams appear
4. Run `source-smoke-test read --config '{}' --catalog <catalog.json>` with a catalog selecting a few streams, and verify records are emitted correctly
5. Pipe output into `destination-pyairbyte` (from #969) to verify end-to-end compatibility

### Notes

- This is stacked on #969 (universal source/destination). Only review the incremental changes, not the base branch.
- Closes #981

Requested by: @aaronsteers  
Devin session: https://app.devin.ai/sessions/02ea317be9054bcba5e48a6ed1622620

---

<a href="https://app.devin.ai/review/airbytehq/pyairbyte/pull/982" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/airbyte/cli/universal_connector/source.py
+++ b/airbyte/cli/universal_connector/source.py
@@ -192,4 +192,5 @@ class SourcePyAirbyteUniversal(Source):
                         namespace=message.record.namespace,
                     ),
                 )
-            # TODO(https://github.com/airbytehq/PyAirbyte/issues/987): Add STATE message handling for incremental sync support.
+            # TODO: Add STATE message handling for incremental sync support.
+            # https://github.com/airbytehq/PyAirbyte/issues/987


airbyte/cli/universal_connector/smoke_test_source.py
```diff
@@ -0,0 +1,689 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable, Mapping
+
+
+logger = logging.getLogger("airbyte")
+
+_DEFAULT_LARGE_BATCH_COUNT = 1000
+
+PREDEFINED_SCENARIOS: list[dict[str, Any]] = [
```
aaronsteers: This is great. For now, let's move the declaration to a sibling module _source_smoke_test_scenarios.py

airbyte/cli/universal_connector/smoke_test_source.py
```diff
@@ -0,0 +1,689 @@
...
+        """Return the connector specification."""
+        return ConnectorSpecification(
+            documentationUrl="https://docs.airbyte.com/integrations/sources/smoke-test",
+            connectionSpecification={
+                "$schema": "http://json-schema.org/draft-07/schema#",
+                "title": "Smoke Test Source Spec",
+                "type": "object",
+                "required": [],
+                "properties": {
+                    "custom_scenarios": {
```
aaronsteers, line 83: Add this as default behavior:

If omitted, all scenarios will be attempted except the high-volume scenario. This is to balance a robust default set of tests while not automatically incurring the cost of the larger volume test.

Also, include a boolean config option "include_high_volume_scenarios" (default false) to enable those slower tests without having to name them. And similarly, "include_default_scenarios" (default true) to enable-disable the other (faster) tests as a group.

Then, this named scenario list is only really needed for precise tuning. In terms of behavior: the named scenarios in this list will be unioned with the other boolean-driven sets, with overlaps deduped in the final scenario list queud for execution.

aaronsteers:
Devin, log a new issue regarding STATE here in the pyairbyte repo, then replace the TK-TODO comment with a regular TODO comment and link to the newly logged issue.

PR #998, opened 2026-03-18
feat(cli,mcp): add destination-smoke-test command and MCP tool

aaronsteers:
## Summary

Adds a new `pyab destination-smoke-test` CLI command and a corresponding `destination_smoke_test` MCP tool. Both send synthetic data from the existing `SourceSmokeTest` (15 predefined scenarios covering type variations, null handling, naming edge cases, etc.) to a destination connector via `Destination.write()` and report pass/fail as structured JSON.

This is the execution layer for destination regression testing — it answers "does this destination accept the smoke test data without errors?" No readback or comparison is performed (that's a future increment).

**Architecture:** Shared logic lives in `airbyte/_util/destination_smoke_tests.py` — both CLI and MCP are thin presentation layers that delegate to `run_destination_smoke_test()`. Neither imports from the other.

**Scenario keywords:**
- `fast` (default) — all fast predefined scenarios, excluding `large_batch_stream`
- `all` — every predefined scenario including `large_batch_stream`
- Comma-separated list — explicit subset of scenario names

**CLI usage:**
```bash
pyab destination-smoke-test \
    --destination=destination-snowflake \
    --config=./secrets/snowflake.json

pyab destination-smoke-test \
    --destination=destination-dev-null \
    --scenarios=basic_types,null_handling

pyab destination-smoke-test \
    --destination=destination-snowflake \
    --config=./secrets/snowflake.json --scenarios=all
```

**MCP tool:** Same functionality exposed as `destination_smoke_test` in the `local` tools module, returning a structured `DestinationSmokeTestResult` Pydantic model. Also supports `custom_scenarios` as inline dicts and `docker_image` override.

### Updates since last revision

- **Stream status TRACE messages (bug fix to smoke test source).** JDK-based destinations (e.g. `destination-dev-null`) require `STARTED`/`RUNNING`/`COMPLETE` stream status trace messages per the Airbyte protocol. The smoke test source was only emitting `RECORD` messages, causing all JDK destinations to fail with `TransientErrorException: streams did not receive a terminal stream status message`. Fixed in `source.py` by emitting the required `TRACE` messages around each stream's records.
- **`fast`/`all` scenario keywords implemented** with `fast` as default. Empty scenario lists are normalized to `fast`.
- **Tested locally against `destination-dev-null`** — 28 records across 13 fast streams delivered successfully, exit code 0.

- [ ] **Stream status TRACE messages are correct.** This fix modifies `source.py` (shipped in PR #996). The sequence is STARTED → RUNNING → records → COMPLETE per stream. Verify this matches the protocol expectations for both JDK and Python CDK destinations — the order matters and incorrect ordering could cause different failures on other destinations.
- [ ] **Test against a real credentialed destination** (e.g. Snowflake, MotherDuck). The local test only covered `destination-dev-null` which is a no-op sink — it validates protocol compliance but not actual data delivery.
- [ ] **Review MCP Docker auto-detection divergence.** When no `docker_image` is specified and Docker is installed, the MCP tool defaults to `docker_image=True`. The CLI instead delegates to `_resolve_destination_job` which has slightly different resolution logic. Confirm this divergence is acceptable.
- [ ] **Error sanitization fallback.** `_sanitize_error()` uses `get_message()` for PyAirbyte exceptions but falls back to `str(ex)` for others. Third-party exceptions could theoretically include config details.

### Recommended test plan
```bash
# 1. Install from branch
pip install 'git+https://github.com/airbytehq/PyAirbyte.git@devin/1773861379-destination-smoke-test'

# 2. Quick smoke test (no credentials needed)
pyab destination-smoke-test --destination=destination-dev-null

# 3. With scenario filtering
pyab destination-smoke-test --destination=destination-dev-null --scenarios=basic_types,null_handling

# 4. All scenarios including large batch
pyab destination-smoke-test --destination=destination-dev-null --scenarios=all

# 5. Against a real destination
pyab destination-smoke-test --destination=destination-snowflake --config=./secrets/snowflake.json

# 6. Verify exit code on failure (bad config)
pyab destination-smoke-test --destination=destination-snowflake; echo "exit: $?"
```

### Notes
- The `DestinationSmokeTestResult` model and `_sanitize_error` helper are exported from the shared module for potential reuse by the ops repo test harness.
- The stream status fix also affects the smoke test source when used standalone (not just via this new command). This is a net positive — it was broken for all JDK destinations before this fix.

Link to Devin session: https://app.devin.ai/sessions/9c72389579884c06bf18cef11c4550e8
Requested by: @aaronsteers

## New Features

* Added a CLI command to run destination smoke tests with configurable scenarios and customization options
* Support for custom test scenarios via direct parameters or external files
* Smoke test results displayed as JSON output with execution metrics and delivery details
* Added stream status lifecycle tracking during test execution

### DIFF
--- a/airbyte/cli/smoke_test_source/source.py
+++ b/airbyte/cli/smoke_test_source/source.py
@@ -25,10 +25,15 @@ from airbyte_cdk.models import (
     AirbyteMessage,
     AirbyteRecordMessage,
     AirbyteStream,
+    AirbyteStreamStatus,
+    AirbyteStreamStatusTraceMessage,
+    AirbyteTraceMessage,
     ConfiguredAirbyteCatalog,
     ConnectorSpecification,
     Status,
+    StreamDescriptor,
     SyncMode,
+    TraceType,
     Type,
 )
 from airbyte_cdk.sources.source import Source
@@ -262,8 +267,7 @@ class SourceSmokeTest(Source):
                 return f"Custom scenario at index {i} is missing 'name'."
             if not isinstance(scenario.get("json_schema"), dict):
                 return (
-                    f"Custom scenario '{scenario['name']}' must provide "
-                    "'json_schema' as an object."
+                    f"Custom scenario '{scenario['name']}' must provide 'json_schema' as an object."
                 )
             if "records" in scenario:
                 if not isinstance(scenario["records"], list):
@@ -320,6 +324,24 @@ class SourceSmokeTest(Source):
         logger.info(f"Discovered {len(streams)} smoke test streams.")
         return AirbyteCatalog(streams=streams)
 
+    def _stream_status_message(
+        self,
+        stream_name: str,
+        status: AirbyteStreamStatus,
+    ) -> AirbyteMessage:
+        """Build an AirbyteMessage containing a stream status trace."""
+        return AirbyteMessage(
+            type=Type.TRACE,
+            trace=AirbyteTraceMessage(
+                type=TraceType.STREAM_STATUS,
+                emitted_at=time.time() * 1000,
+                stream_status=AirbyteStreamStatusTraceMessage(
+                    stream_descriptor=StreamDescriptor(name=stream_name),
+                    status=status,
+                ),
+            ),
+        )
+
     def read(
         self,
         logger: logging.Logger,
@@ -339,6 +361,10 @@ class SourceSmokeTest(Source):
                 logger.warning(f"Stream '{stream_name}' not found in scenarios, skipping.")
                 continue
 
+            # Emit STARTED status
+            yield self._stream_status_message(stream_name, AirbyteStreamStatus.STARTED)
+            yield self._stream_status_message(stream_name, AirbyteStreamStatus.RUNNING)
+
             records = get_scenario_records(scenario)
             logger.info(f"Emitting {len(records)} records for stream '{stream_name}'.")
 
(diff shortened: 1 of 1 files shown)


airbyte/cli/pyab.py
```diff
@@ -631,6 +633,212 @@ def sync(
...
+    help=(
+        "Comma-separated list of smoke test scenario names to run. "
+        "If not provided, all fast (non-high-volume) scenarios are included. "
+        "Available scenarios include: basic_types_stream, timestamp_stream, "
+        "large_decimal_stream, nested_json_stream, null_values_stream, "
+        "column_name_edge_cases, table_name_with_dots, table_name_with_spaces, "
+        "CamelCaseStream, wide_table_50_cols, empty_stream, single_record_stream, "
+        "unicode_special_strings, no_pk_stream, long_column_names."
+    ),
+)
```
aaronsteers, line 682: Rewrite this as `--scenarios` and default to `all` which is a special trigger to take all, rather than a CSV subset.

aaronsteers: Oops. Nm. We need a keyword for all-fast and all-fast-and-slow. Chat with me about this in slack unless we already have a good convention.

airbyte/cli/pyab.py
```diff
@@ -631,6 +633,212 @@ def sync(
...
+    schema variations, and batch sizes.
+
+    This command does NOT read back data from the destination or compare
+    results. It only verifies that the destination accepts the data without
+    errors.
+
+    Examples:
+    \b
+        # Test with a Docker destination:
+        pyab destination-smoke-test \\
```
aaronsteers: Markdown docstrings, not reST.

PR #1000, opened 2026-03-23
feat: add destination readback introspection for smoke tests

aaronsteers:
## Summary

Adds automatic readback introspection to destination smoke tests. After writing data, the system reads back stats from the destination using existing PyAirbyte cache implementations as query adapters. This covers 5 destinations (~95% of user traffic): BigQuery, DuckDB, MotherDuck, Postgres, Snowflake.

### Core changes

**`Destination.get_sql_cache(schema_name=...)` and `Destination.is_cache_supported`** (`destinations/base.py`):
- New method following the same pattern as `SyncResult.get_sql_cache()` in cloud sync results
- Builds a cache from the destination's hydrated config via `destination_to_cache()`
- `is_cache_supported` property for early guard before attempting cache construction

**Table statistics on `SqlProcessorBase`** (`shared/sql_processor.py`):
- `fetch_row_count(table_name)` — row count via `SELECT COUNT(*)`
- `fetch_column_info(table_name)` — column names/types via SQLAlchemy `Inspector` (shared across calls)
- `_get_column_stats(table_name, columns)` — per-column null/non-null counts using positional aliases (`nn_0`, `nn_1`, ...) to avoid identifier truncation on PostgreSQL
- `fetch_table_statistics(stream_names)` — full orchestration: resolves table names via the processor's normalizer, queries all streams, returns `dict[str, TableStatistics]`
- New Pydantic models: `ColumnStatistics`, `TableStatistics`

**Smoke test integration** (`destination_smoke_tests.py`):
- Readback runs automatically when a compatible cache exists — no new CLI args needed
- Runs even on write failure to support partial-success inspection
- `DestinationSmokeTestResult` gains `table_statistics: dict[str, TableStatistics] | None` and `tables_not_found: dict[str, str] | None`

**Translation files moved** (`caches/_utils/`):
- `destinations/_translate_dest_to_cache.py` → `caches/_utils/_dest_to_cache.py`
- `destinations/_translate_cache_to_dest.py` → `caches/_utils/_cache_to_dest.py`
- Resolves the circular import cycle (`caches.bigquery` → `destinations.__init__` → `destinations.base` → `caches.bigquery`)
- Inline imports in `destinations/base.py` promoted to top-level now that the cycle is broken

**Other**:
- Fixed `paired_destination_name` copy-paste bugs in `SnowflakeCache`, `PostgresCache`, and `MotherDuckCache`
- Added dict-config handling to all `*_destination_to_cache()` functions (strip `destinationType` dispatch key before model construction)
- Docstring style standardized to Markdown; `CONTRIBUTING.md` updated with style guide

### Namespace resolution approach

Investigation confirmed that the **catalog namespace** (set on each stream by the source) is the primary and *sufficient* mechanism for directing destinations to write into the test schema. Both Postgres and Snowflake correctly create final tables in the catalog-specified namespace without needing a config-level schema key override.

The config schema key override was removed because:
- It is redundant when the catalog namespace is set correctly
- It caused Postgres failures due to 63-char identifier truncation on raw table names (long namespace prefix + raw table name exceeded the limit)

`_prepare_destination_config()` now only forces `disable_type_dedupe=False` (required because some GSM configs have it set to `True`, which causes destinations to skip final typed tables that readback depends on). Uses `validate=False` to skip redundant spec fetch.

The schema override is instead applied at the **cache level** via `destination_to_cache(schema_name=...)`, which sets `cache.schema_name` and calls `dispose_engine()` to refresh the `schema_translate_map`.

### Updates since last revision

**Credential handling fixes for the `Destination.get_sql_cache()` path:**

The `*_destination_to_cache()` functions were originally designed for the cloud API path, which returns obfuscated credentials. The new local `Destination.get_sql_cache()` path passes plaintext credentials from the destination's hydrated config, which required two fixes:

- **Snowflake** (`snowflake_destination_to_cache`): Plaintext passwords were being passed to `get_secret()` as if they were secret names. Now detects plaintext passwords (no `****`) and uses them directly.
- **BigQuery** (`bigquery_destination_to_cache`): Always fell back to `BIGQUERY_CREDENTIALS_PATH` env var. Now extracts plaintext `credentials_json` from the destination config, writes it to a temp file via `tempfile.mkstemp()`, and passes the path to `BigQueryCache`. Falls back to the env var only when credentials are obfuscated (cloud API path).

**`validate=False` on `set_config`**: `_prepare_destination_config()` now skips re-validation when toggling `disable_type_dedupe`, avoiding an unnecessary spec fetch + JSON schema check.

### E2E test results
| Destination | Status | Notes |
|---|---|---|
| Postgres | ✅ 14/14 tables | Docker-based, catalog namespace only |
| Snowflake | ✅ 14/14 tables | GSM creds, catalog namespace only |
| BigQuery | ✅ 14/14 tables | GSM creds; fixed backtick quoting + credentials_json handling |
| MotherDuck | ⏳ Blocked | Expired JWT in GSM |
| DuckDB | Skipped | Stale/non-compliant destination |

- [ ] **BigQuery temp file for credentials**: `bigquery_destination_to_cache()` writes `credentials_json` to a temp file via `mkstemp()` that is intentionally never deleted (the cache needs the path to remain valid). Verify this doesn't cause issues in long-running or repeated invocations (temp file accumulation). Consider whether a cleanup hook or `atexit` handler is warranted.
- [ ] **Obfuscation detection via `"****"`**: Both `snowflake_destination_to_cache()` and `bigquery_destination_to_cache()` detect obfuscated credentials by checking for `"****"` in the string. Verify this heuristic is reliable — could a legitimate password or credentials JSON contain that substring?
- [ ] **`dispose_engine()` after schema override**: In `destination_to_cache()`, `cache.processor.sql_config.dispose_engine()` is called after setting `schema_name` so the engine's `schema_translate_map` is refreshed. Verify this doesn't cause issues if the cache is used immediately after (engine should lazily recreate). Also confirm no other code paths set `schema_name` post-construction without disposing the engine.
- [ ] **`_get_column_stats` positional aliases**: The `nn_0`, `nn_1` pattern avoids PostgreSQL's 63-char truncation but assumes result columns are returned in declaration order. Verify this holds across BigQuery (backtick quoting), Snowflake (uppercases aliases to `NN_0`), and Postgres.
- [ ] **`destinationType` key stripping in `*_destination_to_cache()` functions**: All conversion functions now filter out `destinationType`/`DESTINATION_TYPE` before Pydantic model construction. Confirm no destination legitimately uses a config field with these exact names.
- [ ] **Recommended test plan**: Run `pyab destination-smoke-test --destination=destination-postgres --config=<config>` and verify JSON output includes populated `table_statistics` with row counts, column types, and null stats for all 14 streams. Also test against an unsupported destination to confirm graceful `table_statistics: null`.

### Notes
- No unit tests are included for the new readback methods. Adding tests against `DuckDBCache` with edge-case column names would be valuable as a follow-up.
- Table name resolution delegates to `cache.processor.get_sql_table_name()` with fallback to the original stream name, so naming conventions are driven by each cache's existing normalizer.
- MotherDuck e2e test is blocked on a refreshed JWT token in GSM (`SECRET_DESTINATION_DUCKDB__MOTHERDUCK__CREDS`).

## Summary by CodeRabbit

* **New Features**
  * Destination smoke tests now perform automatic readback introspection after successful writes for DuckDB, Postgres, Snowflake, BigQuery, and MotherDuck when supported.
  * Readback returns detailed statistics: table row counts, column names/types, and per-column null/non-null counts. If readback isn't supported or fails, it's skipped and reported without failing the smoke test.

* **Documentation**
  * CLI and tool docs updated to describe the new readback behavior and reported statistics.

Link to Devin session: https://app.devin.ai/sessions/f285ca4ab752412581eb228490d47e03
Requested by: @aaronsteers

---

<a href="https://app.devin.ai/review/airbytehq/pyairbyte/pull/1000" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/airbyte/caches/_utils/_dest_to_cache.py
+++ b/airbyte/caches/_utils/_dest_to_cache.py
@@ -3,6 +3,9 @@
 
 from __future__ import annotations
 
+import os
+import tempfile
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 from airbyte_api.models import (
@@ -104,15 +107,38 @@ def bigquery_destination_to_cache(
 
     We may have to inject credentials, because they are obfuscated when config
     is returned from the REST API.
+
+    When the destination config contains a plaintext `credentials_json` field
+    (the local `Destination.get_sql_cache()` path), the JSON is written to a
+    temporary file and used directly.  Otherwise we fall back to the
+    `BIGQUERY_CREDENTIALS_PATH` secret/env-var (the cloud API path).
     """
-    credentials_path = get_secret("BIGQUERY_CREDENTIALS_PATH")
+    # Extract credentials_json before converting to the Pydantic model,
+    # because DestinationBigquery may strip or obfuscate the field.
+    raw_credentials_json: str | None = None
     if isinstance(destination_configuration, dict):
+        raw_credentials_json = destination_configuration.get("credentials_json")
         filtered = {
             k: v
             for k, v in destination_configuration.items()
             if k not in {"destinationType", "DESTINATION_TYPE"}
         }
         destination_configuration = DestinationBigquery(**filtered)
+    elif hasattr(destination_configuration, "credentials_json"):
+        raw_credentials_json = destination_configuration.credentials_json
+
+    if raw_credentials_json and "****" not in raw_credentials_json:
+        # Plaintext credentials from a local destination config — write to
+        # a temp file so BigQueryCache can use it as a credentials path.
+        # The file is intentionally *not* deleted on close because the
+        # cache needs the path to remain valid after this function returns.
+        tmp_fd, tmp_path = tempfile.mkstemp(suffix=".json", prefix="bq_creds_")
+        Path(tmp_path).write_text(raw_credentials_json, encoding="utf-8")
+        # Close the file descriptor opened by mkstemp (write_text uses its own).
+        os.close(tmp_fd)
+        credentials_path = tmp_path
+    else:
+        credentials_path = get_secret("BIGQUERY_CREDENTIALS_PATH")
 
     return BigQueryCache(
         project_name=destination_configuration.project_id,


airbyte/_util/destination_readback.py
```diff
@@ -0,0 +1,534 @@
...
+# Deterministic table name resolution
+# ---------------------------------------------------------------------------
+
+# Destinations normalize stream names into SQL table names.  The logic
+# differs per destination.  We hard-code the known conventions here so
+# that we can *optimistically* compute the expected table name without
+# scanning the schema.
+
+
+def _normalize_table_name_default(stream_name: str) -> str:
```
aaronsteers: I think cache implementations already have a built-in specified normalizer. We should use that when available.

airbyte/_util/destination_readback.py
```diff
@@ -0,0 +1,534 @@
...
+def _resolve_expected_column_name(
+    destination_name: str,
+    column_name: str,
+) -> str:
+    """Deterministically resolve the expected SQL column name.
+
+    For now all destinations use the same LowerCaseNormalizer for columns.
+    """
+    _ = destination_name  # Reserved for per-destination overrides
+    return _normalize_column_name_default(column_name)
```
aaronsteers: Everything in this file above this line should be gotten from class members of the cache or SQL processor. All should already exist

airbyte/_util/destination_readback.py
```diff
@@ -0,0 +1,534 @@
...
+
+# ---------------------------------------------------------------------------
+# Core readback logic
+# ---------------------------------------------------------------------------
+
+
+def _query_table_row_count(
+    cache: CacheBase,
+    table_name: str,
+) -> int | None:
```
aaronsteers: This and anything else with cache as the first arg should probably be a private class member there.

airbyte/_util/destination_readback.py
```diff
@@ -0,0 +1,534 @@
...
+    },
+}
+
+SUPPORTED_DESTINATIONS: frozenset[str] = frozenset(_DESTINATION_TO_CACHE_INFO.keys())
+"""Destination connector names that support readback introspection."""
+
+
+def _get_readback_supported(destination_name: str) -> bool:
+    """Return True if readback is supported for the given destination."""
+    return destination_name in SUPPORTED_DESTINATIONS
```
aaronsteers: Likewise this can move into the cache class and/or local destination class.

airbyte/_util/destination_readback.py
aaronsteers, line 1: This file no longer adds value - it is now almost entirely consumed by documenting its own handoffs, with no benefit added that I can see

airbyte/_util/destination_readback.py
```diff
@@ -0,0 +1,277 @@
...
+            )
+        return result
+
+
+# ---------------------------------------------------------------------------
+# Readback orchestration
+# ---------------------------------------------------------------------------
+
+
+def run_destination_readback(
```
aaronsteers: I don't think we need this anymore. Should just be business logic inside the destination smoke test, on top of the generic cache methods

airbyte/destinations/base.py
```diff
@@ -61,6 +64,77 @@ def __init__(
...
+        destination_name: str | None = None,
+        destination_config: dict[str, Any] | None = None,
+        version: str | None = None,
+    ) -> CacheBase:
+        """Return a SQL Cache for querying data written by this destination.
+
+        This follows the same pattern as
+        :pymethod:`airbyte.cloud.sync_results.SyncResult.get_sql_cache`:
+        it builds a cache from the destination's configuration using
+        ``destination_to_cache()``.
```
aaronsteers: Should be markdown, not rest.

airbyte/caches/base.py
```diff
@@ -439,6 +439,210 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+    # ---- Readback introspection helpers ----
+    # These private methods support destination readback introspection,
+    # allowing the cache to query stats about data written by a destination.
+
+    def _readback_query_row_count(
+        self,
+        table_name: str,
+    ) -> int | None:
+        """Query the row count for a table. Returns None if the table doesn't exist."""
+        import logging  # noqa: PLC0415
```
aaronsteers: NO INLINE IMPORTS!!!

airbyte/caches/base.py
```diff
@@ -439,6 +441,212 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+    # allowing the cache to query stats about data written by a destination.
+
+    def _readback_quote_identifier(self, identifier: str) -> str:
+        """Quote an identifier for use in readback SQL queries.
+
+        Defaults to ANSI double-quote style.  Subclasses whose SQL dialect
+        uses a different quoting convention (e.g. BigQuery backticks) should
+        override this method.
+        """
+        return f'"{identifier}"'
```
aaronsteers: This exists already in sqlprocessor implementation. Delete and use existing where needed.

airbyte/caches/base.py
```diff
@@ -439,6 +441,212 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+
+    def _readback_query_row_count(
```
aaronsteers: Ditto. Push this down to SQLProcessor layer generically, with no 'readback' mention.

airbyte/caches/base.py
```diff
@@ -439,6 +439,130 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+            if result:
+                return int(result[0]["row_count"])
+            return 0  # noqa: TRY300
+        except Exception:
+            logging.getLogger(__name__).debug(
+                "Table %s.%s not found or not accessible.",
+                self.schema_name,
+                table_name,
+            )
+            return None
```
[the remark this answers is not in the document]
aaronsteers: Blanket try/catches are illegal. Don't do them. Let things fail unless you are catching something specific and expectd.

airbyte/caches/base.py
```diff
@@ -439,6 +441,212 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+
+    def _readback_query_column_info(
+        self,
+        table_name: str,
+    ) -> list[dict[str, str]]:
+        """Query column names and types for a table.
+
+        Returns a list of dicts with 'column_name' and 'column_type' keys.
+        """
+        try:
```
aaronsteers: Ditto. No blanket try.

airbyte/caches/base.py
```diff
@@ -439,6 +441,212 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+            return 0  # noqa: TRY300
+        except Exception:
+            logging.getLogger(__name__).debug(
+                "Table %s.%s not found or not accessible.",
+                self.schema_name,
+                table_name,
+            )
+            return None
+
+    def _readback_query_column_info(
```
aaronsteers: I believe we already have methods in sqlprocessor which returns types for columns. Use existing rather than rewriting.

airbyte/caches/base.py
```diff
@@ -439,6 +441,212 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+                    "column_name": col_name,
+                    "null_count": total_rows - non_null_count,
+                    "non_null_count": non_null_count,
+                    "total_count": total_rows,
+                }
+            )
+
+        return stats
+
+    def _readback_get_table_report(
```
aaronsteers: This helper us useless. Delete.

airbyte/caches/base.py
```diff
@@ -439,6 +441,212 @@ def __iter__(  # type: ignore [override]  # Overriding Pydantic model method
...
+
+        return {
+            "table_name": table_name,
+            "expected_stream_name": stream_name,
+            "row_count": row_count,
+            "columns": raw_columns,
+            "column_stats": raw_stats,
+        }
+
+    def _readback_get_results(
```
aaronsteers: Rename to `get_table_statistics` and make public. And return a map (dict) with stream name mapped to a `TableStatistics` dataclass or pydantic model.

airbyte/shared/sql_processor.py
```diff
@@ -1179,3 +1210,161 @@ def _table_exists(
...
+
+        For each stream, resolves the expected table name via the processor's
+        normalizer, queries row counts, column info, and per-column null/non-null
+        stats.
+
+        If the normalized table name is not found, falls back to the original
+        stream name (some destinations preserve original casing).
+
+        Returns a dict mapping stream name to a ``TableStatistics`` instance.
+        Streams whose tables are not found are omitted from the result.
```
aaronsteers: Markdown not rest. Everywhere.

airbyte/shared/sql_processor.py
```diff
@@ -84,6 +84,38 @@ class SQLRuntimeError(Exception):
...
+    """The SQL data type name as reported by the database."""
+
+    null_count: int
+    """Number of NULL values in this column."""
+
+    non_null_count: int
+    """Number of non-NULL values in this column."""
+
+    total_count: int
+    """Total row count (null_count + non_null_count)."""
```
aaronsteers: These ints should be nullable.

airbyte/shared/sql_processor.py
```diff
@@ -84,6 +84,38 @@ class SQLRuntimeError(Exception):
...
+
+
+class TableStatistics(BaseModel):
+    """Statistics for a single table: row count, column info, and per-column stats."""
+
+    table_name: str
+    """The table name as found in the destination."""
+
+    row_count: int
```
aaronsteers: Should be nullable

airbyte/shared/sql_processor.py
```diff
@@ -1179,3 +1210,161 @@ def _table_exists(
...
+        if row is None:
+            return 0
+        # Case-insensitive key lookup: Snowflake uppercases unquoted aliases.
+        row_ci = {k.lower(): v for k, v in row.items()}
+        return int(row_ci.get("row_count", 0))
+
+    def get_column_info(
+        self,
+        table_name: str,
+    ) -> list[dict[str, str]]:
```
aaronsteers: Don't we already have something like "get table schema"?

airbyte/destinations/base.py
```diff
@@ -61,6 +64,77 @@ def __init__(
...
+                (or `None`, which is treated as `"latest"`) is accepted.
+                Any other value raises `NotImplementedError`.
+
+        Raises:
+            NotImplementedError: If `version` is not `"latest"` or `None`.
+            ValueError: If the destination type is not supported.
+        """
+        from airbyte.destinations._translate_dest_to_cache import (  # noqa: PLC0415
+            destination_to_cache,
+        )
```
aaronsteers: This should not be imported inline. Only at top. If you have a circular dependency problem, tell me in slack.

airbyte/destinations/base.py
```diff
@@ -61,6 +64,77 @@ def __init__(
...
+        # Ensure the config carries a destinationType key so that
+        # destination_to_cache() can dispatch correctly.
+        if "destinationType" not in config and "DESTINATION_TYPE" not in config:
+            dest_type = resolved_name.replace(_CANONICAL_PREFIX, "")
+            config["destinationType"] = dest_type
+
+        cache: CacheBase = destination_to_cache(config)
+
+        if schema_name is not None:
+            cache = cache.model_copy(update={"schema_name": schema_name})
```
aaronsteers: This is bad. Don't do it. A cache is a heavy object which should not be copied. Just create it correctly the first time by passing schema_name to the `destinaton_to_cache` implementation.

airbyte/shared/sql_processor.py
```diff
@@ -84,6 +84,38 @@ class SQLRuntimeError(Exception):
...
+
+    total_count: int | None = None
+    """Total row count (null_count + non_null_count)."""
+
+
+class TableStatistics(BaseModel):
+    """Statistics for a single table: row count, column info, and per-column stats."""
+
+    table_name: str
+    """The table name as found in the destination."""
```
aaronsteers, line 110: Need to include the database name and schema name as properties here also in the table statistics model.

airbyte/shared/sql_processor.py
aaronsteers: You should state this in the docstring then:

> This method differs from `_get_sql_column_definitions` in that it always returns actual detected column types from the database. It will never return previously-cached types or 'expected' types based on the stream json schema.

Devin, also, since introspection can be expensive, we should consider modifying the method signature to operate on a list of input tables rather than a single table, so that the inspection result can be used for all tables that need it. (Unless you are positive is already optimized in its current form.)

airbyte/shared/sql_processor.py
```diff
@@ -1179,3 +1216,163 @@ def _table_exists(
...
+                "column_type": str(col["type"]),
+            }
+            for col in columns
+        ]
+
+    def get_column_stats(
+        self,
+        table_name: str,
+        columns: list[dict[str, str]],
+    ) -> list[dict[str, Any]]:
```
aaronsteers, line 1278: Returning `Any` as part of the return type for a public function is completely unacceptable. Needs to give a strongly typed return or else it needs to be made private.

airbyte/shared/sql_processor.py
```diff
@@ -1179,3 +1216,179 @@ def _table_exists(
...
+                    "column_name": col_name,
+                    "null_count": total_rows - non_null_count,
+                    "non_null_count": non_null_count,
+                    "total_count": total_rows,
+                }
+            )
+
+        return stats
+
+    def get_table_statistics(
```
aaronsteers: Let's use `fetch` as a more descriptive verb for what is happening here. `fetch_table_statistics` is more descriptive because it properly communicates the cost of retrieval, where 'get' might imply returning values we already have access to.

airbyte/_util/destination_smoke_tests.py
```diff
@@ -194,6 +295,12 @@ def _sanitize_error(ex: Exception) -> str:
     return f"{type(ex).__name__}: {ex}"
 
 
+def _get_stream_names_from_source(source_obj: Source) -> list[str]:
+    """Extract stream names from a configured source."""
+    catalog = source_obj.get_configured_catalog()
+    return [stream.stream.name for stream in catalog.streams]
```
aaronsteers: Don't we already have a Source.selected_streams helper or similar? Delete this helper and just call the appropriate method(s)/property(ies) on the Source directly.

airbyte/_util/destination_smoke_tests.py
```diff
@@ -267,6 +373,33 @@ def run_destination_smoke_test(
 
     elapsed = time.monotonic() - start_time
 
+    # Perform readback introspection if the write succeeded
+    readback_result: DestinationReadbackResult | None = None
+    if success:
+        try:
+            cache = destination.get_sql_cache(schema_name=namespace)
+            table_statistics = cache.fetch_table_statistics(stream_names)
+            tables_missing = [name for name in stream_names if name not in table_statistics]
```
aaronsteers: Rename `tables_missing` with `tables_not_found`. Then delete all this extra code related to `DestinationReadbackResult` as well as the declaration of the `DestinationReadbackResult` type itself, and simply add `tables_not_found` and `table_statistics` to the `DestinationSmokeTestResult` model as its returned. This will remove a ton of useless code.

aaronsteers: Also, make `tables_not_found` into a mapping (dict[str, str]) where stream_name is mapped to expected (not found) table name.

airbyte/_util/destination_smoke_tests.py
```diff
@@ -267,6 +302,31 @@ def run_destination_smoke_test(
...
+    tables_not_found: dict[str, str] | None = None
+    if success:
+        try:
+            cache = destination.get_sql_cache(schema_name=namespace)
+            table_statistics = cache.fetch_table_statistics(stream_names)
+            tables_not_found = {
+                name: cache.processor.get_sql_table_name(name)
+                for name in stream_names
+                if name not in table_statistics
+            }
```
aaronsteers, line 359: Remove this try block and instead (1) add `Destination.is_cache_supported` as a simple boolean 'get' property, then (2) you can check early and exit/skip gracefully if you know in advance that the destination doesn't support the cache interface.

airbyte/_util/destination_smoke_tests.py
```diff
@@ -267,6 +302,23 @@ def run_destination_smoke_test(
 
     elapsed = time.monotonic() - start_time
 
+    # Perform readback introspection if the write succeeded
+    table_statistics: dict[str, TableStatistics] | None = None
+    tables_not_found: dict[str, str] | None = None
+    if success and destination.is_cache_supported:
```
aaronsteers: I don't think we should short-circuit when `success == True`. There could be partial success and stats could still be helpful in those cases.

airbyte/_util/destination_smoke_tests.py
```diff
@@ -267,6 +302,23 @@ def run_destination_smoke_test(
...
+        tables_not_found = {
+            name: cache.processor.get_sql_table_name(name)
+            for name in stream_names
+            if name not in table_statistics
+        }
+    else:
+        logger.info(
+            "Readback not supported for destination '%s'.",
+            destination.name,
+        )
```
aaronsteers, line 372:
```suggestion
        logger.info(
            "Skipping table and column statistics retrieval for "
            "destination '%s' because no SQL interface mapping has been "
            "defined.",
            destination.name,
        )
```

airbyte/destinations/base.py
```diff
@@ -61,6 +64,90 @@ def __init__(
...
+            self.name,
+        ).replace(_CANONICAL_PREFIX, "")
+        return dest_type in get_supported_destination_types()
+
+    def get_sql_cache(
+        self,
+        *,
+        schema_name: str | None = None,
+        destination_name: str | None = None,
+        destination_config: dict[str, Any] | None = None,
```
aaronsteers: Should we delete destination_name and destination_config as inputs? Now that this lives entirely as a method on the Destination itself? Likewise, I think we can remove the version input since now the Destination will know its own version and can handle resolution internally.

PR #944, opened 2026-01-18
ci: update to semantic-pr-release-drafter@v0.4.0 and simplify publish workflow

aaronsteers:
## Summary

This PR makes two related changes to the release and publish workflows:

**1. Release Drafter Update (release_drafter.yml)**
- Updated to `aaronsteers/semantic-pr-release-drafter@v0.4.0` with new `attach-files` feature
- Added `workflow_dispatch` trigger for manual runs
- Simplified workflow to call the action only once (after building): install uv → checkout → `uv build` → release-drafter with attach-files
- Relies on uv-dynamic-versioning to determine version from git (no explicit version bypass)

**2. Publish Workflow Simplification (pypi_publish.yml)**
- Drastically simplified to a single job that downloads pre-built assets from the release and publishes to PyPI
- Removed `build` job entirely (building now happens in release-drafter)
- Removed `workflow_dispatch` trigger (prereleases should use release-drafter workflow)
- Only triggers on `release: published`
- Uses `robinraju/release-downloader@v1.12` with `fileName: "*"` for declarative asset download
- Simplified permissions from `contents: write` to `contents: read`

## Updates since last revision

- Simplified release-drafter to call the action only once (after building), matching the awesome-python-template pattern
- Removed explicit version bypass - now relies on uv-dynamic-versioning to determine version from git
- Removed unused step ID from release-drafter workflow
- Consolidated asset download to single step with `fileName: "*"` pattern
- Added OIDC comment for clarity

## Review & Testing Checklist for Human

- [ ] **Verify uv-dynamic-versioning works correctly** - The workflow no longer uses `POETRY_DYNAMIC_VERSIONING_BYPASS`; confirm `uv build` produces correctly versioned artifacts
- [ ] Verify release-drafter correctly builds and attaches wheel/sdist to draft releases
- [ ] Test the full release flow: merge to main → release-drafter creates draft with assets → publish release → PyPI publish downloads assets and publishes
- [ ] Confirm the publish workflow fails gracefully if assets are missing from the release

### Test Plan
1. Merge this PR
2. Push a commit to main and verify release-drafter creates/updates a draft release with wheel and tar.gz attached (check the version in the artifact names)
3. Publish the draft release and verify the publish workflow downloads the pre-built assets and publishes to PyPI

### Notes

This is part of a batch migration across multiple repos to:
1. Use the new asset attachment feature in semantic-pr-release-drafter v0.4.0
2. Simplify publish workflows to only download and publish (no building)
[...]

PR #1044, opened 2026-06-06
chore: widen cryptography constraint and bump CDK to 7.21.1

aaronsteers:
## Summary

Widens `cryptography` from `>=44.0.0,<45.0.0` to `>=44.0.0,<47.0.0,!=45.0.0,!=45.0.1` — the negation constraints exclude 45.0.0/45.0.1 which had `load_pem_private_key()` regressions (fixed in 45.0.2+). PyAirbyte imports `cryptography` directly in `airbyte/_processors/sql/snowflake.py` for key-pair auth, so we keep the explicit dependency.

Bumps `airbyte-cdk` lower bound from `>=7.3.9` to `>=7.21.1` — CDK 7.21.1 includes the same widened cryptography constraint, so the two packages no longer conflict on resolution.

Restores `workflow_dispatch` trigger to `pypi_publish.yml` (removed by #944), which broke the `/prerelease` slash command. Adds security hardening (`persist-credentials: false`, conditional cache disable) and uses SHA-pinned action references.

Unblocks `boring-semantic-layer>=0.3.0` in downstream projects (specifically `airbytehq/airbyte-transforms-poc`) which transitively requires `cryptography>=45.0.3`.

Link to Devin session: https://app.devin.ai/sessions/9f900ef4021147adb369727617d08827
Requested by: @aaronsteers

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dependencies = [
     "airbyte-api>=0.53.0,<1.0",
     "airbyte-cdk>=7.21.0.post4.dev27052511259,<8.0.0",
     "airbyte-protocol-models-pdv2>=0.13.0,<1.0",
-    "cryptography>=44.0.0,<47.0.0,!=45.0.0,!=45.0.1",
+    "cryptography>=44.0.0,<47.0.0,!=45.0.0,!=45.0.1",  # 45.0.0/45.0.1 broke load_pem_private_key()
     "cyclopts>=4.0,<5.0",
     # TODO: Restore broader version constraints after verifying compatibility
     # "duckdb>=1.4.0,<2.0",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ dependencies = [
     "airbyte-api>=0.53.0,<1.0",
     "airbyte-cdk>=7.3.9,<8.0.0",
     "airbyte-protocol-models-pdv2>=0.13.0,<1.0",
-    "cryptography>=44.0.0,<45.0.0",
+    "cryptography>=44.0.0",
     "cyclopts>=4.0,<5.0",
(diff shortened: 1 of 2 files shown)


aaronsteers:
/prerelease
> **Prerelease Build Started**
>
> Building and publishing prerelease package from this PR...
> [Check job output.](https://github.com/airbytehq/PyAirbyte/actions/runs/27054033685)
> **Prerelease Published to PyPI**
>
> Version: `0.47.0.dev104427054033685`
> [View publish workflow](https://github.com/airbytehq/PyAirbyte/actions/runs/27054041577)
>
> Install with:
> ```bash
> pip install airbyte==0.47.0.dev104427054033685
> ```

pyproject.toml
```diff
@@ -9,9 +9,9 @@ requires-python = ">=3.10,<3.13"
 dynamic = ["version"]
 dependencies = [
     "airbyte-api>=0.53.0,<1.0",
-    "airbyte-cdk>=7.3.9,<8.0.0",
+    "airbyte-cdk>=7.21.0.post4.dev27052511259,<8.0.0",
```
aaronsteers: CDK update just release. You can update to stable now.

pyproject.toml
```diff
@@ -9,9 +9,9 @@ requires-python = ">=3.10,<3.13"
...
     "airbyte-protocol-models-pdv2>=0.13.0,<1.0",
-    "cryptography>=44.0.0,<45.0.0",
+    "cryptography>=44.0.0,<47.0.0,!=45.0.0,!=45.0.1",  # 45.0.0/45.0.1 broke load_pem_private_key()
```
aaronsteers, line 14: Do we actually import/use cryptography directly? We can drop the explicit declaration if so, and rely on CDK for the package.

PR #1067, opened 2026-07-07
chore(deps): Upgrade snowflake connector to prevent security issues

ohadmata:
The existing Snowflake connector specified in the pyproject file (<4.0) uses an old cryptography/OpenSSL packages that has some security issues:

https://nvd.nist.gov/vuln/detail/CVE-2026-34073
https://nvd.nist.gov/vuln/detail/CVE-2026-26007
https://github.com/advisories/ghsa-537c-gmf6-5ccf

The idea is to allow snowflake-connector  < 5 To prevent these issues

## Summary by CodeRabbit

* **Chores**
  * Expanded the supported version range for the Snowflake connector, allowing compatibility with newer releases while keeping the current minimum version unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -36,7 +36,7 @@ dependencies = [
     "pyyaml>=6.0.2,<7.0",
     "requests!=3.32.0",  # 3.32.0 breaks docker tests
     "rich>=13.7.0,<14.0",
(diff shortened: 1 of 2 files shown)


ohadmata:
It looks like I need some permissions from the Airbyte team to deploy the Vercel app:

image

PR #1077, opened 2026-07-14
fix(mcp): resolve `get_connector_info` spec from registry instead of installing

aaronsteers:
## Summary

`get_connector_info` could hang for minutes on the hosted Cloud MCP server. On a
hosted, no-Docker runtime, populating `config_spec_jsonschema` fell through to a
fresh pip/venv install of the connector on **every call** — an unbounded operation
that has stalled requests for over four minutes.

PyAirbyte #1052 gated the `local.py` filesystem/config tools with
`requires_client_filesystem`, but `get_connector_info` in `registry.py` was never
gated, so it stayed exposed on hosted Cloud MCP and kept running the install path.

This gates only the expensive step on Docker availability, so the tool still returns
the fast registry metadata (docs URL, manifest URL, connector metadata) everywhere,
and simply omits `config_spec_jsonschema` when it can't be resolved cheaply:

```python
config_spec_jsonschema = None
if is_docker_installed():
    # Docker => fast image pull. No Docker (hosted) => unbounded pip/venv install.
    with contextlib.suppress(Exception):
        connector.install()
        config_spec_jsonschema = connector.config_spec
```

Behavior after the fix:
- Local with Docker: unchanged — installs (fast image pull) and returns `config_spec_jsonschema`.
- Hosted / no Docker: skips the install, returns `config_spec_jsonschema=None`, no hang.

`config_spec_jsonschema` is already `Optional`, so returning `None` is within the
existing contract. `is_docker_installed()` is already imported and used two lines
above to choose `docker_image=`.

Added a parametrized regression test asserting `connector.install()` is called only
when Docker is available and `config_spec_jsonschema` is `None` otherwise.

Link to Devin session: https://app.devin.ai/sessions/9274bd4f3e0a4b0880795fc1ef9c806b
Requested by: @aaronsteers

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connector information retrieval by resolving configuration schemas from the public connector registry (preferring the cloud spec, falling back to OSS).
  * When no schema is available, `config_spec_jsonschema` is now returned as `None` rather than relying on Docker-based execution.
  * Prevents unnecessary connector setup steps in Docker-restricted environments while keeping metadata such as the manifest URL available.
* **Tests**
  * Added/updated unit tests to ensure schema resolution is driven by registry lookups (cloud → OSS → none) and that connector installation is not invoked.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

### DIFF
--- a/airbyte/mcp/registry.py
+++ b/airbyte/mcp/registry.py
@@ -170,13 +170,18 @@ def get_connector_info(
     # Resolve the config spec from the public registry endpoint keyed by version,
     # which works in any runtime (hosted or local) without installing the
     # connector. Prefer the `cloud` spec and fall back to `oss`.
+    version = connector_metadata.latest_available_version if connector_metadata else None
     config_spec_jsonschema: dict[str, Any] | None = get_connector_spec_from_registry(
         connector_name,
+        version=version,
         platform="cloud",
-    ) or get_connector_spec_from_registry(
-        connector_name,
-        platform="oss",
     )
+    if config_spec_jsonschema is None:
+        config_spec_jsonschema = get_connector_spec_from_registry(
+            connector_name,
+            version=version,
+            platform="oss",
+        )
 
     manifest_url = _DEFAULT_MANIFEST_URL.format(
         source_name=connector_name,
--- a/airbyte/mcp/registry.py
+++ b/airbyte/mcp/registry.py
@@ -157,9 +157,11 @@ def get_connector_info(
     if connector_name not in get_available_connectors():
         return "Connector not found."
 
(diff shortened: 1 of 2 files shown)


aaronsteers:
Is there a different approach we can take? Now that we have, I think, other code in production that is able to get config information from GCS-hosted public HTTP endpoints, only needing to have the version string to calculate the spec based on a stored OSS and cloud spec file hosted in GCP and available via public endpoint. 

You can check this repo, and you can check the ops repo for the working pattern, but I think we can just always avoid calling docker here if all we need docker for is to get the spec.json

PR #1078, opened 2026-07-14
feat(mcp): reuse caller's verified token for Airbyte Cloud API passthrough

aaronsteers:
## Summary

The Cloud MCP (hosted at `cloud-mcp`) should call the Airbyte Cloud API as the **caller**, not solely as a statically-configured application. When an agent authenticates the MCP transport with an Airbyte-Cloud-issued bearer token (verified against the `_airbyte-application-clients` realm when `MCP_AUTH_AIRBYTE_CLOUD=true`), that same JWT is itself a valid Cloud API bearer — so it should flow through to downstream Cloud calls, delegating them to the caller's identity. Configured `client_id`/`client_secret` remain the fallback for stdio/local mode where there is no caller token.

The transport-side wiring for this already landed in #1073 (`_resolve_transport_bearer_token` → `BEARER_TOKEN_CONFIG_ARG` defaulting to the verified transport token, plus the Airbyte Cloud realm defaults). The **missing piece** was the "bearer wins over client creds" resolution at the point the MCP tools build a `CloudClient`.

`_get_cloud_client` read all three of `bearer_token`, `client_id`, `client_secret` from MCP config and passed them together to `CloudClient`. But `CloudClient` (via `_AirbyteCredentials.from_auth`) rejects receiving both a bearer token and client credentials at once. In hosted mode — transport token present *and* `AIRBYTE_CLOUD_CLIENT_ID`/`SECRET` set in the environment — this raised `Cannot use both client credentials and bearer token authentication`, breaking the passthrough.

This PR makes the MCP client-construction helper prefer the bearer token, mirroring Ops MCP's `_get_access_token` ("if `bearer_token` is provided, use it directly; otherwise exchange client credentials"):

```python
# airbyte/mcp/cloud.py :: _get_cloud_client
bearer_token = get_mcp_config(ctx, MCP_CONFIG_BEARER_TOKEN) or None
client_id = get_mcp_config(ctx, MCP_CONFIG_CLIENT_ID) or None
client_secret = get_mcp_config(ctx, MCP_CONFIG_CLIENT_SECRET) or None
...
if bearer_token:            # bearer wins; CloudClient rejects both at once
    client_id = None
    client_secret = None
return CloudClient(client_id=..., client_secret=..., bearer_token=bearer_token, ...)
```

### How it mirrors Ops MCP

- **Realm-verified token + `BEARER_TOKEN` config arg defaulting to the transport token** — already present in PyAirbyte from #1073 (`server.py::_create_auth`, `_tool_utils.py::_resolve_transport_bearer_token` / `BEARER_TOKEN_CONFIG_ARG`), identical in shape to `airbyte_ops_mcp/mcp/server.py`.
- **`_get_access_token`-style "bearer wins over client creds"** — Ops MCP centralizes this in `cloud_admin/api_client.py::_get_access_token`. In PyAirbyte the low-level `api_util.get_airbyte_server_instance` intentionally requires *exactly one* of bearer vs. client creds (a documented `CloudClient`/`CloudClientConfig` invariant), so the analogous, scoped place to apply "bearer wins" is the MCP client-construction helper `_get_cloud_client`. This keeps the change confined to the MCP auth layer without altering the public Cloud client's mutual-exclusivity contract.

### Expected behavior

- **Hosted (`cloud-mcp`)**: caller's verified Airbyte Cloud token → used directly as the Cloud API bearer (delegated identity). Env client creds present but ignored while a token is passed.
- **stdio / local**: no transport token (`_resolve_transport_bearer_token` returns `""`) → falls back to `AIRBYTE_CLOUD_CLIENT_ID`/`AIRBYTE_CLOUD_CLIENT_SECRET`.

Requested by AJ Steers (@aaronsteers).

## Test Plan

Added focused unit tests:

- `tests/unit_tests/test_mcp_cloud.py::test_get_cloud_client_prefers_bearer_token_over_client_credentials` — parametrized: bearer token present → `CloudClient` gets the bearer and `client_id`/`client_secret` are dropped (no mutual-exclusivity error); empty bearer (stdio) → falls back to client creds.
- `tests/unit_tests/test_mcp_tool_utils.py::test_resolve_transport_bearer_token` — parametrized: `get_access_token()` returns `None` (stdio) or an empty token → `""`; a verified token → returned as-is.

Local checks (all green):

- `uv run ruff format --check .` (changed files)
- `uv run ruff check .` (changed files)
- `uv run pyrefly check` (0 errors; the 3 warnings are pre-existing `redundant-cast` in `cloud.py`)
- `uv run pytest tests/unit_tests/test_mcp_cloud.py tests/unit_tests/test_mcp_tool_utils.py` → 18 passed

Link to Devin session: https://app.devin.ai/sessions/a8d355a69ca34409bd3b13d2a8122123

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication handling for cloud connections by prioritizing bearer-token authentication when available.
  * Added a reliable fallback to client ID and secret authentication when no bearer token is provided.
  * Improved handling of empty or unavailable access tokens.

* **Tests**
  * Added coverage for authentication precedence and access-token resolution scenarios.

### DIFF
--- a/airbyte/mcp/cloud.py
+++ b/airbyte/mcp/cloud.py
@@ -267,6 +267,16 @@ def _get_cloud_workspace(
     return _get_cloud_client(ctx).get_workspace(resolved_workspace_id)
 
 
+def _none_if_empty(value: str) -> str | None:
+    """Return `value` unless it is empty, in which case return `None`.
+
+    Emptiness is checked with `len` rather than truthiness because a
+    `SecretString` overrides `__bool__` to always be truthy, so an empty secret
+    would otherwise be mistaken for a present value.
+    """
+    return value if len(value) > 0 else None
+
+
 def _get_cloud_client(
     ctx: Context,
     *,
@@ -283,20 +293,23 @@ def _get_cloud_client(
     receiving both at once, so the client credentials are dropped when a bearer
     token is available.
     """
-    bearer_token = get_mcp_config(ctx, MCP_CONFIG_BEARER_TOKEN) or None
-    client_id = get_mcp_config(ctx, MCP_CONFIG_CLIENT_ID) or None
-    client_secret = get_mcp_config(ctx, MCP_CONFIG_CLIENT_SECRET) or None
+    bearer_token = _none_if_empty(get_mcp_config(ctx, MCP_CONFIG_BEARER_TOKEN))
+    client_id = _none_if_empty(get_mcp_config(ctx, MCP_CONFIG_CLIENT_ID))
+    client_secret = _none_if_empty(get_mcp_config(ctx, MCP_CONFIG_CLIENT_SECRET))
     api_url = get_mcp_config(ctx, MCP_CONFIG_API_URL)
     config_api_url = get_mcp_config(ctx, MCP_CONFIG_CONFIG_API_URL)
 
-    if bearer_token:
-        client_id = None
-        client_secret = None
+    if bearer_token is not None:
+        return CloudClient(
+            bearer_token=bearer_token,
+            public_api_root=api_url,
+            config_api_root=config_api_url,
+            organization_id=organization_id,
+        )
 
     return CloudClient(
         client_id=client_id,
         client_secret=client_secret,
-        bearer_token=bearer_token,
         public_api_root=api_url,
         config_api_root=config_api_url,
         organization_id=organization_id,
--- a/airbyte/mcp/cloud.py
+++ b/airbyte/mcp/cloud.py
@@ -272,12 +272,26 @@ def _get_cloud_client(
     *,
     organization_id: str | None = None,
 ) -> CloudClient:
-    """Get an authenticated `CloudClient` from MCP config."""
-    bearer_token = get_mcp_config(ctx, MCP_CONFIG_BEARER_TOKEN)
-    client_id = get_mcp_config(ctx, MCP_CONFIG_CLIENT_ID)
-    client_secret = get_mcp_config(ctx, MCP_CONFIG_CLIENT_SECRET)
+    """Get an authenticated `CloudClient` from MCP config.
+
+    A bearer token takes precedence over `client_id`/`client_secret`. When the
+    server verifies transport against Airbyte Cloud's realm, the verified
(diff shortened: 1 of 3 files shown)


airbyte/mcp/cloud.py
```diff
@@ -272,13 +272,27 @@ def _get_cloud_client(
...
+    bearer_token = get_mcp_config(ctx, MCP_CONFIG_BEARER_TOKEN) or None
+    client_id = get_mcp_config(ctx, MCP_CONFIG_CLIENT_ID) or None
+    client_secret = get_mcp_config(ctx, MCP_CONFIG_CLIENT_SECRET) or None
     api_url = get_mcp_config(ctx, MCP_CONFIG_API_URL)
     config_api_url = get_mcp_config(ctx, MCP_CONFIG_CONFIG_API_URL)
 
+    if bearer_token:
+        client_id = None
+        client_secret = None
+
```
aaronsteers: Before we throw away the client ID and client secret, should we check if the bearer token is even valid? If it's not valid, then using the client ID and secret instead to get a new token would be the preferred approach, rather than failing hard when all three are provided

airbyte/mcp/cloud.py
```diff
@@ -267,22 +267,49 @@ def _get_cloud_workspace(
...
     api_url = get_mcp_config(ctx, MCP_CONFIG_API_URL)
     config_api_url = get_mcp_config(ctx, MCP_CONFIG_CONFIG_API_URL)
 
+    if bearer_token is not None:
+        return CloudClient(
+            bearer_token=bearer_token,
+            public_api_root=api_url,
+            config_api_root=config_api_url,
+            organization_id=organization_id,
+        )
```
aaronsteers, line 308: I don't understand why we shouldn't keep the prior implementation and make CloudClient handle this for us.

```
    return CloudClient(
        client_id=client_id,
        client_secret=client_secret,
        bearer_token=bearer_token,
        public_api_root=api_url,
        config_api_root=config_api_url,
        organization_id=organization_id,
    )
```

I also don't understand the claim that the CloudClient class would fail auth or fail to auth in the prior implementaton. Everything appears to have been correct before this PR.

PR #1082, opened 2026-07-21
feat(mcp): wire trusted-execution gate and function-layer guards

aaronsteers:
## Summary

Wires the Airbyte MCP server into the generic secure-by-default trusted-execution gate landed in `fastmcp-extensions` v0.10.3 (airbytehq/fastmcp-extensions#71), and adds an **independent function-layer** of enforcement so trusted-machine capabilities can never be reached by an untrusted (e.g. hosted HTTP) caller — even if a future registration mistake left a tool visible.

This is PR2 of the MCP security hardening effort. It maps to remediations **R1** (trusted-execution gate), **R3** (server-side secret suppression / refusal), and the domain-config part of **R6** for the public `cloud-mcp` threat model. No infrastructure (Cloud Run / Pulumi), OAuth-client, or network-egress changes are in scope here.

### Two independent layers of defense

1. **Visibility layer (from PR1):** registering a `trusted_execution` config arg lets `fastmcp-extensions` hide `requires_client_filesystem` tools when the gate is off.
2. **Function layer (new here):** backend helpers call `require_trusted_execution(...)` themselves and hard-fail when the gate is off, independently of tool visibility.

Because the layers are independent, a mistake in either one alone cannot expose a trusted-machine capability.

### What changed

1. **Generic gate mapped to the Airbyte env var.** New `TRUSTED_EXECUTION_CONFIG_ARG` maps the generic `trusted_execution` config `name` to `AIRBYTE_MCP_TRUSTED_EXECUTION` (default `"0"`), keeping the generic library Airbyte-agnostic. Registered in `server.py`. It has **no** `http_header_key` — the gate *widens* the surface, so it must never be caller-controllable.

   ```python
   TRUSTED_EXECUTION_CONFIG_ARG = MCPServerConfigArg(
       name=CONFIG_TRUSTED_EXECUTION,      # generic name from fastmcp_extensions
       env_var=MCP_TRUSTED_EXECUTION_ENV_VAR,  # "AIRBYTE_MCP_TRUSTED_EXECUTION"
       default="0",
       required=False,
   )
   ```

2. **Permanent HTTP prohibition.** `http_main.main()` calls `assert_http_trusted_execution_disabled(app)` immediately before `app.run(...)`, hard-failing startup if trusted execution was explicitly enabled on the hosted entrypoint.

3. **Reusable function-layer guard** (`airbyte/mcp/_guards.py`):

   ```python
   def is_trusted_execution_enabled() -> bool: ...          # reads env only
   def require_trusted_execution(feature: str) -> None:     # raises when off
   ```

   Raises `PyAirbyteTrustedExecutionRequiredError` (new, under a new `PyAirbyteMCPError` base) with remediation in `guidance`, not in the deterministic message.

4. **Guarded backend helpers.** Guards were added so a *direct* call hard-fails when untrusted:
   - `_arg_resolvers.resolve_connector_config` — gates only the dangerous branches (`config_file`, `config_secret_name`, and inline `secret_reference::` values). An ordinary inline `config` dict is **still allowed** untrusted, because `cloud.py` legitimately resolves caller-provided config that way.
   - `local._get_mcp_source` (connector install/execution) and the local cache / secret-inventory / destructive tools: `list_cached_streams`, `describe_default_cache`, `run_sql_query`, `list_dotenv_secrets`, `list_connector_config_secrets`, `destination_smoke_test`.

5. **Secret-manager suppression when untrusted** (`_config.load_secrets_to_env_vars`). Dotenv and Google Secret Manager backends are no longer *registered as secret sources* when trusted execution is off, so `config_secret_name` / `secret_reference::` cannot resolve server-side secrets. Ordinary environment variables are still loaded from any referenced dotenv file so safe, non-secret-manager config keeps working.

6. **Domain-config hard-fail** (`_tool_utils.validate_airbyte_domains`, called after tool registration in `server.py`). Instead of silently dropping a requested domain, it hard-fails on (a) setting both `AIRBYTE_MCP_DOMAINS` and `AIRBYTE_MCP_DOMAINS_DISABLED`, and (b) naming a domain with no registered tools. Remediation names the setting to clear and says to restart.

## Test plan

New `tests/unit_tests/test_mcp_trusted_execution.py` (31 cases), covering:
1. env parsing / default-off for the gate;
2. `require_trusted_execution` raise-when-off / pass-when-on;
3. `resolve_connector_config`: inline config allowed untrusted; `config_file`, `config_secret_name`, and inline `secret_reference::` rejected untrusted; `config_file` + `secret_reference::` allowed when trusted;
4. parametrized function-layer rejection for all guarded `local` helpers when untrusted;
5. `assert_http_trusted_execution_disabled` raises when trusted is on, passes when off;
6. secret-manager suppression when untrusted (env var still loaded) vs. registration when trusted;
7. `validate_airbyte_domains` include+exclude conflict, unknown-domain, and valid-single-domain paths.

Local verification (uv):
```
uv run ruff format --check .   # 219 files already formatted
uv run ruff check .            # all checks passed
uv run pyrefly check           # 0 errors (2 preexisting warnings in cloud.py)
uv run pytest tests/unit_tests/test_mcp_trusted_execution.py   # 31 passed
uv run pytest tests/unit_tests/test_mcp_cloud.py test_mcp_tool_utils.py test_mcp_connector_registry.py  # 47 passed
```

## Limitations / out of scope

1. No Cloud Run / Pulumi, OAuth-client, or metadata-egress changes (separate follow-up PRs R2/R5 and infra wiring).
2. The generic `fastmcp-extensions` read API is used as-is; no new raising helper was added there (per maintainer direction).
3. The CI regression guard + `CONTRIBUTING.md` guidance (R8) is a separate follow-up PR.

https://app.devin.ai/sessions/1696a2c620134d61a5255f959ef4c4ef

Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Added a trusted-execution environment switch for MCP, covering connector config handling, secret hydration, dotenv secret processing, cache tools, and local SQL/destination checks.
  * Added startup validation for MCP domain include/exclude configuration.
* **Security**
  * Enforced hard failures for untrusted contexts across local privileged MCP capabilities.
  * Blocked enabling trusted execution via HTTP-exposed startup.
* **Bug Fixes**
  * Introduced MCP-specific error types for trusted-execution requirements and MCP server failures.
* **Tests**
  * Added unit tests covering trusted-execution gating, connector config restrictions, HTTP enforcement, secret/env loading behavior, and domain validation.

### DIFF
--- a/airbyte/exceptions.py
+++ b/airbyte/exceptions.py
@@ -222,6 +222,34 @@ class PyAirbyteNoStreamsSelectedError(PyAirbyteInputError):
     available_streams: list[str] | None = None
 
 
+# MCP Server Errors
+
+
+@dataclass
+class PyAirbyteMCPError(PyAirbyteError):
+    """An error occurred in the PyAirbyte MCP server."""
+
+
+@dataclass
+class PyAirbyteTrustedExecutionRequiredError(PyAirbyteMCPError):
+    """A trusted-execution-only capability was invoked while trusted execution is disabled.
+
+    Trusted execution grants the MCP server its trusted-machine capabilities: local
+    filesystem access, local connector installation/execution, and server-side secret
+    resolution. It defaults to *off* on every transport and is permanently unavailable
+    over the HTTP transport, so a backend helper that exposes one of those capabilities
+    hard-fails when the gate is disabled -- independently of whether the corresponding
+    tool was hidden from the tool listing.
+    """
+
+    guidance = (
+        "Set `AIRBYTE_MCP_TRUSTED_EXECUTION=1` on the MCP server process and restart it. "
+        "Trusted execution is only available on the stdio transport; it can never be "
+        "enabled for an HTTP/hosted deployment."
+    )
+    feature: str | None = None
+
+
 # Normalization Errors
 
 
--- a/airbyte/exceptions.py
+++ b/airbyte/exceptions.py
@@ -227,7 +227,7 @@ class PyAirbyteNoStreamsSelectedError(PyAirbyteInputError):
 
 @dataclass
 class AirbyteMCPError(PyAirbyteError):
-    """An error occurred in the PyAirbyte MCP server."""
+    """An error occurred in the Airbyte MCP server."""
 
 
 @dataclass
--- a/airbyte/exceptions.py
+++ b/airbyte/exceptions.py
@@ -226,12 +226,12 @@ class PyAirbyteNoStreamsSelectedError(PyAirbyteInputError):
 
 
 @dataclass
-class PyAirbyteMCPError(PyAirbyteError):
+class AirbyteMCPError(PyAirbyteError):
     """An error occurred in the PyAirbyte MCP server."""
 
 
 @dataclass
-class PyAirbyteTrustedExecutionRequiredError(PyAirbyteMCPError):
+class AirbyteTrustedExecutionRequiredError(AirbyteMCPError):
     """A trusted-execution-only capability was invoked while trusted execution is disabled.
 
     Trusted execution grants the MCP server its trusted-machine capabilities: local
--- a/airbyte/mcp/_arg_resolvers.py
+++ b/airbyte/mcp/_arg_resolvers.py
@@ -14,7 +14,7 @@ from typing import Any, overload
 import yaml
 
 from airbyte.constants import SECRETS_HYDRATION_PREFIX
-from airbyte.mcp._guards import require_trusted_execution
+from airbyte.mcp._guards import raise_if_untrusted_execution_context
 from airbyte.secrets.hydration import deep_update, detect_hardcoded_secrets
 from airbyte.secrets.util import get_secret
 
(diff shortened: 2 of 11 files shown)


airbyte/mcp/_arg_resolvers.py
```diff
@@ -137,6 +159,11 @@ def _raise_invalid_type(file_config: object) -> None:
         else:
             raise ValueError(f"Config must be a dict or JSON string, got: {type(config).__name__}")
 
+    if _contains_secret_reference(config_dict):
+        require_trusted_execution(
```
aaronsteers:
```suggestion
        raise_if_untrusted_execution_context(
```

Or 

```suggestion
        guard_untrusted_execution_context(
```

airbyte/exceptions.py
```diff
@@ -222,6 +222,34 @@ class PyAirbyteNoStreamsSelectedError(PyAirbyteInputError):
...
+# MCP Server Errors
+
+
+@dataclass
+class PyAirbyteMCPError(PyAirbyteError):
+    """An error occurred in the PyAirbyte MCP server."""
+
+
+@dataclass
+class PyAirbyteTrustedExecutionRequiredError(PyAirbyteMCPError):
```
aaronsteers: The PyAirbyte prefix is inappropriate.

```suggestion
class AirbyteMCPError(PyAirbyteError):
    """An error occurred in the PyAirbyte MCP server."""


@dataclass
class AirbyteTrustedExecutionRequiredError(PyAirbyteMCPError):
```

tests/unit_tests/test_mcp_trusted_execution.py
aaronsteers, line 1: Most of these tests don't use parametrized decorator. I suspect you could consolidate.

PR #1089, opened 2026-07-24
fix(mcp): make Cloud credential guidance execution-mode-aware

aaronsteers:
## Summary

Cloud MCP tools told users to "set the `AIRBYTE_CLOUD_*` environment variables" to authenticate. That guidance is wrong under the hosted (HTTP) MCP transport, where credentials arrive as request headers, not server-side env vars. This makes the credential/workspace guidance execution-mode-aware and moves it into self-rendering exception classes that both PyAirbyte and the Ops MCP can share.

**Mode detection is a dependency-light process-global flag** — no FastMCP in the credential/guidance path, since FastMCP is becoming an optional extra and core code must not import it there:

- The flag and its accessors live in `airbyte/constants.py`: `set_hosted_mcp_mode()` / `is_hosted_mcp_mode()`, default `False`. They're in `constants.py` (a stdlib-only leaf module) deliberately — `exceptions.py` already imports from `constants`, so reading the flag there adds no new import edge and avoids the module-import cycle that a dedicated `_util` submodule introduced.
- The hosted HTTP entrypoint `airbyte/mcp/http_main.py` calls `set_hosted_mcp_mode()` at startup. The stdio/local path never sets it, so an unset flag ⇒ not hosted. No token/header/request inspection is used to infer mode.

**Guidance lives in two new self-rendering exception classes in `airbyte/exceptions.py`** (call sites just `raise` — no inline message/guidance strings). Each computes mode-aware `guidance` in `__post_init__`, reading only `is_hosted_mcp_mode()` plus the header/env-var constants (no FastMCP, no MCP-layer import):

```python
class AirbyteNoCloudCredentialsError(PyAirbyteInputError):
    guidance: str | None = None
    _allow_bearer: bool = True   # client-creds-only callers (e.g. connector-QA) set False
    _env_vars: bool = True       # False => do not suggest env vars
    # hosted => Authorization / X-Airbyte-Cloud-Client-Id + -Secret headers
    # local  => bearer_token or client_id+client_secret (args or AIRBYTE_CLOUD_* env vars)

class AirbyteMissingWorkspaceContextError(PyAirbyteInputError):
    guidance: str | None = None
    # hosted => X-Airbyte-Workspace-Id header (or workspace_id param)
    # local  => AIRBYTE_CLOUD_WORKSPACE_ID env var (or workspace_id param)
```

The underscore-prefixed control fields are excluded from the rendered error text, and the explicit `guidance: str | None = None` field makes it unambiguous that mode-aware guidance is applied by default (the inherited `PyAirbyteInputError.guidance` is an unannotated class attr, not a dataclass field default). `_credentials.py` raises `AirbyteNoCloudCredentialsError(_env_vars=env_vars)`; `mcp/cloud.py` raises `AirbyteMissingWorkspaceContextError`. The old `get_mcp_credential_guidance()` helper in `_tool_utils.py` is removed.

**Static tool help / field descriptions** (baked at import, can't vary per request) are reworded to mention *both* paths and name the exact headers/env vars via constants: `CLOUD_AUTH_TIP_TEXT`, `WORKSPACE_ID_TIP_TEXT` (`airbyte/mcp/cloud.py`), and the "Cloud operations" line in `MCP_SERVER_INSTRUCTIONS` (`airbyte/mcp/server.py`).

**Import cycle / CodeQL:** `exceptions.py` used to type `AirbyteError.workspace` via a `TYPE_CHECKING` import of `airbyte.cloud.workspaces.CloudWorkspace`, which placed `exceptions` on the `_credentials` → `exceptions` → `cloud` → `_credentials` module cycle that CodeQL surfaces once this PR edits `_credentials.py`. That import is now replaced with a local `_WorkspaceWithUrl` `Protocol` — structural typing for the single `workspace_url` attribute the error needs, so any object exposing it (e.g. `CloudWorkspace`) satisfies it without an ancestor or import. `exceptions` therefore no longer imports `airbyte.cloud` at all, which removes the cycle edge; CodeQL is clean.

Out of scope (intentionally unchanged): `CloudClientConfig.from_env` / `CloudWorkspace.from_env` docstrings (env-var by definition), and all non-credential env-var text.

## Test plan

- `ruff format --check .` and `ruff check .` clean; `import airbyte` clean.
- `tests/unit_tests/test_exceptions.py` covers both classes in hosted + local modes (by toggling the shared flag), asserts `_env_vars=False` drops env-var guidance, and that control fields don't leak into rendered text; `test_cloud_credentials.py` updated for the new wording. The Ops MCP consumes the same shared exceptions and flag.

Link to Devin session: https://app.devin.ai/sessions/3ef501472b3d49ca831f588f36f95b29
Requested by: @aaronsteers


airbyte/cloud/_credentials.py
```diff
@@ -87,7 +87,10 @@ def from_auth(
             )
         if not resolved_bearer_token and not resolved_client_id:
             guidance = (
-                "Set Airbyte Cloud credentials in environment variables."
+                "Provide Airbyte Cloud credentials as explicit arguments "
+                "(client_id and client_secret, or bearer_token), or via the "
+                "`AIRBYTE_CLOUD_*` environment variables. When connecting to a "
+                "hosted MCP server, credentials are supplied via request auth headers."
                 if env_vars
```
aaronsteers: This isn't an improvement. Goes against the principle of providing _targeted_ instruction or just _not_ providing it.

airbyte/mcp/cloud.py
```diff
@@ -261,7 +279,8 @@ def _get_cloud_workspace(
     if not resolved_workspace_id:
         raise PyAirbyteInputError(
             message="Workspace ID is required but not provided.",
-            guidance="Set AIRBYTE_CLOUD_WORKSPACE_ID env var or pass workspace_id parameter.",
+            guidance=f"{get_mcp_credential_guidance(workspace_only=True)} "
+            "You can also pass the workspace_id parameter.",
         )
```
aaronsteers: This should just be a new exception class AirbyteMissingWorkspaceContextError

That exception class can ihnerit from another one of our custom classes, and can automatically render its message without the call site needing to write the message or import the logic.

airbyte/mcp/_tool_utils.py
```diff
@@ -182,6 +187,27 @@ def check_guid_created_in_session(guid: str) -> None:
...
+            f"the `{MCP_WORKSPACE_ID_HEADER}` header."
+        )
+
+    if workspace_only:
+        return f"Set the `{CLOUD_WORKSPACE_ID_ENV_VAR}` environment variable."
+    return (
+        f"Set the `{CLOUD_BEARER_TOKEN_ENV_VAR}` environment variable, or set both "
+        f"`{CLOUD_CLIENT_ID_ENV_VAR}` and `{CLOUD_CLIENT_SECRET_ENV_VAR}`."
+        f" Set the `{CLOUD_WORKSPACE_ID_ENV_VAR}` environment variable for the workspace."
+    )
```
aaronsteers: Maybe this can just live in the custom exceptions module. (Assuming no circular dependencies.)

airbyte/cloud/_credentials.py
```diff
@@ -86,14 +86,9 @@ def from_auth(
...
-            guidance = (
-                "Set Airbyte Cloud credentials in environment variables."
-                if env_vars
-                else "Provide either bearer_token or both client_id and client_secret."
-            )
             raise PyAirbyteInputError(
                 message="No Airbyte credentials found.",
-                guidance=guidance,
+                guidance="Provide either bearer_token or both client_id and client_secret.",
             )
```
aaronsteers: Ditto. This probably should just be a custom Exception class: AirbyteNoCloudCredentialsError which knows how to render itself.

airbyte/exceptions.py
```diff
@@ -44,7 +44,18 @@
 from textwrap import indent
 from typing import TYPE_CHECKING, Any
 
-from airbyte.constants import AIRBYTE_PRINT_FULL_ERROR_LOGS
+from airbyte._util.mcp_mode import is_hosted_mcp_mode
```
aaronsteers: If this is the primary source of the import cycle, these simple helpers could be moved into the constants module. It might be worth doing that anyway, and the constants module should just be simple logical constants without any other external imports. Alternatively, I think we also have a hidden meta sub-module of the telemetry module that's used for similar flags.

PR #1094, opened 2026-07-27
feat(mcp): let agents discover their own Cloud org and workspace IDs

aaronsteers:
## Summary

Agents connecting to Cloud MCP with only client credentials had no way to learn their own organization or workspace ID, so they had to stop and ask the user for a UUID. The API could already answer the question — `GET /organizations` and `GET /workspaces` both succeed with no ID supplied — but the MCP surface never exposed it: `list_cloud_workspaces` hard-required an org selector and its docstring said it "will NOT list workspaces across all organizations", and there was no organization-listing tool at all.

This PR exposes the existing capability rather than adding new API calls:

- New `CloudClient.list_organizations()` returning typed `CloudOrganization`s, backed by the existing `api_util.list_organizations_for_user()`. `get_organization()` now filters that list instead of duplicating the call — behavior is unchanged since `CloudOrganization` already scopes its own credentials to `organization_id` in its constructor.
- New read-only `list_cloud_organizations` MCP tool taking no inputs.
- `list_cloud_workspaces` now takes the org-less public-API path when neither `organization_id` nor `organization_name` is given, and keeps the Config API `list_by_organization_id` path when one is:

```python
client.list_workspaces(
    organization_id=(
        _resolve_organization_id(...) if organization_id or organization_name else None
    ),
    ...
)
```

Both discovery tools return a result model with a `message` field rather than raising on the two expected dead ends — zero results, and 401/403 from under-permissioned (e.g. workspace-scoped) credentials that can't read organizations. Anything else still raises.

Guidance in `MCP_SERVER_INSTRUCTIONS`, `CLOUD_AUTH_TIP_TEXT`, and `AirbyteMissingWorkspaceContextError` now tells agents to discover first and to ask the user to choose when discovery returns multiple candidates — never auto-select. No session/selection state is added here; this is read-only discovery only.

Verified live against Airbyte Cloud with sandbox client credentials and no org/workspace configured: `list_cloud_organizations({})` returned `Airbyte Team`, and `list_cloud_workspaces({})` returned `Devin's Sandbox`.

Link to Devin session: https://app.devin.ai/sessions/0f9b254d63ef4705a5dbdcc22bc40d7f
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Added authenticated discovery of Cloud organizations and workspaces.
  * Added structured discovery results with optional guidance messages.
  * Workspace discovery now searches across available organizations when none is selected.
  * Users are prompted to choose when multiple organizations or workspaces are available.

* **Bug Fixes**
  * Improved permission-denied handling with empty results and actionable guidance.
  * Enhanced workspace context guidance for hosted and non-hosted usage.

* **Tests**
  * Expanded coverage for organization, workspace, and permission-handling scenarios.


tests/unit_tests/test_cloud_credentials.py
aaronsteers, line 1: Please make your tests more dense. Fewer tests, greater degree of parametrization.

PR #1120, opened 2026-08-25
fix(mcp): validate Host and Origin headers on the HTTP transport

pablo-airbyte:
## Summary

The HTTP MCP entrypoint accepts any `Host` and `Origin` header, so a browser page can drive a locally or privately run `airbyte-mcp-http` server (CWE-346, reported externally by a third party researcher; requested by Pablo Vega). The pinned `fastmcp` 3.2.0 constructs `StreamableHTTPSessionManager` without `TransportSecuritySettings` and ships no host/origin guard, so there is no flag to flip; this adds the check as an ASGI middleware in PyAirbyte instead of bumping a public dependency.

`HostOriginGuardMiddleware` in `airbyte/mcp/_transport_security.py` wraps the whole app, outside the existing client-credentials wrapper:

```python
wrapper = lambda http_app: HostOriginGuardMiddleware(wrap_if_enabled(http_app), allowed_hosts)
```

- missing or non-allowlisted `Host` -> `421`
- present but non-allowlisted `Origin` -> `403`; absent `Origin` is allowed (non-browser clients)
- comparison is `fnmatch` on the hostname only, port-stripped, bracket-stripped, lowercased

There is deliberately **no** same-origin fallback (accepting `Origin == scheme://<request Host>`, as upstream fastmcp's newer guard does). Both headers are attacker-controlled in the rebinding scenario, so that fallback makes the Origin check bypassable whenever the Host allowlist is permissive.

The allowlist resolves from three sources, none of them request-derived: loopback defaults (`127.0.0.1`, `localhost`, `::1`), the hostname parsed from `MCP_SERVER_URL` (already injected with the public URL by the hosted deployment, so hosted traffic keeps working with no new config), and the new `AIRBYTE_MCP_ALLOWED_HOSTS` (comma-separated, patterns like `*.run.app` allowed) for ingress hostnames, in-cluster Service DNS names, or anything else the process cannot infer about itself.

`AIRBYTE_MCP_HTTP_HOST` now overrides the previously hard-coded bind address. The default stays `0.0.0.0` so hosted deployments are unaffected; the guard, not the bind address, is what closes LAN and browser access here.

**Before rolling this out to a deployment**, confirm every legitimate `Host` value is covered. Load-balancer or kubelet health probes that send an IP-literal `Host`, and direct `*.run.app` hits where ingress still permits them, will get a `421` unless they are added to `AIRBYTE_MCP_ALLOWED_HOSTS` or pointed at an unguarded path with an explicit `Host`.

## Test plan

`tests/unit_tests/test_mcp_transport_security.py` drives the middleware through `starlette.testclient.TestClient`: allowed loopback hosts pass; `Host: attacker.example:8080` gets `421`; `Origin: http://attacker.example:8080` with an allowed `Host` gets `403` (the regression test for the reported vector); missing `Origin` passes; missing `Host` gets `421`; `AIRBYTE_MCP_ALLOWED_HOSTS` entries and pattern entries pass; `resolve_allowed_hosts` includes the `MCP_SERVER_URL` hostname. The existing `http_main` delegation test now asserts the wrapper produces the guard.

Run: `pytest tests/unit_tests/test_mcp_transport_security.py tests/unit_tests/test_mcp_auth.py` (38 passed locally), plus `ruff format --check .`, `ruff check .`, and `mypy` on the changed modules. Full-repo `mypy` is blocked by a preexisting duplicate `setup` module error in the test fixture sources, unrelated to this branch.

Link to Devin session: https://app.devin.ai/sessions/d15a969d92424c7990d399577f916d89
Requested by: @pablo-airbyte

## Summary by CodeRabbit

* **New Features**
  * Added configurable HTTP server bind host settings.
  * Added allowed-host configuration through `AIRBYTE_MCP_ALLOWED_HOSTS`.
  * Added host and origin validation, including wildcard support.
  * Rejected invalid, missing, duplicate, or disallowed host headers and origins with appropriate responses.

* **Documentation**
  * Documented the HTTP server host and allowed-host environment variables.

* **Tests**
  * Added coverage for accepted and rejected hosts, origins, missing and duplicate headers, wildcard configuration, and server URL host detection.

### DIFF
--- a/airbyte/mcp/__init__.py
+++ b/airbyte/mcp/__init__.py
@@ -212,7 +212,9 @@ set; the interactive path activates once the OIDC client credentials are set.
 - `MCP_SERVER_URL` — public base URL of the server (also used for OIDC redirect
   callbacks); defaults to `http://localhost:8080`.
 - `AIRBYTE_MCP_ALLOWED_HOSTS` — comma-separated allowed hostnames or `fnmatch`
-  patterns for HTTP `Host` and `Origin` validation; entries may include ports.
+  patterns for HTTP `Host` and `Origin` validation. Ports are ignored: an entry
+  may carry one for readability, but matching is on hostname only, so
+  `example.com:8443` also allows `example.com` on any port.
 - `AIRBYTE_MCP_HTTP_HOST` — host interface to bind for the HTTP server (defaults
   to `0.0.0.0`).
 - `AIRBYTE_MCP_OIDC_CLIENT_ID`, `AIRBYTE_MCP_OIDC_CLIENT_SECRET` — enable
--- a/airbyte/mcp/__init__.py
+++ b/airbyte/mcp/__init__.py
@@ -211,6 +211,10 @@ set; the interactive path activates once the OIDC client credentials are set.
 
 - `MCP_SERVER_URL` — public base URL of the server (also used for OIDC redirect
   callbacks); defaults to `http://localhost:8080`.
+- `AIRBYTE_MCP_ALLOWED_HOSTS` — comma-separated allowed hostnames or `fnmatch`
+  patterns for HTTP `Host` and `Origin` validation; entries may include ports.
+- `AIRBYTE_MCP_HTTP_HOST` — host interface to bind for the HTTP server (defaults
+  to `0.0.0.0`).
 - `AIRBYTE_MCP_OIDC_CLIENT_ID`, `AIRBYTE_MCP_OIDC_CLIENT_SECRET` — enable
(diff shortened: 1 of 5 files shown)


aaronsteers (approved):
Approving to merge when ready, pending other bot feedback.

Also: I suggest after this does merge, we trigger a follow-up to move the wiring upstream into fastmcp-extensions, and then another PR to refactor to use the upstream. Devin, you can tag me with the HITL tools afterwards when that is ready and I'm happy to review. (Not a blocker though for this phase.)

PR #1122, opened 2026-08-25
feat(mcp): add Cloud connector check and sync cancellation tools

### LINKED ISSUE #1118: feat(mcp): add Cloud connector check and sync cancellation tools
## Summary

Two Cloud MCP tools are missing that agents consistently need, and in both cases the underlying capability already exists (or nearly exists) below the MCP layer:

1. **`check_cloud_source` / `check_cloud_destination`** — run a connection check against a *deployed* Cloud actor.
2. **`cancel_cloud_sync`** — cancel an in-flight sync job.

## 1. Connector check tools

`CloudConnector.check()` already exists in core and hits the Config API (`/sources/check_connection`, `/destinations/check_connection`) via `api_util.check_connector`, but no MCP tool exposes it:

- https://github.com/airbytehq/PyAirbyte/blob/f65f2272882ffc8fae7ca9b459976574d07d51d5/airbyte/cloud/connectors.py#L165-L200
- https://github.com/airbytehq/PyAirbyte/blob/f65f2272882ffc8fae7ca9b459976574d07d51d5/airbyte/_util/api_util.py#L1849-L1896

Today an agent can `deploy_source_to_cloud` and `update_cloud_source_config` with bad credentials and only discover the problem from a failed sync job — an expensive, slow, and noisy feedback loop. The local surface has `validate_connector_config`, but that is gated behind trusted execution and is structurally unavailable on the hosted HTTP transport, so hosted agents have no check path at all.

Suggested shape: read-only-ish tools taking `source_id` / `destination_id` (plus the standard `workspace_id`), returning a typed result with `succeeded: bool` and the failure `message`. Natural follow-on: an optional `check=True` on the deploy/update-config tools.

## 2. Sync cancellation

`cancel` appears nowhere in `airbyte/_util/api_util.py` or `airbyte/cloud/`, so this needs a core method (public API `DELETE /v1/jobs/{jobId}`) plus the MCP wrapper.

`run_cloud_sync` can start a sync that runs for hours; there is currently no way for the agent (or the user talking to it) to stop it. This is also a cost/safety gap, not only convenience: an agent that starts the wrong sync cannot undo it.

Suggested shape: `cancel_cloud_sync(connection_id, job_id=None, ...)` defaulting to the currently running job, returning the resulting job status. `list_cloud_sync_jobs` and `list_deployed_cloud_connections` already surface running-job info, so the agent has the job ID it needs.

## Context

Found while auditing the current Cloud MCP surface (36 cloud tools at `f65f227`); these were the two most clear-cut functional gaps.

aaronsteers:
## Summary

Closes #1118. Adds the two Cloud MCP capabilities requested by AJ Steers (@aaronsteers): checking a *deployed* Cloud actor, and cancelling an in-flight sync job.

**1. `check_cloud_source` / `check_cloud_destination`** — pure presentation wrappers over the existing `CloudConnector.check()`, called with `raise_on_error=False` so a failed check comes back as data rather than an exception:

```python
ConnectorCheckResult(
    connector_id=source_id,
    connector_type=source.connector_type,   # Literal["source", "destination"]
    succeeded=check_result.success,
    message=None if check_result.success else (check_result.error_message or check_result.internal_error),
)
```

Annotated `read_only=True, idempotent=True` — a check runs a connector-side check job but mutates no Airbyte config, so it stays available in read-only/hosted sessions (where `validate_connector_config` is unavailable behind trusted execution).

**2. `cancel_cloud_sync`** — needed new core support, added bottom-up:

- `api_util.cancel_job(job_id, ...)` over the public API's `DELETE /v1/jobs/{jobId}`, via the pinned SDK's `jobs.cancel_job(api.CancelJobRequest(job_id=...))`. Unlike `get_job_info`, a non-2xx here is frequently *not* a missing job — 409/422 means "exists but not cancellable" and 401/403 means auth — so only 404 raises `AirbyteMissingResourceError` and everything else (including a 2xx with an empty payload) raises `AirbyteError`.
- `CloudConnection.cancel_sync(job_id=None)` resolves the target job and then cancels it:

```python
if job_id is None:
    # latest *sync* job only: the API's default job-type filter also matches resets
    latest = get_previous_sync_logs(limit=1, job_type=JobTypeEnum.SYNC)
    -> PyAirbyteInputError if absent, or already in FINAL_STATUSES
else:
    job = api_util.get_job_info(job_id)
    -> PyAirbyteInputError if job.connection_id != self.connection_id
    -> PyAirbyteInputError if job status in FINAL_STATUSES
```

  The explicit-`job_id` ownership check matters because `cancel_job` is addressed purely by job ID: without it, a job ID from *another* connection in an accessible workspace would be cancelled even though the caller named this connection. Returns a `SyncResult` seeded with the cancel response, so the caller sees the post-cancel status without a second fetch.
- The MCP tool is `destructive=True` (not `read_only`, not `idempotent`) and returns the existing `SyncJobResult`.

**Safe mode decision: `cancel_cloud_sync` is deliberately *not* gated by `check_guid_created_in_session`.** Safe mode exists to stop an agent from destroying pre-existing resources; cancelling a job destroys nothing and is reversible — the sync can simply be re-run. Gating it would also break the primary use case, since the runaway sync an agent or user most wants stopped is typically on a connection the current session did not create. The omission is noted in a code comment so it doesn't read as an oversight.

## Test plan

Mock-based unit tests at the three seams — `api_util` (SDK request shape; 404 vs 409 error types), `CloudConnection.cancel_sync` (sync-only default selection, no-jobs, already-finished, foreign-connection job, successful cancel), and the MCP mapping (both check tools including `error_message` → `internal_error` fallback, plus `cancel_cloud_sync` → `SyncJobResult`).

Locally: `ruff format`, `ruff check .`, `mypy airbyte/` clean, and the repo's fast suite — `628 passed, 1 skipped, 85 deselected`. The full-repo `mypy .` run hits a duplicate-`run`-module error in `tests/integration_tests/fixtures/source-broken`, confirmed pre-existing on a clean `origin/main` worktree.

Neither new code path is exercised against live Cloud in CI: `check_*` needs a deployed actor and `cancel_cloud_sync` needs a running job.

## Summary by CodeRabbit

- **New Features**
  - Added the ability to cancel cloud sync jobs, including the latest sync when no job is specified.
  - Added connector checks for cloud sources and destinations.
  - Connector checks report success or provide an actionable failure message.
- **Bug Fixes**
  - Added validation for missing, completed, or unrelated sync jobs before cancellation.
  - Improved error handling when a requested job cannot be found or cancellation fails.

Link to Devin session: https://app.devin.ai/sessions/fe45a6bd8ca74a459e02c8f3b09a840e

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

### DIFF
--- a/airbyte/cloud/connections.py
+++ b/airbyte/cloud/connections.py
@@ -23,6 +23,7 @@ from airbyte.cloud._connection_state import (
     _normalize_state_to_protocol,
 )
 from airbyte.cloud.connectors import CloudDestination, CloudSource
+from airbyte.cloud.constants import FINAL_STATUSES
 from airbyte.cloud.models import (
     CloudConnectionInfo,
     CloudJobInfo,
@@ -301,9 +302,17 @@ class CloudConnection:  # noqa: PLR0904  # Too many public methods
         return sync_result
 
     def cancel_sync(self, job_id: int | None = None) -> SyncResult:
-        """Cancel a running sync job. Defaults to the connection's most recent job."""
+        """Cancel a running sync job.
+
+        Defaults to the connection's most recent sync job. Other job types must be
+        targeted with an explicit `job_id`.
+        """
         if job_id is None:
-            sync_result = self.get_sync_result()
+            sync_results = self.get_previous_sync_logs(
+                limit=1,
+                job_type=JobTypeEnum.SYNC,
+            )
+            sync_result = sync_results[0] if sync_results else None
             if sync_result is None:
                 raise PyAirbyteInputError(
                     message="No sync jobs found for this connection.",
@@ -317,6 +326,28 @@ class CloudConnection:  # noqa: PLR0904  # Too many public methods
                     ),
                 )
             job_id = sync_result.job_id
+        else:
+            job_info = api_util.get_job_info(
+                job_id=job_id,
+                api_root=self.workspace.api_root,
+                client_id=self.workspace.client_id,
+                client_secret=self.workspace.client_secret,
+                bearer_token=self.workspace.bearer_token,
+            )
+            if job_info.connection_id != self.connection_id:
+                raise PyAirbyteInputError(
+                    message=(
+                        f"Job {job_id} belongs to connection '{job_info.connection_id}', "
+                        f"not '{self.connection_id}'."
+                    ),
+                )
+            job_status = CloudJobInfo.from_api_response(job_info).status
+            if job_status in FINAL_STATUSES:
+                raise PyAirbyteInputError(
+                    message=(
+                        f"Job {job_id} is already finished with status " f"'{job_status.value}'."
+                    ),
+                )
 
         job_response = api_util.cancel_job(
             job_id=job_id,
--- a/airbyte/cloud/connections.py
+++ b/airbyte/cloud/connections.py
@@ -301,56 +301,64 @@ class CloudConnection:  # noqa: PLR0904  # Too many public methods
 
         return sync_result
 
+    def _get_latest_cancellable_sync_job_id(self) -> int:
+        """Get the latest cancellable sync job ID."""
+        sync_results = self.get_previous_sync_logs(
+            limit=1,
+            job_type=JobTypeEnum.SYNC,
+        )
+        sync_result = sync_results[0] if sync_results else None
+        if sync_result is None:
+            raise PyAirbyteInputError(
+                message="No sync jobs found for this connection.",
+            )
+        if sync_result.is_job_complete():
+            raise PyAirbyteInputError(
+                message=(
+                    f"The latest sync job is already finished with status "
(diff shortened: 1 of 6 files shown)


airbyte/cloud/connections.py
```diff
@@ -300,6 +301,68 @@ def run_sync(
...
+            if sync_result.is_job_complete():
+                raise PyAirbyteInputError(
+                    message=(
+                        f"The latest sync job is already finished with status "
+                        f"'{sync_result.get_job_status().value}'. "
+                        "Pass an explicit job_id to target a different job."
+                    ),
+                )
+            job_id = sync_result.job_id
+        else:
```
aaronsteers: Use the call/return early paradigm so we don't need "else".

PR #1125, opened 2026-08-25
fix(mcp): scale organization discovery and resolve default org

aaronsteers:
## Summary

Makes organization and workspace discovery resolve a context instead of dumping everything the credentials can see, and moves organization search onto an endpoint that can actually filter and paginate.

Two problems remained after the resolution work earlier in this branch, both only visible with instance-admin credentials (167k visible organizations):

1. `list_cloud_organizations` was unbounded. The public `GET /organizations` endpoint the SDK wraps ignores `limit`, `offset` and `pageSize` — every variant returns the same 23.6 MB / 167,651-row body. Over the hosted SSE transport that payload never makes it back to the agent: the stream pings and closes at 40-70s. The Config API's `POST /organizations/list_by_user_id` does support `nameContains` and `pagination {pageSize, rowOffset}` server-side, so filtered and limited lookups now go there:

   ```python
   client.list_organizations(name_contains="Airbyte Team")  # 2 rows, ~1.2s (was ~29s, all 167k rows scanned client-side)
   client.list_organizations(limit=5)                        # 5 rows, ~0.6s
   client.list_organizations()                               # unchanged: one public-API request, all rows
   ```

   The unfiltered library path deliberately stays on the public API — one 2.3s request beats 1,600 paginated ones. The MCP tool defaults `limit` to 100 and always states in `message` that the result is the first N of possibly more, so a truncated list is never presented as complete.

2. `get_organization()` with no arguments errored, so an agent asking "which organization am I in?" had to go through `list_cloud_workspaces` and then describe by ID. It now resolves the same way workspace listing does — configured `organization_id`, then the parent organization of `default_workspace_id`, then the authenticated user's sole membership:

   ```python
   client.get_organization()  # -> Airbyte Team (664c690e-…), ~0.6s
   ```

   With several memberships it raises `PyAirbyteInputError` whose **message text** enumerates up to ten candidates (`<id> (<name>)`, plus `showing 10 of N`), not just its `context` — `describe_cloud_organization` returns a single object and the error string is the only channel an agent sees. Lookups by ID are now a single `organizations/get_organization_info` call instead of a full-listing scan; name lookups narrow server-side first. All of these fall back to the public listing on `AirbyteError` so self-managed deployments keep working.

Also in this change, from agent-behavior testing of the preview deployment:

- `CLOUD_AUTH_TIP_TEXT` and `AirbyteMissingWorkspaceContextError.guidance` no longer tell agents to call `list_cloud_organizations` first — that text was the direct cause of every no-arg call that hung. They now point at `list_cloud_workspaces` (which resolves the organization itself) and mention the organization tool only for `name_contains` searches.
- `list_cloud_workspaces` populates `organization_name` on each row and, when it resolved the organization implicitly, says so in `message` ("Resolved organization Airbyte Team (664c690e-…)"), saving the agent a follow-up call. The extra lookup is one Config API call and failures are swallowed rather than failing the listing.

No new caps beyond the two stated above (MCP default `limit=100`, always announced; ten-candidate ambiguity list).

## Test plan

- `pytest tests/unit_tests/test_cloud_api_util.py tests/unit_tests/test_cloud_credentials.py tests/unit_tests/test_exceptions.py` — 101 passed. New cases cover the Config API pagination/filter payload, the Config-vs-public routing in `list_organizations` and its fallback, all three no-argument `get_organization` resolution paths, the ten-candidate ambiguity message and context, the single-call ID lookup, and `describe_cloud_organization` with no arguments.
- `ruff format .`, `ruff check .`, `pyrefly check` — clean.
- Live against Airbyte Cloud with instance-admin credentials: the four timings quoted above, plus bare `list_workspaces()` returning the Airbyte Team workspaces with `organization_id` populated.
- Not covered: the multi-membership ambiguity path is mocks-only — both credentials available here are single-organization.

Requested by AJ Steers.

Link to Devin session: https://app.devin.ai/sessions/290cd017df694ae4bdba43f2247dd27a
Open in Devin Desktop: https://app.devin.ai/desktop/session/290cd017df694ae4bdba43f2247dd27a?variant=devin
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Improved organization and workspace discovery using IDs, names, workspace context, configured defaults, or memberships.
  * Added organization name filtering and result limits.
  * Workspace listings now provide clearer organization details and available organization options.
  * Added automatic organization resolution for cloud workspace operations.

* **Bug Fixes**
  * Improved handling of ambiguous memberships and unexpected API responses.
  * Added fallback behavior when organization membership data is unavailable.

* **Documentation**
  * Updated missing-workspace guidance to recommend listing workspaces first.

### DIFF
--- a/airbyte/_util/api_util.py
+++ b/airbyte/_util/api_util.py
@@ -2483,7 +2483,17 @@ def list_organizations_for_user_id(
             bearer_token=bearer_token,
         )
 
+        if not isinstance(json_result, dict):
+            raise AirbyteError(
+                message="The organizations API returned an unexpected response.",
+                context={"response": json_result},
+            )
         organizations = json_result.get("organizations", [])
+        if not isinstance(organizations, list):
+            raise AirbyteError(
+                message="The organizations API returned an unexpected response.",
+                context={"response": json_result},
+            )
 
         if not organizations:
             break
@@ -2885,7 +2895,7 @@ def get_user_by_auth_id(
     bearer_token: SecretString | None,
 ) -> dict[str, Any]:
     """Get an Airbyte user by the authentication provider user ID."""
-    return _make_config_api_request(
+    result = _make_config_api_request(
         path="/users/get_by_auth_id",
         json={
             "authUserId": auth_user_id,
@@ -2896,6 +2906,13 @@ def get_user_by_auth_id(
         client_id=client_id,
         client_secret=client_secret,
         bearer_token=bearer_token,
+    )
+    if isinstance(result, dict):
+        return result
+
+    raise AirbyteError(
+        message="The user API returned an unexpected response.",
+        context={"response": result},
     )
 
 
--- a/airbyte/_util/api_util.py
+++ b/airbyte/_util/api_util.py
@@ -13,6 +13,7 @@ directly. This will ensure a single source of truth when mapping between the `ai
 
 from __future__ import annotations
 
+import base64
 import json
 from http import HTTPStatus
 from typing import TYPE_CHECKING, Any, Literal
@@ -47,6 +48,7 @@ if TYPE_CHECKING:
 JOB_WAIT_INTERVAL_SECS = 2.0
 JOB_WAIT_TIMEOUT_SECS_DEFAULT = 60 * 60  # 1 hour
 PAGE_SIZE = 100
+JWT_PART_COUNT = 3
 
 # Job ordering constants for list_jobs API
 JOB_ORDER_BY_CREATED_AT_DESC = "createdAt|DESC"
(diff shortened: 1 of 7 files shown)


airbyte/_util/api_util.py
```diff
@@ -554,11 +554,27 @@ def list_workspaces(
     name: str | None = None,
     name_filter: Callable[[str], bool] | None = None,
     limit: int | None = None,
+    max_pages: int | None = None,
```
aaronsteers: Remove max_pages. We need to apply this constrains ourselves with limit. Pages is meaningless to the caller and redundant with limit.

airbyte/mcp/cloud.py
```diff
@@ -1531,30 +1528,67 @@ def list_cloud_workspaces(
             default=None,
         ),
     ],
+    workspace_id: Annotated[
+        str | None,
+        Field(
+            description="Optional workspace ID used to resolve its organization.",
+            default=None,
+        ),
+    ],
```
aaronsteers: This arg doesn't belong here.

aaronsteers: If workspace ID is already set via env var or auth header, we can get it's parent. But don't expect caller asking for workspaces list to give you a workspace.

airbyte/mcp/cloud.py
```diff
@@ -1531,30 +1528,67 @@ def list_cloud_workspaces(
...
+    all_organizations: Annotated[
+        bool,
+        Field(
+            description=(
+                "Whether to explicitly list across all organizations instead of resolving "
+                "an organization context."
+            ),
+            default=False,
+        ),
+    ],
```
aaronsteers: I'd delete this as an input arg. If the caller gives an organization ID, we can filter by it. If not, and if no org or workspace ID context exists elsewhere, we can presume to search all orgs.

airbyte/mcp/cloud.py
```diff
@@ -224,6 +226,15 @@ class CloudOrganizationResult(BaseModel):
     Defaults to False unless we have affirmative evidence of a locked state."""
 
 
+class CloudOrganizationCandidate(BaseModel):
```
aaronsteers: I don't think this needs a "candidate" class. We probably already have a CloudOrganizationResult already in use which can and should carry these same two properties.

airbyte/mcp/cloud.py
```diff
@@ -270,6 +281,9 @@ class CloudWorkspaceListResult(BaseModel):
     message: str | None = None
     """Additional guidance when discovery returns no results."""
 
+    candidate_organizations: list[CloudOrganizationCandidate] | None = None
+    """Organizations to choose from when the credentials match multiple organizations."""
```
aaronsteers: Call this "available_organizations"

airbyte/mcp/cloud.py
```diff
@@ -1534,28 +1531,45 @@ def list_cloud_workspaces(
...
+            candidate_name = candidate.get("organization_name")
+            candidate_organizations.append(
+                CloudOrganizationCandidate(
+                    id=candidate_id,
+                    name=candidate_name if isinstance(candidate_name, str) else None,
+                )
+            )
+        message = error.get_message()
+        if candidate_organizations:
+            message += " Retry with one of the candidate organization IDs."
```
aaronsteers:
```suggestion
        if available_organizations:
            message += " Retry with an explicit organization ID from the provided list of the available organizations."
```

airbyte/cloud/client.py
```diff
@@ -80,6 +82,11 @@ def organization_id(self) -> str | None:
         """Default organization ID for organization-scoped operations."""
         return self._credentials.organization_id
 
+    @property
+    def workspace_id(self) -> str | None:
```
aaronsteers: If we're this as a new property here, and if it is not a hard constraint on scope, then we probably should name it "default_workspace_id" - assuming it is not a hard constraint.

tests/unit_tests/test_cloud_credentials.py
aaronsteers, line 1: It doesn't look like your tests are properly condensed. You should write fewer, denser unit tests with higher degrees of parameterization decorators.

airbyte/cloud/client.py
aaronsteers, line 1: We should have some documentation on what the resolution protocol is for organizations and workspaces. Probably best to describe this under a couple of conditional headers, such as:
- If workspace ID is or isn't provided
- If organization ID is or isn't provided


Under those categories, deduplicating them as appropriate, we'll have two or three ordered bulleted lists describing the resolution process for finding organizations and/or workspaces.

Unless there's a better place to put this, I lean towards putting it here as a module-level docstring, which will register and render in our auto-generated pdoc documentation and possibly also inline in a resolver. I'll leave that as optional. It might be better to just have a file-level docstring and then link that docstring with backtick-enveloped references to the classes or public methods that leverage it.

aaronsteers:
Devin, please deploy this PR to cloud-mcp-preview so we can test. Ping me in slack if you need help.

tests/unit_tests/test_cloud_api_util.py
aaronsteers, line 1: Make sure all of your tests are properly condensed. This means you use as few test functions as possible, with the highest degree of parameterization and decoration feasible for readability. Then your reader has the minimal amount of code to read, and they can just read through your parameterization sets to see if the behavior is working as desired.

PR #1127, opened 2026-08-27
feat(cloud): add Airbyte Agents support in PyAirbyte and `cloud-mcp`: `execute` API and MCP tools, `AgentOrganization`, and `AgentWorkspace`

aaronsteers:
## Summary

Requested by AJ Steers.

Adds Airbyte Agents connector execution to PyAirbyte's public API and to the Cloud MCP server. Cloud application credentials authenticate against the Agents API, so nothing new is needed from users: the same `AIRBYTE_CLOUD_*` environment variables and the same MCP config args work here.

Agents lives in its own package rather than on the Cloud classes, because the Agents API is a distinct host with its own auth scoping and resolution rules:

```python
from airbyte.agents import AgentWorkspace

workspace = AgentWorkspace.from_env()
connector = workspace.get_connector(name="GitHub")

print(connector.describe().context_store_entities)  # ["issues", ...]

page = connector.execute("issues", "list", {"state": "open"}, limit=50)
for issue in page.entities:
    ...
if page.has_next_page:
    connector.execute("issues", "list", {"state": "open"}, cursor=page.end_cursor)
```

Design points worth knowing before reading the diff:

1. Connector-specific arguments go in one `api_args` dict; every argument PyAirbyte itself owns (`select_fields`, `exclude_fields`, `limit`, `cursor`, `skip_truncation`, `intent`) stays explicit and keyword-only. `limit`/`cursor` are merged into the wire `params`, and a duplicate passed through `api_args` is rejected rather than silently overridden.
2. `entity` and `action` are free strings. `inspect` publishes only Context Store-supported entities, not an exhaustive executable entity/action matrix, so validating against it would reject legitimate calls. `download` is the one action rejected before transport — it returns a binary stream PyAirbyte cannot yet consume, and without the guard it surfaces as a confusing non-JSON parse error.
3. The Agents API root is a private module constant, not configurable. It is a single hosted service, so there is nothing for callers to override.
4. `AgentExecuteResult.entities` validates every item is a dict rather than filtering silently, and raises with guidance to use `.result` when the action's payload is not a list of entities.
5. Conversions exist both directions. `CloudWorkspace.as_agent_workspace()` verifies Agents reachability by default (`check=True`) — Cloud credentials authenticating does not imply the organization has an Agents subscription, so the failure is worth surfacing at conversion time rather than on first execute.

### MCP tools

`airbyte/mcp/agents.py` is a presentation layer only; all logic lives in `airbyte/agents/`. Five tools:

```text
list_agent_workspaces(organization_id=None)      read-only
list_agent_connectors(workspace_id=None)         read-only
describe_agent_connector(connector_id)           read-only
execute_agent_connector_ro(...)                  read-only, idempotent
execute_agent_connector(..., read_only=None)     write-capable
```

The two execute tools are thin wrappers over one `_execute()` helper. The split exists because readonly-mode filtering is per-tool (`readOnlyHint` is a static annotation read before any call), so a single mixed read/write tool would vanish entirely in readonly mode. `execute_agent_connector_ro` restricts `action` to `Literal["list", "get", "search", "api_search"]` and survives readonly mode; `execute_agent_connector` carries the full action literal and additionally accepts `read_only: bool | None` as a caller-side guard, which rejects write actions in the shared helper before any request is sent.

`workspace_id` appears only on `list_agent_connectors`, and drops out of the schema when `AIRBYTE_CLOUD_WORKSPACE_ID` is statically configured — same `_add_defaults_for_exclude_args` mechanism `cloud.py` uses. It is deliberately absent from describe and execute: the Agents API addresses those by connector ID alone, so a workspace argument there would be a decorative parameter an agent has to reason about.

### Known limitation

MCP config has no organization argument, so on multi-organization credentials `describe`/`execute` cannot send `X-Organization-Id`. The Agents API's own error is surfaced with guidance naming `AIRBYTE_CLOUD_ORGANIZATION_ID`. Adding an MCP org config arg is a Cloud-wide change and is left out of this PR.

## Test plan

- `uv run pytest tests/unit_tests/test_agents.py tests/unit_tests/test_mcp_agents.py` — 46 tests covering URL/header construction, organization routing, request bodies, pagination merge and conflict rejection, field selection, `download` rejection, strict entity validation, describe caching, Cloud/Agent conversions, HTTP error propagation, plus the MCP layer: result shaping, `api_args` JSON-string coercion, CSV field lists, the `read_only` guard, and the registered `readOnlyHint` values asserted against the live server tool list.
- `uv run poe test-fast`, `uv run ruff check .`, `uv run ruff format --check .`, `uv run pyrefly check`.
- `uv run poe mcp-docs-md` regenerates `docs/mcp-generated/agents.md`.
- Not covered by automated tests: live execution against the Agents API. That was exercised manually during development against a real GitHub connector (list/get with pagination), but no cassettes or credential-gated tests are included here.

Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Added Airbyte Agents support for discovering organizations, workspaces, and connectors.
  * Added connector inspection and action execution with pagination, field selection, and execution metadata.
  * Added MCP tools for listing, describing, and executing connector actions, including read-only controls.
  * Added conversion between Agents and Cloud organization and workspace representations.

* **Bug Fixes**
  * Improved validation and error handling for invalid, malformed, or non-JSON API responses.
  * Added clearer handling for ambiguous or missing workspace and connector matches.

* **Documentation**
  * Documented Agents usage examples and noted that the interface is experimental.

> [!NOTE]
> **Auto-merge may have been disabled. Please check the PR status to confirm.**


airbyte/constants.py
```diff
@@ -258,6 +258,19 @@ def _str_to_bool(value: str) -> bool:
 - https://github.com/airbytehq/airbyte-platform-internal/blob/master/oss/airbyte-api/server-api/src/main/openapi/config.yaml
 """
 
+AGENTS_API_ROOT_ENV_VAR: str = "AIRBYTE_AGENTS_API_URL"
+"""The environment variable name for the Airbyte Agents API URL."""
```
aaronsteers: Delete.

airbyte/constants.py
```diff
@@ -258,6 +258,19 @@ def _str_to_bool(value: str) -> bool:
...
+
+AGENTS_API_ROOT: str = "https://api.airbyte.ai/api/v1"
+"""The Airbyte Agents API root URL.
```
aaronsteers: Non-public, not intended for overriding. Prefix with "_"

aaronsteers: Or just keep without _ prefix - but be clear users shouldn't expect or want to override.

airbyte/agents/__init__.py
aaronsteers, line 1: Add a warning here in this doc string that the agent interfaces are experimental and subject to change.

airbyte/agents/_lookup.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) 2026 Airbyte, Inc., all rights reserved.
+"""Shared argument handling for Agents lookups by ID or name."""
+
+from __future__ import annotations
+
+from airbyte.exceptions import PyAirbyteInputError
+
+
+def resolve_id_or_name(
```
aaronsteers: I think this is connector id/name specific.

Should probably be moved back into `connectors.py` and renamed to `_resolve_connector_id_or_name()`

airbyte/agents/connectors.py
```diff
@@ -0,0 +1,302 @@
...
+    ```
+    """
+
+    def __init__(
+        self,
+        connector_id: str,
+        *,
+        credentials: _AirbyteCredentials,
+        name: str | None = None,
+    ) -> None:
```
aaronsteers, line 134: What about this approach?

```suggestion
    def __init__(
        self,
        _id_or_name: str | None = None,
        /,
        *,
        id: str | None = None,
        name: str | None = None,
        credentials: _AirbyteCredentials,
    ) -> None:
```

aaronsteers: Agreed. Take this approach for the .get_connector() method and similar resolver methods if applicable.

airbyte/agents/organizations.py
```diff
@@ -0,0 +1,169 @@
...
+    def _as_workspace(self, info: AgentWorkspaceInfo) -> AgentWorkspace:
+        """Build an `AgentWorkspace` from workspace info, reusing these credentials."""
+        return AgentWorkspace(
+            workspace_id=info.id,
+            organization_id=info.organization_id or self.organization_id,
+            name=info.name,
+            client_id=self._credentials.client_id,
+            client_secret=self._credentials.client_secret,
+            bearer_token=self._credentials.bearer_token,
+        )
```
aaronsteers: I can't rationalize this method as it is currently named. Care to defend? Or coalesce into `.get_workspace()`?

airbyte/agents/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+
+## Usage Examples
+
+Execute an action against a connector:
+
+```python
+from airbyte import agents
+
+workspace = agents.AgentWorkspace.from_env()
+connector = workspace.get_connector("GitHub")  # by ID or name
```
aaronsteers:
```suggestion
connector = workspace.get_connector("GitHub")  # by ID or name (case insensitive)
```

airbyte/agents/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+
+result = connector.list_entities(
+    "issues",
+    api_args={"repository": "airbytehq/PyAirbyte"},
```
aaronsteers, line 38:
```suggestion
    api_args={"repository": "airbytehq/PyAirbyte"},  # Passthrough API args
```

airbyte/agents/__init__.py
```diff
@@ -0,0 +1,99 @@
...
+    limit=50,
+)
+for entity in result.entities:
+    print(entity["title"])
+```
+
+Page through results using the cursor the connector reports:
+
+```python
+cursor: str | None = None
```
aaronsteers: `cursor` value ineffectual? Remove if so?

airbyte/agents/connectors.py
```diff
@@ -0,0 +1,335 @@
...
+        """Initialize an `AgentConnector`. Prefer `AgentWorkspace.get_connector()`."""
+        self.connector_id = connector_id
+        """The connector ID."""
+
+        self._credentials = credentials
+        self._name = name
+        self._details: AgentConnectorDetails | None = None
+
+    @classmethod
+    def from_auth(
```
aaronsteers: This `from_auth` class method factor function is awkwardly named and implemented? Remove or make private?

tests/unit_tests/test_agents.py
aaronsteers, line 1: It doesn't look like you've properly collapsed your tests. Make sure you have optimized test density by reducing the number of unit tests and increasing the degree of parameterization using the `parameterized` decorator.

airbyte/agents/__init__.py
aaronsteers: Shouldn't we have a (probably default) version of this where the `list_entities()` implementation tracks its own cursor, so caller can just do this?:

```
for issue in connector.list_entities(
    "issues",
    api_args={"repository": "airbytehq/PyAirbyte"},
):
    print(f"Found: {issue}")
```

aaronsteers: I don't mind a separate method for `iter_entities()` - so you can create that as suggested. We should consider this aliased or just directly named as a `.entities` property - since the intuitive call against an `.entities` property would (I think) be a lazy generator. My only concern with this approach (also valid for other approaches) is the cleanup process if caller abandons after only iterating `n` items and not the full set.

airbyte/agents/connectors.py
```diff
@@ -0,0 +1,383 @@
...
+    ) -> AgentExecuteResult:
+        """Run the `list` action, which returns a page of entities for `entity`."""
+        return self.execute(entity, "list", api_args, **kwargs)
+
+    def iter_entities(
+        self,
+        entity: str,
+        api_args: dict[str, Any] | None = None,
+        *,
+        max_entities: int | None = None,
```
aaronsteers: We use `limit` elsewhere. Why not name as `limit` here?

airbyte/agents/connectors.py
```diff
@@ -0,0 +1,383 @@
...
+                connector_id=self.connector_id,
+                request_body=request_body,
+                credentials=self._credentials,
+                organization_id=self._credentials.organization_id,
+            )
+        )
+
+    def list_entities(
+        self,
+        entity: str,
```
aaronsteers: I'm not comfortable with `entity` anywhere as an input arg. What you want is `entity_type: str` (e.g. `issues`) and the name implies you want the name of a specific entity (e.g. `issue #1234`).

airbyte/agents/connectors.py
aaronsteers: Don't use `limit` anywhere in our public signatures for page size. If you want page size, ask for `page_size`.

airbyte/mcp/agents.py
```diff
@@ -0,0 +1,542 @@
...
+    open_world=True,
+    extra_help_text=AGENTS_AUTH_TIP_TEXT,
+)
+def describe_agent_connector(
+    ctx: Context,
+    connector_id: Annotated[
+        str,
+        Field(description="The ID of the Airbyte Agents connector."),
+    ],
+) -> AgentConnectorDetailsResult:
```
aaronsteers, line 339: I don't know why you aren't taking workspace ID as an input arg. For correctness, safety, security, and a bunch of other reasons, it's simpler if we require this for every mcp tool call that touches a connector.

aaronsteers: I think you'll find elsewhere in this repo precedent where it's not strictly needed, but it's used for validation. That's probably the pattern we should have here.

airbyte/agents/__init__.py
```diff
@@ -0,0 +1,110 @@
...
+```
+
+Pass `result.end_cursor` back as `cursor` to page through manually:
+
+```python
+cursor = None
+while True:
+    result = connector.list_entities(
+        "issues", api_args={"repository": "airbytehq/PyAirbyte"}, cursor=cursor
+    )
```
aaronsteers, line 55:
```suggestion
    result = connector.list_entities(
        "issues",
        api_args={"repository": "airbytehq/PyAirbyte"},
        cursor=cursor,
    )
```

airbyte/agents/connectors.py
```diff
@@ -0,0 +1,396 @@
...
+        return AgentExecuteResult.model_validate(
+            _api_util.execute_agent_connector_action(
+                connector_id=self.connector_id,
+                request_body=request_body,
+                credentials=self._credentials,
+                organization_id=self._credentials.organization_id,
+            )
+        )
+
+    def list_entities(
```
aaronsteers, line 231: Are we able to list a `.entity_types` property as its own `list[str]`?

airbyte/agents/__init__.py
aaronsteers: Yes: leave iter_entities() as-is

And tell me what happens if the caller abaondons the iterator. Do we cleanup gracefully or at least eventually do GC? I guess this isn't huge stakes, since we aren't leaving a DB connection or cursor open. Just an API call that won't get called back, right?

airbyte/agents/connectors.py
aaronsteers: I think we ignore until we solve upstream handler for full set.

airbyte/agents/connectors.py
```diff
@@ -0,0 +1,396 @@
...
+        *,
+        organization_id: str | None = None,
+        client_id: str | SecretString | None = None,
+        client_secret: str | SecretString | None = None,
+        bearer_token: str | SecretString | None = None,
+    ) -> AgentConnector:
+        """Create an `AgentConnector` from credentials, without a workspace lookup.
+
+        Internal: callers should use `AgentWorkspace.get_connector()`. This exists for the
+        MCP layer, which is handed a connector ID and credentials with no workspace context.
```
aaronsteers: Is this still needed now that we expect callers to pass workspace ID?

aaronsteers:
Devin, it looks like pdoc docs are not rendering for the new modules. Please check that `__all__` is correct in their parent modules and that the MCP module uses the correct convention for MCP docs rendering.

aaronsteers:
Screenshot and attach docs previews in a new comment as evidence.

airbyte/agents/organizations.py
```diff
@@ -0,0 +1,238 @@
...
+        matches = [
+            workspace
+            for workspace in workspaces
+            if workspace.name and workspace.name.lower() == name_lower
+        ]
+        if not matches:
+            raise AirbyteError(
+                message="No workspace found with the given ID or name.",
+                guidance="Use `list_workspaces()` to see the available workspaces.",
+                context={"name": name},
```
calebevans-ab:
```suggestion
                context={"name": lookup.name},
```

Per Claude:

> `get_workspace` error messages use the keyword parameter name (which is None when the user passes a positional argument) instead of lookup.name (which always holds the resolved value).

airbyte/agents/organizations.py
```diff
@@ -0,0 +1,238 @@
...
+            )
+        if len(matches) > 1:
+            raise AirbyteError(
+                message="Multiple workspaces matched the given name.",
+                guidance="Pass `workspace_id` instead of `name`.",
+                context={"name": name, "match_count": len(matches)},
```
calebevans-ab:
```suggestion
                context={"name": lookup.name, "match_count": len(matches)},
```
Ditto ^

calebevans-ab (approved):
Small cosmetic thing Claude found, but lgtm!

PR #1124, opened 2026-08-25
feat(mcp): emit MCP tool-call telemetry via `fastmcp-extensions`

aaronsteers:
## Summary

Requested by AJ (@aaronsteers): make the Airbyte MCP server emit one `mcp_tool_call` Segment event per tool invocation, on every transport, using the shared `fastmcp-extensions` telemetry — no telemetry implementation in this repo.

Until now, MCP tool calls emitted no telemetry at all. The only MCP-attributable events in Segment were `sync`/`install` events carrying the `flags_mcp` process flag, which say "this process was serving MCP", not "this tool was called" — and sync tools are disabled in most MCP deployments, so the useful actions (log checks, connector discovery, config generation) were invisible.

This branch now consumes the auto-registration API added in [airbytehq/fastmcp-extensions#110](https://github.com/airbytehq/fastmcp-extensions/pull/110): `mcp_server()` registers `ToolCallTelemetryMiddleware` itself, so there is no `add_middleware` call here at all, and stdio and hosted HTTP are both instrumented without either entrypoint opting in.

```python
# airbyte/mcp/server.py
app = mcp_server(
    name="airbyte-mcp",
    package_name="airbyte",
    ...,
    telemetry=TelemetryConfig(
        package_name="airbyte",
        segment_write_key=segment_write_key,          # None => structured-log sink only
        segment_user_id=SEGMENT_USER_ID,              # fixed "airbyte-mcp"
        extra_properties=lambda: {"is_hosted_mcp": is_hosted_mcp_mode()},
    ),
)
```

Three choices worth a reviewer's attention:

1. **`extra_properties` is a callable, not a dict.** `set_hosted_mcp_mode()` runs in `http_main.main()`, *after* this module is imported, so a value captured at import time would report `is_hosted_mcp=False` for the hosted Cloud MCP server. The callable is resolved per tool call, which is what makes hosted-vs-local separable downstream (previously they were indistinguishable in the payload).
2. **Default write key is PyAirbyte's own** (`PYAIRBYTE_APP_TRACKING_KEY`), so events land in the same Segment source as `install`/`sync` and are warehouse-modelable next to them; `AIRBYTE_MCP_SEGMENT_WRITE_KEY` overrides it per deployment. `DO_NOT_TRACK` and `AIRBYTE_OFFLINE_MODE` suppress the Segment sink; the structured-log sink stays on.
[...]

PR #1131, opened 2026-08-27
feat(mcp): add anonymized MCP attribution to tool-call telemetry

aaronsteers:
## Summary

Follow-up to #1124, requested by AJ. The `mcp_tool_call` event currently carries only `name`/`success`/`duration_ms`/`error_type`/`package_version`/`is_hosted_mcp`, and `userId` is the constant label `airbyte-mcp`. That supports counting tool calls but gives no denominator: no sessions, no distinct callers, no way to tell prod from preview. This adds attribution dimensions without emitting any raw identifier.

New properties on each event, all optional and independently guarded:

| property | source | notes |
| --- | --- | --- |
| `session_id_hash` | `Context.session_id` | hashed, not raw — session IDs are not guaranteed anonymous and can be used to reach stateful artifacts |
| `caller_hash` | first hop of `X-Forwarded-For` | on Cloud Run `request.client.host` is the load balancer, not the caller |
| `auth_subject_hash` | `get_access_token().claims["sub"]` | more stable than IP; survives network changes |
| `mcp_endpoint` | request host + path | plaintext **only** for Airbyte-owned hosts |
| `mcp_endpoint_hash` | request host | always set when a host is known, so third-party deployments are countable without leaking their hostname |
| `mcp_client_name` / `mcp_client_version` | `initialize` handshake | which client is calling (Claude Desktop, Cursor, …) |

The hash construction is the part worth reviewing:

```python
hmac.new(salt.encode(), f"{scope_label}|{endpoint}|{value}".encode(), hashlib.sha256).hexdigest()[:16]
```

It is keyed, not a bare digest, and that is not incidental. IPv4 is only 2^32 addresses, so `sha256(ip)` — or a hash seeded solely with public inputs like the endpoint — is invertible by exhaustive enumeration in minutes, and 64-bit truncation does not prevent that (one candidate in 4.3 billion matches, so a hit is a confirmation). The secret key is the only thing making the value one-way. The `scope_label` and endpoint are in the message for separation rather than secrecy: they stop the same caller from correlating across prod and preview, and stop the IP surrogate from colliding with the session surrogate.

Salt resolution is `AIRBYTE_MCP_TELEMETRY_SALT`, falling back to PyAirbyte's existing persisted anonymous analytics ID. Hosted deployments must set the env var explicitly — the fallback is regenerated per container, which would silently degrade the surrogate from per-caller to per-instance. When the user has opted out of analytics there is no salt and no attribution is emitted.

`extra_properties` is resolved at emit time inside `on_call_tool`, so this runs inside the live request scope and needs no `fastmcp-extensions` change.

## Test plan

- `uv run ruff format --check .` and `uv run ruff check .` — clean.
- `uv run pytest tests/unit_tests/test_mcp_telemetry.py tests/unit_tests/test_mcp_http_main.py` — 32 passed.
- New unit tests assert the keyed HMAC construction, salt precedence and fallback, that a different salt yields a different digest, that the same IP on two endpoints yields two different `caller_hash` values, first-hop selection from a multi-entry `X-Forwarded-For`, Airbyte-owned vs third-party endpoint handling, and — explicitly — that the raw IP, raw session ID, raw subject and third-party hostname appear nowhere in the emitted payload.

Not yet exercised against a live hosted deployment: that needs the PyAirbyte release, a Cloud MCP pin bump, and the salt provisioned in the deploy environment. Warehouse [#3472](https://github.com/airbytehq/airbyte-warehouse/pull/3472) will absorb these columns rather than landing without them.

Relates to #1129.

Link to Devin session: https://app.devin.ai/sessions/c6e358d911424ec298aa41df9110ed3b
Open in Devin Desktop: https://app.devin.ai/desktop/session/c6e358d911424ec298aa41df9110ed3b?variant=devin
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Improved MCP telemetry privacy with anonymized caller attribution.
  * Added configurable anonymization and recognized public-domain controls.
  * Hosted connections now resolve attribution dynamically, while local connections use the local analytics identity.

* **Bug Fixes**
  * Prevented caller IP addresses from being used as a telemetry fallback.
  * Improved telemetry behavior across default, offline, and do-not-track modes.

* **Chores**
  * Updated the MCP extensions integration to a newer compatible version.


airbyte/mcp/_telemetry_attribution.py
```diff
@@ -0,0 +1,169 @@
+# Copyright (c) 2025 Airbyte, Inc., all rights reserved.
+"""Privacy-safe attribution properties for MCP tool-call telemetry.
+
+Hosted deployments should set `AIRBYTE_MCP_TELEMETRY_SALT` explicitly. The
+analytics-ID fallback is regenerated per container, which would otherwise make
+the surrogates identify instances instead of callers.
+"""
```
aaronsteers: This doesn't need to be named MCP-specific. Even if currently this is only MCP-leveraged, there's no reason it should be named or called out explicitly in regards to MCP.

```suggestion
"""Hash and obfuscation processes for Airbyte anonymized telemetry.

Deployments can optionally set `AIRBYTE_TELEMETRY_ANONYMIZATION_SALT` explicitly. The analytics-ID fallback is regenerated per container, which would otherwise make the surrogates identify instances instead of callers.
 
Deployments can use `DO_NOT_TRACK` to opt out of anonymized telemetry.
"""
```

tests/unit_tests/test_telemetry_anonymization.py
aaronsteers, line 1: It doesn't look like you properly condensed your tests. You should use fewer tests and higher degrees of parameterization decorators to make them more dense and have fewer tests for a reviewer to read through.

airbyte/mcp/_telemetry_attribution.py
aaronsteers: > the module itself still lives at airbyte/mcp/_telemetry_attribution.py.

Move it.

aaronsteers: Should be adjacent to other telemetry module(s).

airbyte/_util/telemetry_anonymization.py
aaronsteers, line 1: Collapse this into _util.telemetry.py. Not enough here to justify a different module.

airbyte/_util/telemetry_anonymization.py
```diff
@@ -0,0 +1,167 @@
...
+from collections.abc import Callable  # noqa: TC003
+from functools import lru_cache
+from typing import TypeVar
+from urllib.parse import urlsplit
+
+from fastmcp.server.dependencies import (
+    get_access_token,
+    get_context,
+    get_http_request,
+)
```
aaronsteers: Remove these dependencies, moving into an mcp module or (better) into the fastmcp extensions package.

airbyte/mcp/server.py
aaronsteers, line 1: This is the probably the right home for anything that can't live generically in the telemetry.py module (without importing fastmcp) and which can't live in fastmcp-extensions.

airbyte/_util/telemetry_anonymization.py
aaronsteers: Update this PR to presume the fastmcp extensions PR has already landed.

PR #1132, opened 2026-08-28
feat(mcp): add typed Airbyte Agents entity tools

aaronsteers:
## Summary

Requested by AJ Steers as the follow-up to the merged Agents base work (#1127), which shipped the generic pair `execute_agent_connector_ro` / `execute_agent_connector`. Those two are enough to reach any connector action, but they push two costs onto the calling agent: it has to know the action vocabulary, and it gets back `result: Any` with no shape.

This adds six action-specific MCP tools in a new `airbyte/mcp/agents_entities.py`, all thin presentation over the same hidden `_execute()` helper the generic tools use — no new transport, no new auth path, no change to `airbyte/agents/`.

```python
# read, annotated read_only=True idempotent=True, so they survive readonly mode
list_agent_entities(...)   -> AgentEntityPageResult   # action="list"
search_agent_entities(..., use_api_search=False) -> AgentEntityPageResult  # "search" | "api_search"
get_agent_entity(...)      -> AgentEntityResult      # action="get"

# write
create_agent_entity(...)   -> AgentWriteResult
update_agent_entity(...)   -> AgentWriteResult
delete_agent_entity(...)   -> AgentWriteResult       # destructive=True
```

Two things this buys beyond naming:

1. **Readonly mode gets real reads.** `readOnlyHint` is a static annotation read by the server's readonly filter before any call, so the generic write-capable tool vanishes whole in readonly mode. The three read tools here are statically read-only, so a readonly deployment can still list, search, and get.
2. **Typed returns instead of `Any`.** The generic result carries `result: Any`; these return the shape the action actually produces — a page (`entities`, `has_next_page`, `end_cursor`), a single `entity | None`, or a write ack. Coercion is strict rather than permissive: `get_agent_entity` raises `PyAirbyteInputError` if the connector returns more than one entity, and any non-object payload is rejected rather than passed through as a mystery blob.

`entity_type` stays a free string, since `inspect` publishes only Context Store entities, not everything `execute()` accepts — typing it would reject legitimate calls.

Deliberately not done: no seventh tool for `api_search` (same read shape as `search`, so it's a `use_api_search` flag), and no separate `record`/`fields` argument on the writes — the payload stays in `api_args`, which is where the connector-specific arguments already live.

`workspace_id` behaves exactly as in the generic tools: exposed as an argument, hidden when statically configured, and the connector is resolved through its workspace so a connector belonging to another workspace is rejected before any action runs.

## Test Plan

- `uv run pytest tests/unit_tests/test_mcp_agent_entities.py tests/unit_tests/test_mcp_agents.py` — 13 new cases, 43 total, passing. Covers page shaping and cursor passthrough, the `use_api_search` action switch, all six coercion outcomes for `get_agent_entity` (including the multi-match and non-object rejections), per-tool action/payload forwarding on the writes, and a live-server assertion that the read tools register with `readOnlyHint=true` and `delete_agent_entity` with `destructiveHint=true`.
- `uv run ruff format` / `uv run ruff check` / `uv run pyrefly check` clean.
- `uv run poe mcp-docs-md` regenerates cleanly and emits `docs/mcp-generated/agents_entities.md` (gitignored, generated at build time).
- Live Agents execution is not exercised here; as in #1127 there are no cassettes for the Agents API.

Open in Devin Desktop: https://app.devin.ai/desktop/session/57a0c3e7b98f4c52a9c09a5cd721ee3a?variant=devin
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Added Airbyte Agents entity tools for listing, searching, viewing, creating, updating, and deleting entities.
  * Added pagination, field selection, excluded-field filtering, search options, and workspace-aware configuration.
  * Registered the new tools with the MCP server, including clear status and warning information in results.

* **Tests**
  * Added coverage for entity retrieval, pagination, search handling, payload validation, write operations, and server registration.

### DIFF
--- a/airbyte/mcp/__init__.py
+++ b/airbyte/mcp/__init__.py
@@ -312,11 +312,12 @@ For issues and questions:
 
 """  # noqa: D415
 
-from airbyte.mcp import agents, cloud, local, registry, server
+from airbyte.mcp import agents, agents_entities, cloud, local, registry, server
 
 
 __all__: list[str] = [
     "agents",
+    "agents_entities",
     "cloud",
     "local",
     "registry",
--- a/airbyte/mcp/agents_entities.py
+++ b/airbyte/mcp/agents_entities.py
@@ -1,0 +1,553 @@
+# Copyright (c) 2026 Airbyte, Inc., all rights reserved.
+"""Airbyte Agents entity MCP operations.
+
+.. include:: ../../docs/mcp-generated/agents_entities.md
+"""
+
+# No public Python API — MCP primitives are registered via decorators and
+# documented via the generated Markdown include above. Setting `__all__` to an
+# empty list tells pdoc (and other doc tools) not to surface the individual
+# tool / helper definitions as a redundant "API Documentation" list.
+__all__: list[str] = []
+
+from typing import Annotated, Any
+
+from fastmcp import Context, FastMCP
+from fastmcp_extensions import mcp_tool, register_mcp_tools
+from pydantic import BaseModel, Field
+
+from airbyte.exceptions import PyAirbyteInputError
+from airbyte.mcp._tool_utils import AIRBYTE_CLOUD_WORKSPACE_ID_IS_SET
+from airbyte.mcp.agents import (
+    AGENTS_AUTH_TIP_TEXT,
+    WORKSPACE_ID_TIP_TEXT,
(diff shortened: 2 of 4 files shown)


aaronsteers:
Closed. (Not doing.)

PR #1135, opened 2026-08-28
fix(mcp): gate Agents MCP tools behind insiders mode, add missing org-id header

aaronsteers:
## Summary

The Agents MCP tools added in #1127 are advertised to every caller today, but most organizations have no Airbyte Agents subscription, so those tools cost context and produce authorization failures nobody can act on. This hides them by default and makes the failures readable when they do happen.

Three changes:

**1. Insiders gate on the `agents` module.** `MCP_INSIDERS_MODULES = {"agents"}` marks modules that are not part of the default surface. `airbyte_module_filter` gains one branch, ordered so an explicit exclusion still wins:

```python
if exclude_modules and tool_module in exclude_modules:
    return False
if tool_module in MCP_INSIDERS_MODULES and not _insiders_enabled(app):
    return tool_module in include_modules   # naming the module is itself an opt-in
if include_modules:
    return tool_module in include_modules
return True
```

So Agents tools appear when `AIRBYTE_MCP_INSIDERS` / `X-MCP-Insiders` is set (they join the normal surface), or when the include list names `agents` (`AIRBYTE_MCP_DOMAINS=agents` still yields Agent-only mode, unchanged). Names follow the [GitHub MCP server's configuration surface](https://github.com/github/github-mcp-server/blob/HEAD/docs/server-configuration.md), which is the closest thing to a convention here — MCP itself has no annotation or header for preview/subscription-gated tools, and tool annotations are client hints, so the filtering has to be server-side regardless.

This is a caller-settable gate that *widens* the tool surface, unlike `AIRBYTE_MCP_TRUSTED_EXECUTION` which deliberately has no header. That is safe because it changes only which tools are advertised: each Agents tool still authorizes every call against the Airbyte API, and no entitlement probe happens during `list_tools`.

**2. `X-Airbyte-Organization-Id`.** Pairs with the existing `AIRBYTE_CLOUD_ORGANIZATION_ID` env var, so hosted callers can scope organization-level listings per request. An explicit `organization_id` tool argument still wins over it.

**3. Readable authorization failures.** `401`/`403` from the Agents API previously surfaced as an `AirbyteError` traceback. The four Agents tools now return their normal result shape with an empty payload and a `message` explaining which of the two situations applies — invalid credentials, versus authenticated-but-not-entitled/no workspace access — and telling the agent to ask rather than retry. Anything else re-raises with its original context, following the `_handle_discovery_permission_error` precedent in `airbyte/mcp/cloud.py`.

Naming/deprecation work discussed alongside this (`AIRBYTE_MCP_DOMAINS` → include/exclude-modules, `AIRBYTE_MCP_HOSTING_URL`, adopting more of GitHub's header names, and a documented index of every setting) is intentionally not here and follows in a separate PR.

Open in Devin Desktop: https://app.devin.ai/desktop/session/57a0c3e7b98f4c52a9c09a5cd721ee3a?variant=devin
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Added organization-scoped configuration for MCP requests through supported configuration sources.
  * Added optional insiders mode to control access to select tools.
  * Added organization-aware workspace and connector operations.
* **Bug Fixes**
  * Improved Agents access handling with clear messages for unauthorized or forbidden requests.
  * Preserved detailed error handling for unrelated failures.
  * Improved handling of blank or unrecognized boolean configuration values.
* **Tests**
  * Added coverage for organization ID resolution, access-control responses, and tool visibility rules.

> [!IMPORTANT]
> **Auto-merge enabled.**
> 

> [!NOTE]
> **Auto-merge may have been disabled. Please check the PR status to confirm.**


tests/unit_tests/test_mcp_tool_filters.py
aaronsteers, line 1: Make sure all tests are fully condensed. Should always prefer fewer, generic tests with higher degrees of parametrization decorator.

airbyte/mcp/_tool_utils.py
```diff
@@ -157,6 +164,20 @@ def check_guid_created_in_session(guid: str) -> None:
 )
 """Config arg for legacy AIRBYTE_MCP_DOMAINS env var."""
 
+INSIDERS_CONFIG_ARG = MCPServerConfigArg(
+    name=MCP_CONFIG_INSIDERS,
+    http_header_key=MCP_INSIDERS_HEADER,
+    env_var=MCP_INSIDERS_ENV_VAR,
+    default="0",
```
aaronsteers: Shouldn't this default to ""?

airbyte/mcp/_tool_utils.py
```diff
@@ -429,21 +463,66 @@ def airbyte_readonly_mode_filter(tool: Tool, app: FastMCP) -> bool:
...
+    Matching is case-insensitive and ignores surrounding whitespace. The accepted spellings
+    mirror `airbyte.constants._str_to_bool`, but an unrecognized value is `None` here rather
+    than true, so that a caller-controlled fallback can still apply.
+    """
+    normalized = (value or "").strip().lower()
+    if normalized in {"1", "true", "t", "yes", "y", "on"}:
+        return True
+    if normalized in {"0", "false", "f", "no", "n", "off"}:
+        return False
+    return None
```
aaronsteers: We should already have something like this, and shouldn't be in an mcp-specific module.

airbyte/constants.py
```diff
@@ -312,6 +312,50 @@ def is_hosted_mcp_mode() -> bool:
...
+
+MCP_INSIDERS_HEADER: str = "X-MCP-Insiders"
+"""HTTP header key that advertises insiders MCP tools.
+
+Unlike `AIRBYTE_MCP_TRUSTED_EXECUTION`, this gate *widens* the tool surface and is still
+caller-settable, which is a deliberate exception: it changes only which tools are
+advertised, and every insiders tool independently enforces its own authorization against
+the Airbyte API. It is not an access-control boundary. A hosted deployment that sets
+`MCP_INSIDERS_ENV_VAR` explicitly overrides this header in either direction.
+"""
```
aaronsteers, line 365: This is the wrong place for explaining our rationale. Remove all of the "TMI" so we focus on what a user needs to know. The contributor guidance and rationale doesn't belong here. Probably in CONTRIBUTING.md (later) and not in this PR.

aaronsteers: Specifically points like this:

> The name follows the GitHub MCP server's `GITHUB_INSIDERS` / `X-MCP-Insiders` pair, which
> is the closest thing to a convention for this setting:
https://github.com/github/github-mcp-server/blob/HEAD/docs/server-configuration.md

aaronsteers: Yes

airbyte/mcp/_tool_utils.py
aaronsteers: No. I think the whole thing can be tri-state. Hence doing it in one place. We should basically always treat empty as `null`, so this behavior should become our default processor.

PR #1144, opened 2026-09-07
fix: pass connector config to declarative sources

sagarrh:
`DeclarativeExecutor` initialized `_config_dict` to an empty dict and never populated it from the connector config. `execute()` then built the CDK source via `ConcurrentDeclarativeSource(config=self._config_dict, ...)`, so any manifest interpolation of `{{ config[...] }}` resolved against an empty mapping.

Manifest-only connectors that template credentials into their authenticator therefore failed to authenticate. For example, `source-google-calendar` interpolates `client_id`, `client_secret` and `refresh_token` into an `OAuthAuthenticator`, and every sync failed with:

    Stream colors is not available: OAuth access token refresh request failed.

The credentials were valid; they simply never reached the source.

The declarative source resolves interpolations at construction time, so the config must be supplied then rather than left for the entrypoint to read later. `execute()` now reads the config from the `--config` argument it was invoked with and merges it over any injected components, which keeps `components.py` injection working.

`_config_dict` is left untouched, and the public `declarative_source` property is retained with its previous no-config behavior.

Verified against a live `source-google-calendar` sync, which now succeeds where it previously failed.

## Summary by CodeRabbit

* **New Features**
  * Added support for loading connector configuration from a JSON file supplied through the command line.
  * Runtime configuration is now combined with injected components when creating declarative sources.
  * Declarative source construction can use a configurable builder.

* **Bug Fixes**
  * Improved handling of missing, unreadable, invalid, or incomplete configuration arguments.
  * Commands that do not require configuration continue to run without one.

### DIFF
--- a/airbyte/_executors/declarative.py
+++ b/airbyte/_executors/declarative.py
@@ -4,6 +4,7 @@
 from __future__ import annotations
 
 import hashlib
+import json
 import warnings
 from pathlib import Path
 from typing import IO, TYPE_CHECKING, Any, cast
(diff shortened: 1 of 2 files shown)


aaronsteers:
/test-pr
> PR test job started... [Check job output.][1]

[1]: https://github.com/airbytehq/PyAirbyte/actions/runs/34415510264
> ✅ Tests passed.

aaronsteers (approved):
🚀

aaronsteers:
@sagarrh - Thank you for your contribution! 🙏 

Will release shortly as `v0.59.1` 🚀

PR #1154, opened 2026-09-10
feat(mcp): support `sql_select` action on agent connector tools

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** Lets agents run the new passthrough SQL action against destination connectors through the MCP connector tools. Previously the tools rejected the action outright, could not find destination connectors at all, and choked on the metadata shape the live API returns for SQL results.

Specifically, this adds `"sql_select"` to `AgentReadAction` / `AgentAction`, adds a Cloud-workspace destination fallback in `_get_agent_connector`, and makes `AgentExecuteResult` accept `null` for `connector_metadata` / `execution_metadata`.

</td></tr></table>

Requested by AJ Steers (@aaronsteers) to validate passthrough SQL end-to-end via `cloud-mcp-preview`.

## Changes

- Add `sql_select` to `AgentReadAction` (read-only tool) and `AgentAction`; extend the `action` descriptions so agents pass `sql`, `sql_dialect`, and optional `dry_run` in `api_args` (`entity_type` is ignored server-side). No plumbing change: `api_args` already flows into request `params`.
- `_get_agent_connector`: destinations are not listed by the Agents API, so when the Agents workspace lookup raises "No connector found", check `CloudWorkspace.list_destinations()` for the same workspace; if the ID is a destination there, build an explicit-ID `AgentConnector`, otherwise re-raise. Unrelated `AirbyteError`s are not swallowed. No bypass flag (`verify_in_workspace` was removed per review).
- `AgentExecuteResult`: `field_validator(mode="before")` maps `None` → `{}` for `connector_metadata` / `execution_metadata` (the live API returns `null` for SQL results), so `has_next_page` / `end_cursor` keep working.
- Call sites use keyword args for `_get_agent_connector(...)` and `.execute(...)`.

## Review Spotlight

- [`agents.py` `_get_agent_connector`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789078510-mcp-sql-select/airbyte/mcp/agents.py#L268-L295): the Cloud-destination fallback and its error filter
- [`models.py` `_none_to_empty`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789078510-mcp-sql-select/airbyte/agents/models.py): null-metadata normalization

## Risks

- The fallback is action-agnostic (applies to any action whose connector ID is a Cloud destination in the caller's workspace). Membership is still validated; which actions a destination supports is enforced server-side by the Agents API. Discussed and accepted in review.

## Test plan

- `uv run pytest tests/unit_tests/test_mcp_agents.py tests/unit_tests/test_agents.py` (39 + 67 passed)
- `uv run ruff check` / `ruff format --check` on touched files; `mypy airbyte/agents/models.py`
- Live: `cloud-mcp-preview` deployed from `f2a2946` and exercised via `execute_agent_connector_ro` against real Snowflake and BigQuery destinations: `SELECT`, `dry_run`, `CURRENT_TIMESTAMP`, write-SQL rejection, and wrong-dialect rejection all behave identically to the direct `api.airbyte.ai` calls. (`SHOW TABLES` failures observed are destination/Sonar-side and unrelated to this PR.)

Link to Devin session: https://app.devin.ai/sessions/7a60f936d7984f659f5ff4722d0ee6ee
Open in Devin Desktop: https://app.devin.ai/desktop/session/7a60f936d7984f659f5ff4722d0ee6ee?variant=devin

### DIFF
--- a/airbyte/mcp/agents.py
+++ b/airbyte/mcp/agents.py
@@ -329,9 +329,9 @@ def _execute(  # noqa: PLR0913  # Mirrors the tool signatures it serves.
             workspace_id=workspace_id,
             organization_id=organization_id,
         ).execute(
-            entity_type,
-            action,
-            _resolve_api_args(api_args),
+            entity_type=entity_type,
+            action=action,
+            api_args=_resolve_api_args(api_args),
             select_fields=resolve_list_of_strings(select_fields),
             exclude_fields=resolve_list_of_strings(exclude_fields),
             page_size=page_size,
--- a/airbyte/mcp/agents.py
+++ b/airbyte/mcp/agents.py
@@ -324,10 +324,10 @@ def _execute(  # noqa: PLR0913  # Mirrors the tool signatures it serves.
 
     try:
         result = _get_agent_connector(
-            ctx,
-            connector_id,
-            workspace_id,
-            organization_id,
+            ctx=ctx,
+            connector_id=connector_id,
+            workspace_id=workspace_id,
+            organization_id=organization_id,
         ).execute(
             entity_type,
             action,
--- a/airbyte/mcp/agents.py
+++ b/airbyte/mcp/agents.py
@@ -47,7 +47,7 @@ from airbyte.constants import (
 from airbyte.exceptions import AirbyteError, PyAirbyteInputError
 from airbyte.mcp._arg_resolvers import resolve_list_of_strings
 from airbyte.mcp._tool_utils import AIRBYTE_CLOUD_WORKSPACE_ID_IS_SET
-from airbyte.mcp.cloud import _add_defaults_for_exclude_args
+from airbyte.mcp.cloud import _add_defaults_for_exclude_args, _get_cloud_workspace
 
 
 AgentReadAction = Literal["list", "get", "search", "api_search", "sql_select"]
@@ -270,8 +270,6 @@ def _get_agent_connector(
     connector_id: str,
     workspace_id: str | None = None,
     organization_id: str | None = None,
-    *,
-    verify_in_workspace: bool = True,
 ) -> AgentConnector:
     """Get an `AgentConnector` from its workspace, using MCP config.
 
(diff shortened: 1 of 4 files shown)


airbyte/mcp/agents.py
```diff
@@ -264,14 +270,23 @@ def _get_agent_connector(
...
     The Agents API addresses a connector by ID alone, but the connector is fetched through
     its workspace anyway, so a connector ID belonging to another workspace raises before
     any action runs.
+
+    Destination connectors targeted by `sql_select` are not listed by the workspace connectors
+    endpoint, so they are addressed by ID alone and the Agents API enforces workspace/org
+    authorization.
     """
-    return _get_agent_workspace(ctx, workspace_id, organization_id).get_connector(connector_id)
+    workspace = _get_agent_workspace(ctx, workspace_id, organization_id)
```
aaronsteers, line 283: Since agent workspaces contain a subset of the cloud workspace, I think we could verify against the cloud workspace and not have to bypass verification in those cases where the destination is not a proper Sonar connector.

airbyte/mcp/agents.py
```diff
@@ -264,14 +270,23 @@ def _get_agent_connector(
     connector_id: str,
     workspace_id: str | None = None,
     organization_id: str | None = None,
+    *,
+    verify_in_workspace: bool = True,
```
aaronsteers: I prefer to not add this argument if we don't need it.

airbyte/mcp/agents.py
```diff
@@ -302,7 +323,12 @@ def _execute(  # noqa: PLR0913  # Mirrors the tool signatures it serves.
         )
 
     try:
-        result = _get_agent_connector(ctx, connector_id, workspace_id, organization_id).execute(
+        result = _get_agent_connector(
+            ctx,
+            connector_id,
+            workspace_id,
+            organization_id,
+        ).execute(
```
aaronsteers, line 331: Always use named keyword arguments if there are more than two or three input args. Each arg on its own line.

aaronsteers: Same for the execute call directly below.

PR #1158, opened 2026-09-12
feat(cloud): resolve live default workspace and its org in `get_default_cloud_context`

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** A single call to the default-context tool now tells an agent which workspace and organization it is operating in, even when the user's permission list still contains grants for deleted workspaces, and every listed member workspace carries its parent organization.

Specifically, `CloudClient` validates direct workspace grants once under `MAX_WORKSPACES_TO_VALIDATE = 25`, `resolve_default_workspace_id` picks the sole *live* grant, and `CloudDefaultContextInfo` / `get_default_cloud_context` gain `default_workspace_name`, `default_workspace_verified`, `default_organization_id`, `default_organization_name`, `unvalidated_workspace_count`; `CloudWorkspaceInfo` gains `organization_name`.

</td></tr></table>

Part of a stacked pair; #1159 (removing `check_airbyte_cloud_workspace`) targets this branch.

Requested by AJ Steers. Related: [PLAT-1301](https://linear.app/airbyteio/issue/PLAT-1301) (platform-side root cause, fix in https://github.com/airbytehq/airbyte-platform-internal/pull/19502), [PLAT-1300](https://linear.app/airbyteio/issue/PLAT-1300).

## Why

Observed with Goose Desktop against hosted cloud-mcp: for a user with one live direct workspace plus several grants on tombstoned workspaces, `get_default_cloud_context` returned `default_workspace_id: null` because the fallback did `len(direct_workspace_ids) != 1` on the raw permission list. Separately, `organization_id` was always `null` in `member_workspaces` (the public `WorkspaceResponse` has no org field), so models made a follow-up `check_airbyte_cloud_workspace` call just to learn the org.

## Changes

- `_validate_direct_workspaces()` (new, cached per client): probes the first `MAX_WORKSPACES_TO_VALIDATE` (25) direct grants via public `GET /workspaces/{id}` (404 ⇒ stale, dropped), enriches each live one with `organization_id`/`organization_name` via Config API `workspaces/get_organization_info` (cached `_get_workspace_organization`; `AirbyteError`/`NotImplementedError` ⇒ org left `None`, workspace kept), and returns `(live_workspaces, unvalidated_count)`. Replaces `MAX_MEMBER_WORKSPACES` / `MAX_DEFAULT_WORKSPACE_CANDIDATES`.
- `resolve_default_workspace_id`: precedence unchanged (configured → user `defaultWorkspaceId` → direct grants); the direct-grant step now returns the sole live workspace only when `unvalidated_count == 0`.
- `_list_member_workspaces` (`list_cloud_workspaces` with `MEMBER_OF`): reuses the cached validated set, then keeps walking grants beyond the cap (public GET only, no org enrichment) so the explicit listing still returns everything, honoring `limit`.
- `get_default_context_for_user`: `member_workspaces` = the validated live set (with org fields); `member_workspaces_truncated` / `unvalidated_workspace_count` reflect grants beyond the cap; `default_*` fields come from the resolved workspace and its org; `default_workspace_verified` is `False` when a configured/user default cannot be fetched. `member_organizations` remains explicit org memberships only (the default workspace's parent org is *not* appended).
- MCP `get_default_cloud_context`: result model mirrors the new fields; message leads with `Resolved default workspace <name> (<id>) in organization <name> (<id>).` or, for an unverifiable configured ID, `Default workspace ID <id> is configured but could not be verified ...`; appends `N additional direct workspace grant(s) were not validated because this call checks at most 25; use list_cloud_workspaces to see them.` when capped. Docstring marks it as the one-call orientation entry point; `check_airbyte_cloud_workspace` docstring points orientation to it.
- `docs/CONTRIBUTING.md` example uses `get_default_cloud_context`.

## Review Spotlight

- `airbyte/cloud/client.py` → `_validate_direct_workspaces` (single cached pass, cap, org enrichment) and `resolve_default_workspace_id`
- `airbyte/cloud/client.py` → `get_default_context_for_user` (new fields, truncation semantics)
- `airbyte/mcp/cloud.py` → `get_default_cloud_context` message composition

## Risks

- Up to 25 public `GET /workspaces/{id}` + 25 `get_organization_info` calls per client for users with many direct grants (cached; typical users have 1–3). Users above the cap still get their first 25 validated workspaces plus an explicit unvalidated count; default inference is withheld for them.
- Stale-grant detection still relies on 404; the root fix is https://github.com/airbytehq/airbyte-platform-internal/pull/19502, after which the probes see no stale grants.

## Test plan

- `uv run ruff format .` / `uv run ruff check .` → clean
- `uv run --with mypy mypy airbyte/cloud/client.py airbyte/cloud/models.py airbyte/mcp/cloud.py` → no issues (full-repo `mypy .` is blocked locally by a pre-existing duplicate `run` module under `tests/integration_tests/fixtures/source-broken/build/`)
- `uv run pytest tests/unit_tests/test_cloud_client.py tests/unit_tests/test_mcp_cloud.py -q` → 52 passed
- Unit coverage: one live + stale grants resolves the live one with org fields populated; two live grants → `None`; 26 grants → exactly 25 `get_workspace` probes and 25 org calls, `unvalidated_workspace_count == 1`, 25 `member_workspaces`, no default inferred, while `list_workspaces(MEMBER_OF)` still returns all 26 (and `limit=3` stops at 3); org-info failure keeps the workspace with `organization_id=None`; configured workspace still wins; unverifiable configured ID → `default_workspace_verified=False` and the "could not be verified" message.

Link to Devin session: https://app.devin.ai/sessions/901bc3edeba246eab5d18ae328943251
Open in Devin Desktop: https://app.devin.ai/desktop/session/901bc3edeba246eab5d18ae328943251?variant=devin
Requested by: @aaronsteers

## Summary by CodeRabbit

* **New Features**
  * Improved default workspace resolution and verification.
  * Added organization details to workspace and default-context information.
  * Added status reporting for unverified workspaces and unvalidated workspace grants.
  * Limited initial workspace validation while continuing to support filtered and limited listings.
  * Preserved organization details for directly associated workspaces.

* **Documentation**
  * Updated cloud MCP tool testing guidance.


airbyte/cloud/client.py
```diff
@@ -760,7 +767,30 @@ def _get_direct_workspace_info(self, workspace_id: str) -> CloudWorkspaceInfo |
...
+            organization = api_util.get_workspace_organization_info(
+                workspace_id=workspace_id,
+                api_root=self.public_api_root,
+                config_api_root=self.config_api_root,
+                client_id=self.client_id,
+                client_secret=self.client_secret,
+                bearer_token=self._get_config_api_bearer_token(),
+            )
+        except AirbyteError:
+            pass
```
aaronsteers: At minimum, an inline comment should state what this case is that we're passing on.

airbyte/cloud/client.py
```diff
@@ -106,6 +106,7 @@
 
 MAX_ORGANIZATION_CANDIDATES = 10
 MAX_MEMBER_WORKSPACES = 25
+MAX_DEFAULT_WORKSPACE_CANDIDATES = 10
```
aaronsteers: If I understand correctly, there are two things we're doing here:
1. We're deciding how many workspaces to validate before giving up.
2. We're deciding how many workspaces to evaluate in consideration of the user's default workspace when trying to whittle down a large N to potentially just one.
We need to make sure we handle both of these cases, and for that reason, I'm wondering if we should just combine this or rename it to something like "Max Workspaces to validate".

And if we reach the limit of max workspaces to validate, we would still return the n+1 membership back to the caller (perhaps with a text message warning that the trailing n number of workspaces were not validated due to reaching the max validation limit for this call). 

Please confirm for me that our `get_default_cloud_context` function is checking for default context (meaning a one-best record), but also returning the list of workspaces and organizations that the agent might want to subsequently check if exactly one cannot be determined.

aaronsteers: Fantastic. And confirm for me that we are returning the member workspaces and member organizations with their names. Are we returning workspaces with the parent organization context, even for the member workspaces result set, and not just the default or configured context? Ideally, we'd also return the member workspaces with a parent organization ID if not also the organization name. Although, if this organization ID is already going to be present in the member organizations, then we could avoid the duplication and just pass the ID. The caller has everything they need to map between workspaces that they're a member of, the organization IDs that contain them, and, from the organizations list, can do a mapping of ID to name if they need to. Does that sound right?

aaronsteers:
coderabbit review

PR #1167, opened 2026-09-14
feat(mcp): gate agents tools to Cloud deployments via API root overrides

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** The Airbyte Agents MCP tools now switch themselves off whenever the MCP server is pointed at an OSS or self-managed Airbyte instead of Airbyte Cloud, since the Agents API only exists on Cloud. The signal is an overridden Cloud API root, which is currently the only deployment indicator we have. Operators who proxy Cloud (or test against a non-production Agents endpoint) can opt back in by setting an explicit Agents API root.

Specifically, `airbyte_module_filter` hides the `agents` module unless `is_agents_api_available(...)`; at runtime `get_agents_api_root(credentials)` raises the new `AirbyteAgentsUnavailableError` when no Agents API exists, and the shared `UserFacingErrorMiddleware` from `fastmcp-extensions` (airbytehq/fastmcp-extensions#118) turns that (and other user-facing errors) into a concise `ToolError`.

</td></tr></table>

Resolves [AGENTIC-2233](https://linear.app/airbyteio/issue/AGENTIC-2233/gate-pyairbyte-mcp-agents-tools-to-cloud-deployments). Requested by AJ Steers.

## Changes

- `airbyte/constants.py`: add `AGENTS_API_ROOT_ENV_VAR = "AIRBYTE_AGENTS_API_URL"`.
- `airbyte/_util/deployment.py` (new): deployment detection shared by Agents and MCP.
  - `is_public_cloud(public_api_root=None, config_api_root=None)` — resolves the effective roots via the existing `resolve_cloud_api_url` / `resolve_cloud_config_api_url` / `get_config_api_root` and compares to `CLOUD_API_ROOT` / `CLOUD_CONFIG_API_ROOT` (so env overrides take precedence exactly as they do for `CloudClientConfig`).
  - `get_agents_api_root_override()` — non-blank `AIRBYTE_AGENTS_API_URL`, normalized.
  - `is_agents_api_available(...)` — Agents override set, or `is_public_cloud(...)`.
- `airbyte/agents/_api_util.py`:
  - `check_public_cloud_api_roots` (used by `from_cloud_*` conversions) honors the Agents override.
  - `get_agents_api_root(credentials)` — mirrors `get_config_api_root`: env override → hosted root → raise `AirbyteAgentsUnavailableError`. `make_agents_api_request` resolves it before the bearer-token exchange, so this is the single runtime failure point.
- `airbyte/agents/workspaces.py`, `organizations.py`: `AgentWorkspace`/`AgentOrganization.__init__` accept `public_api_root` / `config_api_root` (forwarded to `_AirbyteCredentials.from_auth`); `from_cloud_*`, `_workspace_from_info`, and the MCP `_get_agent_*` helpers carry the configured Cloud roots so the client-credentials token exchange stays on the same deployment instead of defaulting to public Cloud.
- `airbyte/exceptions.py`: add `AirbyteAgentsUnavailableError(AirbyteError)` with guidance pointing at the public roots / `AIRBYTE_AGENTS_API_URL`.
- `airbyte/mcp/_tool_utils.py`: `is_agents_api_available(app_or_ctx)` reads the MCP `api_url` / `config_api_url` config and delegates to `airbyte._util.deployment`; visibility gate for `_AGENTS_MCP_MODULE = "agents"` in `airbyte_module_filter`, placed before the insiders block so it wins over insiders mode and over an include list naming `agents`.
- `airbyte/mcp/_error_handling.py` (new): only the PyAirbyte-specific parts — `MCP_TOOL_USER_FACING_ERRORS` (`PyAirbyteInputError`, `AirbyteMCPError`, `AirbyteAgentsUnavailableError`) and `format_user_facing_error(error) -> message + guidance`. The generic middleware lives in `fastmcp-extensions` (airbytehq/fastmcp-extensions#118) and is registered via `mcp_server(..., user_facing_errors=MCP_TOOL_USER_FACING_ERRORS, user_facing_error_formatter=format_user_facing_error)` in `server.py`; other exceptions pass through to FastMCP's default handling. Tool modules keep using `fastmcp_extensions.mcp_tool` unchanged. No MCP-layer hard-fail in `agents.py`.
- `pyproject.toml` / `uv.lock`: bump `fastmcp-extensions>=0.23.0,<1.0.0` (the release that ships airbytehq/fastmcp-extensions#118); no git source pin.
- Tests: `test_mcp_cloud_deployment_gate.py`, `test_mcp_tool_errors.py` (in-memory `FastMCP` + `Client` through the middleware), `test_deployment.py` (new); override/env-var cases in `test_mcp_tool_filters.py`, root resolution and root propagation in `test_agents.py` / `test_mcp_agents.py`.

## Review Spotlight

- [`_util/deployment.py`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789418525-gate-agents-tools-to-cloud/airbyte/_util/deployment.py): override detection and Agents availability; [`agents/_api_util.py` `get_agents_api_root`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789418525-gate-agents-tools-to-cloud/airbyte/agents/_api_util.py#L49-L90)
- [`_error_handling.py`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789418525-gate-agents-tools-to-cloud/airbyte/mcp/_error_handling.py): which error bases become concise `ToolError`s and how they're formatted (middleware itself is in airbytehq/fastmcp-extensions#118)
- [`_tool_utils.py` filter ordering](https://github.com/airbytehq/PyAirbyte/blob/devin/1789418525-gate-agents-tools-to-cloud/airbyte/mcp/_tool_utils.py): Cloud-only gate before the insiders gate

## Gating

- Visibility: `agents` tools are hidden when `AIRBYTE_CLOUD_API_URL` / `AIRBYTE_CLOUD_CONFIG_API_URL` are overridden, unless `AIRBYTE_AGENTS_API_URL` is set. The API-root config args have no HTTP-header source, so a remote caller cannot flip this gate.
- Runtime: any Agents API call resolves its root through `get_agents_api_root`, which raises `AirbyteAgentsUnavailableError` in the same condition.
- The insiders mechanism (`AIRBYTE_MCP_INSIDERS` / `X-MCP-Insiders`) is untouched and still applies on top.

## Risks

- The middleware converts `PyAirbyteInputError` / `AirbyteMCPError` raised by *any* tool into `ToolError` (message + guidance, no chained traceback). Client-visible text is shorter and more actionable; the server still logs via FastMCP's own handler.
- `AIRBYTE_AGENTS_API_URL` is not scheme-validated, matching the other root overrides (local `http://` dev is a stated use case).

## Test plan

- `pytest tests/unit_tests/test_mcp_cloud_deployment_gate.py tests/unit_tests/test_mcp_tool_filters.py tests/unit_tests/test_mcp_agents.py tests/unit_tests/test_mcp_tool_errors.py tests/unit_tests/test_agents.py`
- `pytest tests/unit_tests -k "agents or mcp or deployment"` (390 passed), `ruff format --check`, `ruff check`, `mypy airbyte/mcp airbyte/agents airbyte/exceptions.py`
- MCP docs regenerated; generated tool list and schemas unchanged (the `functools.wraps` wrapper preserves signatures).
- Run via the repo venv; `poetry run` rejected the Hatch-based `pyproject.toml` on this machine.

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring the Agents API endpoint with `AIRBYTE_AGENTS_API_URL`.
  * Agents tools now provide clearer, actionable error messages, including guidance when unavailable or inputs are invalid.
  * Agent workspaces and organizations preserve configured API endpoints.

* **Behavior Changes**
  * Agents tools are hidden for non-public Cloud endpoints unless an Agents API endpoint is configured.
  * Blank endpoint overrides are treated as unset.
  * Agents API availability is validated before authentication.

* **Tests**
  * Expanded coverage for endpoint resolution, availability detection, error handling, and tool visibility.

Link to Devin session: https://app.devin.ai/sessions/9c07ef661caa4faf94a5a741f22d5f72
Open in Devin Desktop: https://app.devin.ai/desktop/session/9c07ef661caa4faf94a5a741f22d5f72?variant=devin
Requested by: @aaronsteers

> [!IMPORTANT]
> **Auto-merge enabled.**
> 


airbyte/mcp/_guards.py
```diff
@@ -47,3 +65,28 @@ def raise_if_untrusted_execution_context(feature: str) -> None:
...
+    ):
+        value = (get_mcp_config(config_source, key) or "").strip()
+        if value and value.rstrip("/") != default:
+            overridden[key] = value
+    return overridden
+
+
+def is_cloud_deployment(config_source: FastMCP | Context) -> bool:
+    """Return whether the MCP server targets public Airbyte Cloud."""
+    return not get_overridden_cloud_api_roots(config_source)
```
aaronsteers: I think we already have an authority on this question in the telemetry or another helper module. If so, I'd rather reuse existing logic or reuse existing helpers rather than maintain it also here. And the telemetry library should also be early loaded in terms of sequencing.

airbyte/mcp/agents.py
```diff
@@ -319,6 +323,7 @@ def _agents_access_message(error: AirbyteError) -> str | None:
 
 def _get_agent_organization(ctx: Context, organization_id: str | None) -> AgentOrganization:
     """Build an `AgentOrganization` from MCP config."""
+    raise_if_not_cloud_deployment(ctx, feature="Airbyte Agents organization tools")
```
aaronsteers: I don't think we need to raise an error. I think the priority is to hide the tool from being advertised. If we are going to raise an error, we should raise it in the agent's module, not the MCP module, which is just a presentation layer on top of the agent module. 

Again, the priority is to hide the tools when they're not going to work anyway, not to prevent the user from accessing any specific functionality, particularly

aaronsteers (commented):
Probably the best way to fail is when deciphering or calculating an API route for the agent's functionality. At least for now, I think that's the right place to fail, at least the runtime failure.

In another comment, I mentioned the priority of tool hiding being the first thing to focus on. In regard to failing at runtime with a specific error, I think we should have a helper, if we don't already, that calculates, returns, or retrieves the agent's base URL, similar to what we do for the config API URL. If we have not received an explicit agent root URL from the user and if we have not determined that the cloud API URL is our public offering, then we can safely fail because there won't be an agent API for that user to leverage.

aaronsteers:
> otherwise it raises `PyAirbyteInputError`.

Better to have a custom exception class like "agent APIs not available" or similar. See if there's any prior art for naming on this. Also, do some basic work to prevent a full stack trace in common scenarios where we might run into this. 

Or, if you can't do so cleanly, then talk to me about it in Slack about an approach. What I do not want you to do is have to touch every single MCP tool with a dumb `try`-`except` block. Except especially for those that already maybe don't have one. Better would be to build something into the MCP decorator that knows what kinds of errors (perhaps a few base classes of errors) it will print instead of raising or sharing the full stack trace.

airbyte/constants.py
```diff
@@ -346,6 +352,15 @@ def is_hosted_mcp_mode() -> bool:
 the include list.
 """
 
+MCP_CLOUD_ONLY_MODULES: frozenset[str] = frozenset({"agents"})
```
aaronsteers: I don't like the name. We use "Cloud" to equally in the code and API to refer to OSS versions and our hosted version. Also, I don't think it needs to be plural, at least right now. The simplest refactor is just to put this as a hidden constant somewhere as a singular MCP agent module name or something like that, probably with an underscore prefix. I don't know if it even needs to go in this module specifically or just in the module that's doing the hiding. 

By changing the question to "What is the agent's MCP module called?" we can avoid having to come up with a concise and clear-enough name for the hosted version of Cloud.

airbyte/_util/deployment.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    overridden: dict[str, str] = {}
+    for name, value, default in (
+        ("api_root", public_api_root, CLOUD_API_ROOT),
+        ("config_api_root", config_api_root, CLOUD_CONFIG_API_ROOT),
+    ):
+        text = value.strip().rstrip("/") if value else ""
+        if text and text != default:
+            overridden[name] = text
+    return overridden
```
aaronsteers: Don't we already have a helper for these? I feel like we already have a helper to get both of these API routes, and it looks like you're reimplementing from scratch.

airbyte/mcp/_guards.py
```diff
@@ -47,3 +61,11 @@ def raise_if_untrusted_execution_context(feature: str) -> None:
...
     if not is_trusted_execution_enabled():
         raise AirbyteTrustedExecutionRequiredError(feature=feature)
+
+
+def is_agents_api_available(config_source: FastMCP | Context) -> bool:
+    """Return whether the MCP server's deployment has an Agents API."""
+    return _is_agents_api_available(
+        public_api_root=get_mcp_config(config_source, MCP_CONFIG_API_URL),
+        config_api_root=get_mcp_config(config_source, MCP_CONFIG_CONFIG_API_URL),
+    )
```
aaronsteers: Just move this to airbyte/mcp/_tool_utils.py where it belongs.

aaronsteers:
@coderabbitai review

PR #1168, opened 2026-09-14
feat(agents): expose only `search` action, translate to `api_search` for the Agents API

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** The Agents MCP tools and Python client now use a single name, `search`, for a connector's native API search, and no longer accept the old name. Because the Agents API itself still expects the old name, the client quietly rewrites the action on the way out until the API catches up.

Specifically, `api_search` is removed from `AgentReadAction` and `AgentAction`, and `AgentConnector.execute()` maps `"search"` -> `"api_search"` in the request body via `_AGENTS_API_ACTION_NAMES`.

</td></tr></table>

Pairs with the Sonar rename that makes `search` canonical server-side:
- https://github.com/airbytehq/sonar/pull/6559

Tracking: https://linear.app/airbyteio/issue/AGENTIC-2234 (low priority, non-blocking).

Requested by AJ Steers (@aaronsteers). Round 1 of 2: this PR "fakes" the new name against today's API; round 2 removes the translation once the Sonar PR is deployed.

## Changes

- `airbyte/mcp/agents.py`: `AgentReadAction` / `AgentAction` literals drop `api_search`; `search` docstring clarifies it is the connector's native API search (parallel to `get`/`list`).
- `airbyte/agents/connectors.py`: `_AGENTS_API_ACTION_NAMES = {"search": "api_search"}` applied to the `action` field of the execute request body, with a `TODO` to delete it once the Agents API accepts `search`.
- Tests: `test_execute_request_body` gains a `search_sent_as_api_search` case; `test_convenience_methods` expects `search_entities()` to send `api_search`; `test_read_only_tool_action_type_excludes_writes` asserts `api_search` is absent from both literals.

## Review Spotlight

- [`connectors.py#L34-L40`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789423851-agents-search-action/airbyte/agents/connectors.py#L34-L40): the temporary name mapping and its removal note
- [`connectors.py#L226`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789423851-agents-search-action/airbyte/agents/connectors.py#L226): the single place the mapping is applied

## Risks

- No backward compatibility for `api_search` in the MCP tools or `AgentConnector.execute()`; these interfaces are brand new, so this was deemed acceptable.
- The translation lives in `AgentConnector.execute()` rather than only in the MCP layer so that `search_entities()` and direct `execute(..., "search")` callers also work against today's API.

## Follow-ups

- Round 2 (after https://github.com/airbytehq/sonar/pull/6559 deploys): delete `_AGENTS_API_ACTION_NAMES` and its use, and flip the two test expectations back to `search`.

## Test plan

- `uv run ruff format airbyte tests && uv run ruff check airbyte tests` -> `All checks passed!`
- `uv run pyrefly check` -> `0 errors`
- `uv run pytest tests/unit_tests/test_agents.py tests/unit_tests/test_mcp_agents.py -q` -> `136 passed`
- Not tested against a live Agents API workspace.

Link to Devin session: https://app.devin.ai/sessions/5fc4a9cada134715b08b0c01ee5cf30a
Open in Devin Desktop: https://app.devin.ai/desktop/session/5fc4a9cada134715b08b0c01ee5cf30a?variant=devin

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent search requests now use the correct API action format.
  * Search parameters are included correctly, with truncation skipped where required.
  * Unsupported action names, including legacy search names, are rejected with a clear validation error.
  * Available agent actions are now consistently validated and represented across agent interfaces.
  * Search convenience methods now produce correctly formatted requests.


airbyte/agents/connectors.py
```diff
@@ -213,9 +222,16 @@ def execute(  # noqa: PLR0913  # Explicit args are the point of this public API.
                 context={"entity_type": entity_type, "action": action},
             )
 
+        if action in _AGENTS_API_ACTION_NAMES.values():
+            raise PyAirbyteInputError(
+                message=f"The {action!r} action is not a PyAirbyte action name.",
```
aaronsteers:
```suggestion
                message=f"The {action!r} action is not a valid action name for `execute`.",
```

airbyte/agents/connectors.py
```diff
@@ -31,6 +31,15 @@
...
 """
 
+_AGENTS_API_ACTION_NAMES: dict[str, str] = {"search": "api_search"}
+"""PyAirbyte action names that the Agents API still knows by a different name.
+
+PyAirbyte exposes `search` for a connector's native API search. Until the Agents API
+adopts that name (it currently expects `api_search`), the request is rewritten on the way
+out, and the API-side name is rejected as input so callers standardize on `search` now.
+TODO: delete this mapping and its uses in `execute()` once the Agents API accepts `search`.
+"""
```
aaronsteers: I don't like this map approach. It built a system for something that we don't need a system for. Just have a StrEnum class and use the enum values inline.

airbyte/agents/connectors.py
```diff
@@ -213,9 +222,16 @@ def execute(  # noqa: PLR0913  # Explicit args are the point of this public API.
                 context={"entity_type": entity_type, "action": action},
             )
 
+        if action in _AGENTS_API_ACTION_NAMES.values():
+            raise PyAirbyteInputError(
+                message=f"The {action!r} action is not a PyAirbyte action name.",
+                guidance="Use `search` for a connector's native API search.",
```
aaronsteers: This is an overfit. You should just check if the value is in the allowed `enum`, which I think we're already doing, and just fail with a clear message if it's not in the `enum` or the set of string inputs that we expect, which, again, I think we already do. You don't need to overfit this just for one case.

airbyte/agents/connectors.py
```diff
@@ -31,6 +32,22 @@
...
+    `SEARCH` is the connector's native API search, parallel to `GET` and `LIST`.
+    """
+
+    LIST = "list"
+    GET = "get"
+    SEARCH = "search"
+    SQL_SELECT = "sql_select"
+    CREATE = "create"
+    UPDATE = "update"
+    DELETE = "delete"
```
aaronsteers, line 47:
```suggestion
    # Read-Only Actions (APIs):
    LIST = "list"
    GET = "get"
    SEARCH = "search"
    
    # Read-Write Actions (APIs):
    CREATE = "create"
    UPDATE = "update"
    DELETE = "delete"

    # Read-Only Queries (SQL Databases):
    SQL_SELECT = "sql_select"
```

airbyte/agents/connectors.py
```diff
@@ -31,6 +32,27 @@
 inside the transport layer.
 """
 
+
+class AgentAction(str, Enum):
+    """Connector actions accepted by `AgentConnector.execute()`.
+
+    `SEARCH` is the connector's native API search, parallel to `GET` and `LIST`.
+    """
```
aaronsteers:
```suggestion
    """Connector actions accepted by `AgentConnector.execute()`."""
```

aaronsteers:
Replaced by https://github.com/airbytehq/PyAirbyte/pull/1169

PR #1169, opened 2026-09-15
feat(agents): expose only `search` action and send it unchanged to the Agents API

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** The Agents MCP tools and Python client use a single name, `search`, for a connector's native API search, and send that name straight to the Agents API. This is the same change as the sibling PR, minus the temporary rename-on-the-wire, so it is the right one to merge once the server-side rename is deployed.

Specifically, actions are tiered as `AgentReadAction` / `AgentWriteAction` str-enums with `AgentAction = AgentReadAction | AgentWriteAction`; the MCP tools and `AgentConnector.execute()` type and validate against those, and the request body carries the action's value unchanged.

</td></tr></table>

Competes with #1168 (pick one):
- #1168

Depends on the Sonar rename being deployed:
- https://github.com/airbytehq/sonar/pull/6559

Tracking: https://linear.app/airbyteio/issue/AGENTIC-2234 (low priority, non-blocking).

Requested by AJ Steers (@aaronsteers). Branched from the head of #1168; deltas are dropping the `search -> api_search` translation and (per review) tiering the action enums.

## Changes

- `airbyte/agents/connectors.py`: `AgentReadAction(str, Enum)` = `list`/`get`/`search`/`sql_select`, `AgentWriteAction(str, Enum)` = `create`/`update`/`delete`, `AgentAction = AgentReadAction | AgentWriteAction`. `execute()` accepts enum members or plain strings, rejects anything outside the set (including legacy `api_search`) with `PyAirbyteInputError`, and sends the `.value` on the wire.
- `airbyte/mcp/agents.py`: the `Literal` action types are gone; tool `action` params are typed with `AgentReadAction` / `AgentAction` directly (pydantic renders them as string enums in the tool schema), and the read-only guard is an `AgentReadAction` membership check.
- Tests: request-body cases for `"search"` and `AgentReadAction.SEARCH` both expect `"action": "search"`; `legacy_api_search_action` asserts `api_search` is rejected; `test_read_only_tool_action_type_excludes_writes` pins both enums' members and that `api_search`/`download` are in neither.

## Review Spotlight

- [`connectors.py#L36-L63`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789431780-agents-search-action-canonical/airbyte/agents/connectors.py#L36-L63): the tiered enums
- [`connectors.py#L248-L265`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789431780-agents-search-action-canonical/airbyte/agents/connectors.py#L248-L265): allowlist check and the request body (no action rewrite)

## Risks

- Against a Sonar deployment that has not yet picked up https://github.com/airbytehq/sonar/pull/6559, `search` is rejected by the API with `'search' is not a valid Action`. Merge #1168 instead if that window matters.
- No backward compatibility for `api_search` in the MCP tools or `AgentConnector.execute()`; these interfaces are brand new.

## Test plan

- `uv run ruff format airbyte tests && uv run ruff check airbyte tests` -> `All checks passed!`
- `uv run pytest tests/unit_tests/test_agents.py tests/unit_tests/test_mcp_agents.py -q` -> `138 passed`
- Not tested against a live Agents API workspace.

Link to Devin session: https://app.devin.ai/sessions/5fc4a9cada134715b08b0c01ee5cf30a
Open in Devin Desktop: https://app.devin.ai/desktop/session/5fc4a9cada134715b08b0c01ee5cf30a?variant=devin


airbyte/mcp/agents.py
```diff
@@ -66,9 +69,7 @@
 stream rather than JSON, which PyAirbyte does not yet support.
 """
 
-AgentAction = Literal[
-    "list", "get", "search", "api_search", "sql_select", "create", "update", "delete"
-]
+AgentAction = Literal["list", "get", "search", "sql_select", "create", "update", "delete"]
```
aaronsteers: Shouldn't we remap these to the strenum?

airbyte/mcp/agents.py
```diff
@@ -55,9 +55,12 @@
 )
 
 
-AgentReadAction = Literal["list", "get", "search", "api_search", "sql_select"]
+AgentReadAction = Literal["list", "get", "search", "sql_select"]
```
aaronsteers: Ditto

airbyte/agents/connectors.py
```diff
@@ -31,6 +32,24 @@
 inside the transport layer.
 """
 
+
+class AgentAction(str, Enum):
```
aaronsteers: Is there a good way to have a two-tier or three-tier strenum structure? So one class can answer read-only and one can answer the full set, etc?

PR #1170, opened 2026-09-15
feat(mcp): label Airbyte Cloud API calls with X-Airbyte-Analytic-Source

girarda:
## Overview

<table><tr><td>

**:point_right: TL;DR:** PyAirbyte now tells the Airbyte platform where each Cloud API call came from (plain library, local MCP, or hosted MCP), so the platform's Segment events can be filtered by origin. No platform change is needed and the platform keeps owning user/org identity.

Specifically, every Airbyte Cloud request now carries the `X-Airbyte-Analytic-Source` header, which the platform's `TrackingClient` already copies onto Segment events as the `airbyte_source` property (the webapp sends `webapp`).

</td></tr></table>

## Changes

- `airbyte/_util/meta.py`: `AIRBYTE_ANALYTIC_SOURCE_HEADER` + `get_analytic_source()`:
  ```python
  not is_mcp_mode()      -> "pyairbyte"
  is_hosted_mcp_mode()   -> "pyairbyte-mcp-hosted"
  otherwise              -> "pyairbyte-mcp-local"
  ```
- `airbyte/_util/api_util.py`: header attached on all Cloud request paths
  - generated public API client: `AirbyteAPI(..., client=requests.Session())` with the header as a session default (Speakeasy merges session headers via `client.prepare_request`)
  - direct Config API calls (`_make_config_api_request`)
  - client-credentials exchange (`get_bearer_token` → `POST /applications/token`)
- Unit tests for value resolution (parametrized over the three modes) and for each request path.

The existing anonymous `mcp_tool` Segment event from the MCP middleware is unchanged. Agents API (`api.airbyte.ai`, Sonar) requests are intentionally out of scope.

## Review Spotlight

- Naming of the source values (`pyairbyte`, `pyairbyte-mcp-local`, `pyairbyte-mcp-hosted`) — easy to change if a different convention is preferred.

## Where the platform consumes the header

`SegmentTrackingClient` in `airbyte-platform-internal` (`oss/airbyte-analytics`):

- Header read — `getAirbyteSource()` pulls `X-Airbyte-Analytic-Source` off the current Micronaut request: https://github.com/airbytehq/airbyte-platform-internal/blob/444546280e/oss/airbyte-analytics/src/main/kotlin/io/airbyte/analytics/TrackingClient.kt#L189-L196
- Segment emit — `track()` sets `airbyte_source` (header or `unknown`) alongside customer id / deployment props and enqueues the `TrackMessage` with `userId = trackingIdentity.customerId`: https://github.com/airbytehq/airbyte-platform-internal/blob/444546280e/oss/airbyte-analytics/src/main/kotlin/io/airbyte/analytics/TrackingClient.kt#L154-L187
- Constants (`AIRBYTE_ANALYTIC_SOURCE_HEADER`, `AIRBYTE_SOURCE = "airbyte_source"`): https://github.com/airbytehq/airbyte-platform-internal/blob/444546280e/oss/airbyte-analytics/src/main/kotlin/io/airbyte/analytics/TrackingClient.kt#L199-L202
- Existing precedent — the webapp sends `X-Airbyte-Analytic-Source: webapp` in `oss/airbyte-webapp/src/core/api/apiCall.ts`.

## Test plan

- `uv run pytest tests/unit_tests/test_cloud_api_util.py tests/unit_tests/test_mcp_telemetry.py` — passed
- `uv run ruff format` / `ruff check` / `pyrefly check` on changed files — clean
- Live verification against a sandbox Cloud workspace, logging each outbound request via a `requests.Session.send` hook:

  Plain PyAirbyte (`CloudWorkspace.list_sources` / `list_connections`):
  ```
  POST /v1/applications/token -> 200 | X-Airbyte-Analytic-Source='pyairbyte'
  GET  /v1/sources            -> 200 | X-Airbyte-Analytic-Source='pyairbyte'
  GET  /v1/connections        -> 200 | X-Airbyte-Analytic-Source='pyairbyte'
  ```
  MCP server (`poe mcp-tool-test list_deployed_cloud_connections '{"limit": 2}'`):
  ```
  POST /v1/applications/token -> 200 | X-Airbyte-Analytic-Source='pyairbyte-mcp-local'
  GET  /v1/connections        -> 200 | X-Airbyte-Analytic-Source='pyairbyte-mcp-local'
  ```
  Tool returned 2 connections, `is_error=False`.
- Not tested live: hosted MCP mode (value covered by unit test only).

Requested by Alexandre (Slack).

## Summary by CodeRabbit

- **New Features**
  - API requests now include standardized Airbyte source metadata.
  - Requests automatically identify standard, local MCP, and hosted MCP execution modes.
  - SDK, bearer-token, and configuration API requests consistently include this metadata.

- **Tests**
  - Added coverage for source identification across standard, local MCP, and hosted MCP modes.
  - Added checks confirming metadata is included in SDK sessions, bearer-token requests, and configuration API requests.

Link to Devin session: https://app.devin.ai/sessions/922aba99e8554e3db49cf097c959b1dc
Open in Devin Desktop: https://app.devin.ai/desktop/session/922aba99e8554e3db49cf097c959b1dc?variant=devin
Requested by: @girarda


girarda:
@codex review

airbyte/_util/meta.py
```diff
@@ -57,6 +58,21 @@ def is_mcp_mode() -> bool:
...
+AIRBYTE_ANALYTIC_SOURCE_HEADER = "X-Airbyte-Analytic-Source"
+"""Request header the Airbyte platform stamps onto Segment events as `airbyte_source`."""
+
+
+def get_analytic_source() -> str:
+    """Return the `X-Airbyte-Analytic-Source` value describing this client.
+
+    The platform's tracking client records the value as the `airbyte_source` property on
+    every Segment event emitted while handling the request (the webapp sends `webapp`).
+    """
```
aaronsteers: Nit: The second block of help text here is TMI, imho. I think we can remove that.

And I would want to make sure we distinguish this telemetry as sent with Cloud Platform API requests, to distinguish between Segment telemetry sent with a similar name.

```suggestion
def get_cloud_api_analytic_source() -> str:
    """Return the `X-Airbyte-Analytic-Source` value describing this client when calling Cloud APIs."""
```

Will probably need to adjust for max length, but something like the above would be preferable I think.

aaronsteers (approved):
Looks good! I just have one nit with inline suggestion to help readability.

Devin, in case there's a place for this (and I don't know if there is) you may want to brief Alex (in slack, not here) with context on the existing analytics ID that is already calculated and used for Segment telemetry. That is a unique anonymous GUID or ULID which can be used to track number of unique calls over time. No action needed but just to make sure that context is known.

airbyte/_util/meta.py
aaronsteers: Proposing a bit more explainer here actually, just to say what this is/isn't and why it isn't affected by opt-out.

```suggestion
def get_analytic_source() -> str:
    """Return the `X-Airbyte-Analytic-Source` value describing this client.
    
    This is an identifier of the client software (MCP or PyAirbyte API) and
    *not* an indicator of the user and/or workspace. Because it is not
    user-identifying and only sent for logged-in API calls, it is not affected
    by the `DO_NOT_TRACK` environment variable.
    """
```

girarda:
@aaronsteers 
> existing analytics ID that is already calculated and used for Segment telemetry. That is a unique anonymous GUID or ULID which can be used to track number of unique calls over time
I found this, but we can store more context on Cloud, which might be useful

This approach will also allow us to track just one metric and partition by source to compare how the APIs are called

PR #1172, opened 2026-09-15
fix(agents): send workspace for `sql_select` and clarify pagination and `organization_id` docs

aaronsteers:
## Overview

<table><tr><td>
<b>TL;DR</b> — Client-side fixes from AJ Steers' Cloud MCP dogfooding: <code>sql_select</code> now sends the connector's <code>workspace_id</code> (it previously sent none, so the Agents API rejected every call) and the MCP tool docs stop implying that the top-level <code>cursor</code>/<code>page_size</code> paginate direct connector actions or that <code>organization_id</code> is optional for multi-org credentials.
</td></tr></table>

Requested by AJ Steers. Companion server-side fixes are in airbytehq/sonar#6570.

Pairs with #1174 (the `iter_entities()` pagination fail-loud change was split out of this PR at AJ's request; the real cursor mapping fix is moving server-side to Sonar):
- #1174

## Changes

1. **`sql_select` gets a workspace** (`airbyte/agents/connectors.py`, `workspaces.py`)
   - `AgentConnector` gains `workspace_id: str | None = None`; `AgentWorkspace.list_connectors()` / `get_connector()` populate it.
   - `AgentConnector.execute()` gains a top-level `workspace_id: str | None = None`. It resolves `workspace_id or self.workspace_id` and, for `action == sql_select` (the only action whose params the Agents API reads a workspace from), sets `params["workspace_id"]` unless `api_args` already carries a non-null `workspace_id`/`workspace_name` (a null selector key is dropped, and a caller's `workspace_name` is never combined with an injected `workspace_id` — the server 400s on both). Direct connector actions are scoped by the connector and their `params` stay untouched.
   - The MCP `execute_agent_connector*` tools forward their existing top-level `workspace_id` to `execute()` instead of using it only for the connector lookup.
2. **Pagination docs** (`airbyte/mcp/agents.py`, `airbyte/agents/models.py`): `cursor`/`page_size` are forwarded verbatim in `params`; only Context Store `search` (`limit`/`cursor`) and `sql_select` (`cursor`) read those names. Direct connector actions (e.g. GitHub `list`, which paginates via `after`) declare their own arguments in `api_args`. The tool descriptions and `end_cursor` docstrings now say so instead of "pass as `cursor`". No behavior change.
3. **Multi-org guidance** (`airbyte/mcp/agents.py`): `list_agent_workspaces` gets its own `LIST_WORKSPACES_ORGANIZATION_ID_TIP_TEXT` stating that `organization_id` is required when credentials belong to more than one organization (the API returns HTTP 400 otherwise) and pointing to `list_cloud_organizations` for discovery. The shared `ORGANIZATION_ID_TIP_TEXT` keeps the omit-when-derivable wording, since workspace-scoped tools resolve the organization from the workspace.

## Review Spotlight

- The `sql_select` injection lives in `AgentConnector.execute()` (core), not the MCP layer, so library callers get it too. A connector constructed directly without `workspace_id` behaves exactly as before.

## Tests

Two parametrized tests in `tests/unit_tests/test_agents.py`: `test_connector_workspace_injection` tabulates (action, api_args, top-level workspace_id) → exact `params` (connector default, explicit arg wins, explicit `api_args` selectors preserved, null `workspace_name` replaced, `list` untouched); `test_connectors_preserve_workspace_id` covers `get_connector()` / `list_connectors()` propagating `workspace_id`.

```bash
uv run ruff format --check . && uv run ruff check .        # All checks passed!
uv run mypy airbyte/agents/connectors.py airbyte/agents/models.py airbyte/agents/workspaces.py airbyte/mcp/agents.py
uv run pytest tests/unit_tests/test_agents.py tests/unit_tests/test_mcp_agents.py -q   # 156 passed
```

## Follow-ups / not in this PR

- Reported but not client-side: GitHub `search` returning 0 results for `created:>` date qualifiers, and `airbyteagentmcp` returning 401 `User not found` (deployment/auth configuration).

Link to Devin session: https://app.devin.ai/sessions/b6b9788bca1645cb9892a74e6e265910
Open in Devin Desktop: https://app.devin.ai/desktop/session/b6b9788bca1645cb9892a74e6e265910?variant=devin
Requested by: @aaronsteers


tests/unit_tests/test_agents.py
aaronsteers, line 1: It looks like you're not properly condensing your tests. You should have a smaller set of tests with a higher degree of parameterization. This makes it easier to read, and all we have to do is read the parameterized inputs and outputs to know if the test is valid and doing what it needs to do.

airbyte/agents/connectors.py
```diff
@@ -225,9 +229,12 @@ def execute(  # noqa: PLR0913  # Explicit args are the point of this public API.
...
         - `select_fields` and `exclude_fields` prune fields from returned entities.
-        - `page_size` and `cursor` are merged into `api_args` as pagination arguments, where
-          the connector receives `page_size` as its own `limit`. Pass the `end_cursor` of a
-          previous result as `cursor` to fetch the next page.
+        - `page_size` and `cursor` are merged into `api_args` as pagination arguments. Context
+          Store `search` uses `limit` and `cursor`, `sql_select` uses the top-level `cursor`, and
+          direct connector actions use their own pagination arguments in `api_args`. Pass the
+          result cursor according to the action type.
+        - `sql_select` is sent the connector's workspace as `workspace_id` unless `api_args`
+          already names a workspace.
```
aaronsteers: This comment makes no sense. `api_args` and `sql_select` should be mutually exclusive inputs. And the way we pass `workspace_id` should be global. We should do it the same everywhere.

It should be a top-level input arg. We can make it optional and infer in other ways and/or parse from API args if provided, but SQL select and API args are indeed mutually exclusive.

PR #1177, opened 2026-09-16
fix(mcp): surface Agents 403 `detail` and list SQL passthrough destinations

aaronsteers:
## 👨 Review Readiness 
- [ ] The human author (@aaronsteers) has reviewed this PR for readiness
- [x] The agent has attached evidence of the PR correctness to this PR ([e2e evidence comment](https://github.com/airbytehq/PyAirbyte/pull/1177#issuecomment-5708343310))

## Overview

<table><tr><td>

**:point_right: TL;DR:** When an agent asks the Cloud MCP to query a destination and the Agents API refuses, the agent now gets told *why* (typically "this destination exists but has not been enabled for Agents access") instead of a generic "no subscription / no workspace access" message. Destinations that can be queried with SQL are now also listed by the connector-listing tool, so an agent can find them without guessing IDs.

Specifically, `_agents_access_message` surfaces the 403 `detail` from `AirbyteError.context["response_text"]`, `list_agent_connectors` appends Snowflake/BigQuery destinations as `connector_kind="destination"` items with `supported_actions=["sql_select"]`, and the `execute_agent_connector_ro` help text points `sql_select` users at that listing and at `SHOW TABLES` / `DESCRIBE TABLE`.

</td></tr></table>

Requested by AJ Steers (`aaronsteers`) from the Fusion / cloud-mcp bug bash triage (findings 1 / 4 / 5: agent could not reach a Cloud Snowflake destination via `sql_select`).

## Root cause

- Sonar's `_check_external_actor_access` raises `HTTPException(403, detail="Actor is not enabled for Agents access.")` when a destination is not toggled on for Agents in Cloud Settings -> Context layer, and its error handler serializes that as `{"message": "...", "errors": [{"message": "..."}]}`, so the reason **does** reach the client.
- PyAirbyte's `_api_util` keeps the body in `AirbyteError.context["response_text"]`, but `_agents_access_message` ignored it and mapped **every** 403 to `AGENTS_FORBIDDEN_MESSAGE` (subscription / workspace access) - the misleading text the tester's agent saw.
- `list_agent_connectors` only returned Agents API sources, so nothing in the tool surface led the agent to the destination ID or to the fact that it needed enabling.

## Changes

- `_agents_access_message`: parse the 403 body via new `_agents_error_detail` (`message` -> `detail` -> `errors[0].message`); for `"Actor is not enabled for Agents access."` append `AGENTS_ENABLE_ACTOR_GUIDANCE` (exists in workspace, org admin must enable under Settings -> Context layer, do not retry); other details become `"The Airbyte Agents API denied access: <detail>"`; no/invalid body keeps the original `AGENTS_FORBIDDEN_MESSAGE` fallback.
- `AgentConnectorResult`: new optional fields `connector_kind`, `supported_actions`, `sql_dialect`, `note` (defaults preserve the existing shape for sources).
- `list_agent_connectors`: after the Agents API sources, append Cloud destinations whose `definition_id` is in `SQL_PASSTHROUGH_DESTINATION_DEFINITION_IDS` via `_get_cloud_workspace(...).list_destinations()`, each marked `sql_select`-only with its `sql_dialect` and a note on enabling access. The Cloud/Agents APIs do not expose the enabled state, so the note says where to enable it rather than reporting a flag.
- `execute_agent_connector_ro`: `connector_id` / `action` descriptions and docstring now say destination IDs come from `list_agent_connectors` and that `SHOW TABLES` / `DESCRIBE TABLE <name>` discover tables and columns.
- Tests: parametrized 403-with-detail / 403-without-detail cases, an end-to-end `sql_select` access-denied case, destination listing (supported vs. unsupported definition IDs), and the early-return path when the Agents listing itself is denied.

## Review Spotlight

- [`agents.py` `_agents_access_message` / `_agents_error_detail`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789599852-agents-destination-403/airbyte/mcp/agents.py#L344-L388): the 403 detail extraction and guidance
- [`agents.py` `list_agent_connectors`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789599852-agents-destination-403/airbyte/mcp/agents.py#L711-L765): destination discovery

## Risks

- `list_agent_connectors` now also calls the Cloud API. If that call fails after the Agents listing succeeded, the error propagates (the tool no longer returns a partial source-only list). If the Agents listing is denied, destinations are not looked up.
- Destinations are appended only when their definition is Snowflake or BigQuery; other destination types remain unlisted, as before.

## Follow-ups

- Whether the tester's destination was actually left disabled in prod (and whether destinations created after org enrollment are auto-enabled) is a platform question for `airbyte-platform-internal` / Sonar onboarding, not addressed here.

## Test plan

- `uv run pytest tests/unit_tests/test_mcp_agents.py` - 81 passed
- `uv run ruff format --check .` / `uv run ruff check .` / `uv run pyrefly check airbyte/mcp/agents.py` - clean (the repo has moved from Poetry/mypy to uv/pyrefly, so the `poetry run mypy .` from the request is not applicable)
- Manual repro (needs Cloud creds for a workspace with a Snowflake/BigQuery destination): `poe mcp-tool-test list_agent_connectors '{"workspace_id": "<ws>"}'` should list the destination with `connector_kind: destination`; `poe mcp-tool-test execute_agent_connector_ro '{"connector_id": "<dest>", "entity_type": "x", "action": "sql_select", "api_args": {"sql": "SHOW TABLES", "sql_dialect": "snowflake"}}'` on a not-yet-enabled destination should return `status: access_denied` with the "Actor is not enabled for Agents access. ... Settings -> Context layer ..." message. Not run in this session (no external-org Cloud credentials available).


airbyte/mcp/agents.py
```diff
@@ -766,7 +868,9 @@ def execute_agent_connector_ro(  # noqa: PLR0913  # Explicit args are the point
                 "The `search` action is the connector's native API search, parallel to `get` "
                 "and `list`. "
                 "For `sql_select`, pass `sql` and `sql_dialect` (snowflake, bigquery, athena, "
-                "trino) in `api_args` and any value for `entity_type`. The `download` action "
+                "trino) in `api_args` and any value for `entity_type`; the `connector_id` is a "
```
aaronsteers: Only snowflake and bigquery are publicly supported advertised. Athena/trino should be internal-only and not advertised. It is fine if the tool accepts them, but don't promote them in docs/help text.

PR #1182, opened 2026-09-17
feat(mcp): honor allowlisted upstream `X-Airbyte-Analytic-Source` in MCP mode

girarda:
## Overview

<table><tr><td>

**:point_right: TL;DR:** Lets a trusted internal caller (the Cloud Support Agent) tell the hosted MCP server which analytics source to report, so support-agent chats stop showing up as customers using external MCP clients.

Specifically, `get_cloud_api_analytic_source()` now returns an allowlisted inbound `X-Airbyte-Analytic-Source` value (`_UPSTREAM_ANALYTIC_SOURCES = {"coral-support-agent"}`) before falling back to the existing mode-based values.

</td></tr></table>

Pairs with:
- https://github.com/airbytehq/coral-agents/pull/159
- https://github.com/airbytehq/airbyte-warehouse/pull/3478

Requested by Alexandre Girard after Demonware appeared in the MCP dashboard with 18 "hosted MCP" calls that were actually in-app Support Agent lookups.

## Changes

- `airbyte/_util/meta.py`: `get_cloud_api_analytic_source()`, in MCP mode, returns an allowlisted inbound header value before the mode-based fallback; plain PyAirbyte never reads headers:
  ```python
  if not is_mcp_mode():
      return "pyairbyte"
  upstream = get_http_headers().get("x-airbyte-analytic-source", "").strip().lower()
  if upstream in _UPSTREAM_ANALYTIC_SOURCES:
      return upstream
  return "pyairbyte-mcp-hosted" if is_hosted_mcp_mode() else "pyairbyte-mcp-local"
  ```
- `airbyte/_util/meta.py`: small `get_http_headers()` wrapper around `fastmcp.server.dependencies.get_http_headers` with a lazy import, so `import airbyte` does not pull in `fastmcp` (about 1s of import time) for non-MCP users.
- `tests/unit_tests/test_cloud_api_util.py`: parametrized cases for allowlisted header (hosted and local MCP), header ignored outside MCP mode, unlisted header, and missing header.

## Review Spotlight

- [`meta.py` `get_cloud_api_analytic_source`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789688819-upstream-analytic-source/airbyte/_util/meta.py#L80-L93): precedence logic

## Risks

- The header is an analytics label only and never affects authorization (every downstream Cloud API call still uses the caller's own bearer token). An external client could mislabel its own traffic as `coral-support-agent`; the allowlist just keeps arbitrary strings out of the `airbyte_source` domain. Keeping the allowlist was confirmed by the maintainer (https://github.com/airbytehq/PyAirbyte/pull/1182#discussion_r4042433251).
- `get_http_headers()` returns `{}` outside an HTTP request, and it is only consulted in MCP mode, so plain PyAirbyte behavior is unchanged. Existing warehouse rows are not reclassified.

## Test plan

- `uv run pytest tests/unit_tests/test_cloud_api_util.py -q` → passed.
- `uv run ruff check airbyte tests`, `uv run ruff format --check airbyte tests`, `uv run pyrefly check airbyte/_util/meta.py` → clean.
- `uv run python -c "import airbyte"` confirms `fastmcp` is not loaded at import time.

Link to Devin session: https://app.devin.ai/sessions/a61b7da4e42746f49fb10924f545c05e
Open in Devin Desktop: https://app.devin.ai/desktop/session/a61b7da4e42746f49fb10924f545c05e?variant=devin
Requested by: @girarda


airbyte/_util/meta.py
```diff
@@ -61,18 +62,27 @@ def is_mcp_mode() -> bool:
 AIRBYTE_ANALYTIC_SOURCE_HEADER = "X-Airbyte-Analytic-Source"
 """Request header the Airbyte platform stamps onto Segment events as `airbyte_source`."""
 
+_UPSTREAM_ANALYTIC_SOURCES: frozenset[str] = frozenset({"coral-support-agent"})
```
girarda, line 64: a malicious user can technically still spoof the value by setting it to coral-support-agent. I think the risk is low.

I would be OK passing through any value here if we don't like the extra complexity.

girarda:
@codex review

PR #1184, opened 2026-09-18
ci: run integration tests once in a sharded job instead of per matrix entry

girarda:
## Overview

<table><tr><td>

**:point_right: TL;DR:** The slow integration tests currently run three times on every PR (once per Ubuntu Python version, plus again in the fast job for the non-slow ones), making the longest CI job ~28 minutes. This runs them once, split into four parallel shards, leaves the OS/Python matrix to unit tests only, and drops the now-redundant fast job.

Specifically, `python_pytest.yml` renames the `pytest` matrix job to `pytest-unit` (now `--exitfirst`), adds a sharded `pytest-integration` job plus a `pytest-coverage-combined` job, and deletes `pytest-fast`. `pytest-no-creds` is unchanged.

</td></tr></table>

Requested by Alexandre Girard in Slack (#ask-devin-ai).

## Changes

- `pytest` -> `pytest-unit`: same `{3.10, 3.11} x {Ubuntu, Windows}` matrix, but every entry now runs `tests/unit_tests/ tests/docs_tests/` (Windows already did; Ubuntu no longer runs the whole tree). Publishes `PyTest Results (Unit)` / `code-coverage/pytest-unit` from 3.10+Ubuntu.
- New `pytest-integration` (Ubuntu, Python 3.10, non-fork only) with a 4-way `matrix.include` over `shard` / `pytest-args`:
  - `cache-types-cloud`: `test_all_cache_types.py -k "BigQuery or Snowflake"`
  - `cache-types-local`: `test_all_cache_types.py -k "not BigQuery and not Snowflake"`
  - `cloud`: `tests/integration_tests/cloud/`
  - `other`: everything else under `tests/integration_tests/`
  Each shard publishes `PyTest Results (Integration, <shard>)`, uploads its raw `.coverage` data as a `coverage-data-*` artifact, and prints `--durations=10`.
- New `pytest-coverage-combined` job (`needs: [pytest-unit, pytest-integration]`, `if: always()`): downloads the `coverage-data-*` artifacts from unit (3.10/Ubuntu) and the four shards, runs `coverage combine`, and uploads the result under the original `code-coverage/pytest` label plus a `combined-test-coverage` html artifact. `code-coverage/pytest-unit` is the unit-only data point.
- `pytest-fast` is deleted: it overlapped with `pytest-unit` (same unit/docs set, single 3.10/Ubuntu) and its fork-safe role is already covered by `pytest-no-creds`, which runs the whole tree with credentials blanked so missing `requires_creds` markers are caught. `pytest-no-creds` is intentionally left untouched (narrowing it would defeat that check).
- `pytest-unit` gains `--exitfirst` to keep the fail-fast signal.

## Expected effect

Based on the last 50 runs: Ubuntu full jobs took 21–37 min (median ~26), Windows 3–6 min. In a representative run, `test_all_cache_types.py` alone was ~10 min (BigQuery ~4, Postgres ~3.5), `cloud/` ~4.5 min, remaining integration tests ~6 min, unit tests ~5 min. Longest job should drop to roughly 6–8 min, and total runner minutes per PR by ~25–30.

## Risks

- Integration tests no longer run on Python 3.11 (unit tests still do). A `workflow_dispatch`/nightly full run could be added later if wanted.
- Job names changed: any required-status-check rules referencing `Pytest (All, Python 3.10, Ubuntu)` etc. need updating to the new names.
- `PyTest Results (Full)` check is replaced by `PyTest Results (Unit)` plus four `PyTest Results (Integration, *)` checks.
- `coverage combine` relies on all contributing jobs checking out to the same path (all are `ubuntu-latest`), so no path remapping is configured.

## Test plan

- `actionlint .github/workflows/python_pytest.yml` (only pre-existing warning: unknown `code-quality` permission scope, already present on `main`).
- Verified locally with `pytest --collect-only` that the four shards collect exactly the same 199 tests as `tests/integration_tests/` with the same `-m` filter (disjoint, no gaps).
- Verified locally that `coverage combine --keep <dir>/*/.coverage` merges data files from separate directories.
- CI on this PR exercises the new jobs directly; check durations of each shard and the `code-coverage/pytest` combined number.

Link to Devin session: https://app.devin.ai/sessions/b99349cdbe494167ada12159248bfebe
Open in Devin Desktop: https://app.devin.ai/desktop/session/b99349cdbe494167ada12159248bfebe?variant=devin
Requested by: @girarda


.github/workflows/python_pytest.yml
```diff
@@ -131,6 +130,7 @@ jobs:
       run: >
         uv run coverage run -m pytest
         --durations=5 --exitfirst
+        tests/unit_tests/ tests/docs_tests/
```
aaronsteers: Split so one line for each pls

.github/workflows/python_pytest.yml
```diff
@@ -212,25 +212,108 @@ jobs:
...
-    - name: Run Pytest [all]
-      if: matrix.os != 'Windows'
+    - name: Run Pytest [unit tests]
       timeout-minutes: 60
       env:
         GCP_GSM_CREDENTIALS: ${{ secrets.GCP_GSM_CREDENTIALS }}
       run: >
         uv run coverage run -m pytest
         --verbose
+        tests/unit_tests/ tests/docs_tests/
```
aaronsteers: Ditto

.github/workflows/python_pytest.yml
```diff
@@ -212,25 +212,108 @@ jobs:
...
+      with:
+        ref: ${{ github.event.pull_request.head.sha || github.sha }}
+    - name: Install uv
+      uses: astral-sh/setup-uv@f06b870e0a91d23284a3013acc55e6f88ab4b904 # v7
+      with:
+        version: "latest"
+        python-version: '3.10'
+    - name: Install dev dependencies
+      # We only need helper libraries
+      run: uv sync --group dev
```
aaronsteers, line 239: I think you want --all-extras

.github/workflows/python_pytest.yml
```diff
@@ -212,25 +212,108 @@ jobs:
...
         uv run coverage run -m pytest
         --verbose
+        tests/unit_tests/ tests/docs_tests/
         -m "not linting and not super_slow and not flaky"
 
-    - name: Run Pytest [unit tests only]
-      if: matrix.os == 'Windows'
+    - name: Print Coverage Report
+      if: always()
+      run: uv run coverage report
```
aaronsteers, line 152: How should we think about coverage with the new splits? Do we still appropriately baseline across all tests in regards to total coverage? (Answer inline, not in slack.)

aaronsteers (commented):
Mostly nits. Devin, answer inline please

.github/workflows/python_pytest.yml
aaronsteers: Yes that approach sounds correct. If each job uploads its artifacts and then a final job combines and uploads them as the final artifact, then I think that meets our requirements. The only additional one that I care about is the fast or unit coverage version of it. I want to know what percentage of tests are passed by unit and/or the fast set. I want to know, as a second dataset or second data point, the percentage of coverage for the combined set of all tests.

aaronsteers: Do we really need pytest-fast as well as pytest-unit? I think they are too similar to warrant both existing. Whichever wins should fast-fail and be the coverage signal for that use case.

.github/workflows/python_pytest.yml
```diff
@@ -247,35 +263,90 @@ jobs:
...
+      run: uv sync --group dev
+
+    - name: Download coverage data
+      uses: actions/download-artifact@018cc2cf5baa6db3ef3c5f8a56943fffe632ef53 # v6
+      with:
+        pattern: coverage-data-*
+        path: coverage-data/
+
+    - name: Combine coverage data
+      run: uv run coverage combine --keep coverage-data/*/.coverage
```
aaronsteers: Devin, can you rewrite this to do a standalone `coverage` cli install using `uv`?

1. Drop the clone step (and later uv sync)
2. Keep the setup-uv step
3. Standalone step for something like `uv tool install coverage`

If it works, great. We can skip clone/checkout and just do the combine here. If not, that's fine. Not a blocker.

aaronsteers: Revert back. If we have to clone anyway then it's not worth the separately pinned coverage library version.

PR #1171, opened 2026-09-15
feat(mcp): route agent tools through Cloud APIs

pkozikow:
Related documentation work: [AGENTIC-1719](https://linear.app/airbyteio/issue/AGENTIC-1719/add-backend-api-and-mcp-support-for-connector-source-skill-docs).

## Summary

Route the experimental `agents` MCP tools through Cloud resource discovery, execution, and connector-docs APIs. Supported tool names remain stable; source reads use get/list/search and destination SQL is restricted to the Snowflake contract.

- Share a request-scoped Cloud token across discovery and execution, verify actor workspace ownership, and sanitize errors. Execution/docs/token calls have bounded reads and no redirects or automatic retries.
- Preserve native response data and metadata, including Snowflake positional rows and truncation. Legacy pagination fields remain unknown when no verified mapping exists.
- Make both execution tools read-only, remove unsupported skill list/search tools and Context Store readiness, and retain insiders controls. Standalone `airbyte.agents` and Cloud CRUD tools remain unchanged.

## Validation

- `uv run --frozen poe check` — passed.
- `uv run --frozen poe test-unit-tests` — 951 passed, 1 existing skip before review fixes.
- After UUID/proxy review fixes: `uv run --frozen pytest tests/unit_tests/test_mcp_agents.py tests/unit_tests/test_mcp_cloud_execution.py -q` — 144 passed; final lint/typing/collection gate passed.
- Focused transport tests use real local HTTP fixtures for redirects, disconnects, timeouts, malformed responses, limits, token reuse, and response cleanup.
- MCP tests use the real Cloud SDK with HTTP fixtures and registered FastMCP stdio/hosted sessions for workspace isolation, read-only enforcement, and schema behavior.
- `uv run --frozen poe mcp-docs-md` — six Agents tools; repeated generation is stable.

## Release prerequisites

Keep this PR unreleased until the Cloud docs/destination routes and paired Sonar support are merged, deployed, and verified with supported bearer/application credentials. Live Cloud execution was not tested. Required PRs remain open: airbytehq/airbyte-platform-internal#19464, airbytehq/airbyte-platform-internal#19472, airbytehq/sonar#6542, airbytehq/sonar#6544. A merged configuration change alone does not establish deployment availability.

## Summary by CodeRabbit

* **New Features**
  * Added Cloud-based execution for agent source and destination connectors.
  * Added access to connector skill documentation.
  * Added workspace scoping, authorization checks, and safer handling of execution requests and responses.
[...]

PR #1187, opened 2026-09-18
feat(mcp): return inline docs summary from `inspect_agent_connector`

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** Inspecting a connector now returns its usage docs summary in the same response, so agents no longer have to make (and usually skip) a second docs call before guessing parameter names.

Specifically, `inspect_agent_connector` returns `docs: AgentSkillDocsResult | None` (the default, no-section read of `docs_skill_id`: execution guidance plus one outline entry per action) with a `docs.guidance` string telling the agent to call `read_agent_skill_docs(skill_id=..., section=<a section_id from docs.outline>)` for a single action's full parameter detail.

</td></tr></table>

Requested by AJ (`aaronsteers`). Standalone change; intentionally not built on #1171.

## Changes

- `AgentConnectorDetailsResult` gains `docs` and drops `docs_skill_id` (the id is `docs.skill_id`; the class is moved below `AgentSkillDocsResult` so the annotation resolves); `AgentSkillDocsResult` gains `guidance: str | None`, set only on the inline summary. The example section in the guidance is the first `available` outline entry; with no available sections the guidance says so instead of pointing at a placeholder.
- `inspect_agent_connector`: after a successful inspect, reads `workspace.read_skill_docs(details.docs_skill_id)` in its own `try`; an `AirbyteError` or `requests.RequestException` (timeout/connection error — `make_agents_api_request` doesn't wrap transport errors) there does not fail inspect — `docs` is returned with `skill_id` set, empty `outline`/`content`, and the failure in `docs.message`, and `"Connector docs are unavailable: <message>"` is appended to `warnings`. No `docs_skill_id` → no docs read, no warning.
- SQL passthrough destination fallback (`_inspect_destination_fallback`) inlines the built-in destination docs the same way.
- `_skill_docs_result()` extracted from the two identical `AgentSkillDocs → AgentSkillDocsResult` mappings (`read_agent_skill_docs`, `_destination_skill_docs_fallback`) now that there is a third caller.
- Tool descriptions updated: inspect says the summary is inline; `read_agent_skill_docs.skill_id` says it is only needed for a single section's detail.

## Review Spotlight

- [`agents.py` inspect docs read](https://github.com/airbytehq/PyAirbyte/blob/devin/1789711409-inspect-inline-docs/airbyte/mcp/agents.py#L997-L1015): fail-open docs read and warning
- [`_docs_guidance`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789711409-inspect-inline-docs/airbyte/mcp/agents.py#L563-L568): the hint text agents will read

## Risks

- Inspect now costs one extra Agents API call per invocation and the payload grows: live GitHub source inspect is ~48k chars (≈45 actions in the outline + guidance blocks); a SQL passthrough destination is ~5k. Open to trimming (e.g. dropping `None`-valued keys from `content` blocks) if reviewers prefer.

## Test plan

- `uv run pytest tests/unit_tests/test_mcp_agents.py tests/unit_tests/test_agents.py -q` → 200 passed. New tests cover: docs inlined on success (and `read_skill_docs` called once with the skill id, no section), docs read failure (`AirbyteError` and `requests.Timeout`) → warning + `docs.skill_id`/`docs.message` populated, no `docs_skill_id` → docs read never runs, empty outline → guidance without example, destination fallback inlines built-in docs.
- `ruff check` / `ruff format --check` / `pyrefly check` on touched files → clean.
- Live smoke via `poe mcp-tool-test inspect_agent_connector` against a Cloud GitHub source and a BigQuery destination (insiders MCP): `docs.outline` and `docs.guidance` present in both.

Link to Devin session: https://app.devin.ai/sessions/a2754b0014d241be8856d2ca793fa5e3
Open in Devin Desktop: https://app.devin.ai/desktop/session/a2754b0014d241be8856d2ca793fa5e3?variant=devin
Requested by: @aaronsteers


airbyte/mcp/agents.py
```diff
@@ -342,6 +322,46 @@ class AgentSkillDocsResult(BaseModel):
     """Why the docs are empty, when the Agents API denied the request."""
 
 
+class AgentConnectorDetailsResult(BaseModel):
```
aaronsteers, line 328: Was it necessary to change the order of this class declaration? You made the diff noisier by moving it below agent skill result.

aaronsteers: On the other hand, I think I prefer the unquoted types, so you can swap the order back to avoid needing the quotations.

airbyte/mcp/agents.py
```diff
@@ -342,6 +322,46 @@ class AgentSkillDocsResult(BaseModel):
...
+    `execute_agent_connector` without appearing here.
+    """
+
+    docs: AgentSkillDocsResult | None = None
+    """The connector's usage docs summary: execution guidance plus the outline of per-action
+    sections. Each `outline[].section_id` can be passed as `section` to `read_agent_skill_docs`
+    for that action's full parameter list, types, and examples."""
+
+    docs_guidance: str | None = None
+    """How to get more detail than the inline docs summary."""
```
aaronsteers: Any reason why we shouldn't include the docs guidance with the docs element?

airbyte/mcp/agents.py
```diff
@@ -335,13 +315,53 @@ class AgentSkillDocsResult(BaseModel):
...
+    """Display name of the connector."""
+
+    workspace_id: str | None = None
+    """The workspace that owns the connector."""
+
+    source_definition_name: str | None = None
+    """The name of the underlying source definition, for example `GitHub`."""
+
+    docs_skill_id: str | None = None
+    """Skill ID for this connector's usage docs, when reported by the Agents API."""
```
aaronsteers: Seems like this also should just be an element inside `docs`

airbyte/mcp/agents.py
```diff
@@ -335,13 +315,50 @@ class AgentSkillDocsResult(BaseModel):
...
+    docs: AgentSkillDocsResult | None = None
+    """The connector's usage docs summary: execution guidance plus the outline of per-action
+    sections. Each `outline[].section_id` can be passed as `section` to `read_agent_skill_docs`
+    for that action's full parameter list, types, and examples. See `docs.guidance`."""
+
+    warnings: list[str]
+    """Warnings the Agents API reported about this connector."""
+
+    message: str | None = None
+    """Why the details are empty, when the Agents API denied the request."""
```
aaronsteers, line 361: Both of these docstrings are overfit and/or too specific. Just say what they'll contain generally, and keep it evergreen rather than projecting what you think it'll have.

PR #1188, opened 2026-09-18
feat(mcp): add `agents_enabled` filter and status to `list_cloud_workspaces`

aaronsteers:
## Summary

First step in merging the Agent MCP tools with their Cloud corollaries (requested by @aaronsteers). `list_agent_workspaces` is left untouched; `list_cloud_workspaces` gains parity with it:

- New input `agents_enabled_only: bool = False` — return only workspaces enabled for Airbyte Agents (Fusion / "Context layer").
- New result field `CloudWorkspaceResult.agents_enabled: bool | None` on every returned workspace.

How enablement is determined: the Agents API `GET /workspaces` (the same call `list_agent_workspaces` already makes via `AgentOrganization.list_workspaces()`) returns only Fusion-enabled workspaces. For each distinct `organization_id` in the Cloud result set we call it once and cross-reference IDs:

```
agents_enabled = workspace_id in agents_api_workspace_ids[org]   # True / False
None  -> workspace has no org id, Agents API unreachable (AirbyteAgentsUnavailableError),
         or Agents API returned a non-403 error
403   -> treated as "org has no Agents access" => all its workspaces False
```

`agents_enabled_only=True` keeps only `agents_enabled is True` (unknown/`None` is excluded), and applies `limit` after filtering so the caller still gets up to `limit` enabled workspaces. A dedicated empty-result message points users at enabling the Context layer or dropping the filter.

Cost: one extra Agents API request per distinct organization in the result set; default behavior otherwise unchanged.

## Testing

```
ruff format --check airbyte/mcp/cloud.py tests/unit_tests/test_cloud_credentials.py
ruff check airbyte/mcp/cloud.py tests/unit_tests/test_cloud_credentials.py
pytest tests/unit_tests/test_cloud_credentials.py -q   # 74 passed
```

`mypy` is not installed in the local venv and `poetry run` fails on this checkout's config, so type checking is deferred to CI.

Link to Devin session: https://app.devin.ai/sessions/a0dfe2d2cab84b2aa7e99654f7f3489d
Open in Devin Desktop: https://app.devin.ai/desktop/session/a0dfe2d2cab84b2aa7e99654f7f3489d?variant=devin


aaronsteers:
We can't ship this. 

1. "Agents enabled" doesn't work in the world where everything is used by agents, even the EL pipelines. 
2. Workspace is not the granularity to start this at, unfortunately. Probably need to start at the connector level. 

I'm going to close this PR. Devin, let's do a spike instead at the connector level. Ask me to approve the proposed names and public signature changes before you create the PR.

PR #1189, opened 2026-09-18
feat(cloud): add `list_connectors()` and `external_access_enabled` / `search_indexing_enabled` flags on org, workspace, connector

aaronsteers:
## Overview

<table><tr><td>

**:point_right: TL;DR:** Cloud users can now see, in Python and through the Cloud MCP tools, whether an organization, a source, or a destination is enabled for AI agents and whether Airbyte indexes its data for fast search, and filter lists down to just those. No tools are removed or renamed.

Specifically, this adds `CloudWorkspace.list_connectors(connector_type, with_feature, name_contains, limit) -> list[CloudConnector]`, the `ConnectorType` / `ConnectorFeature` enums in `airbyte.cloud.connectors`, and `external_access_enabled` / `search_indexing_enabled: bool` properties on `CloudOrganization`, `CloudWorkspace`, and `CloudConnector`. The MCP tools `list_deployed_cloud_source_connectors`, `list_deployed_cloud_destination_connectors`, and `list_cloud_organizations` gain `with_feature` and surface both flags.

</td></tr></table>

Requested by AJ Steers (aaronsteers). Supersedes https://github.com/airbytehq/PyAirbyte/pull/1188.

## Changes

```python
class ConnectorType(str, Enum):      SOURCE = "source"; DESTINATION = "destination"
class ConnectorFeature(str, Enum):   EXTERNAL_ACCESS = "external_access"; SEARCH_INDEXING = "search_indexing"

CloudWorkspace.list_connectors(*, connector_type=None, with_feature=None, name_contains=None, limit=None) -> list[CloudConnector]
CloudClient.list_organizations(*, name_contains=None, with_feature=None, limit=None)

CloudOrganization / CloudWorkspace / CloudConnector:
    external_access_enabled: bool
    search_indexing_enabled: bool
```

- `list_connectors()` returns mixed `CloudSource` / `CloudDestination` items, does one Context layer `list_connectors` + N `inspect` for the workspace, and seeds each item's feature cache so `with_feature` filtering and later property reads make no further calls. `list_sources()` / `list_destinations()` are unchanged typed conveniences.
- Incompatible API root (self-managed, no `AIRBYTE_AGENTS_API_URL`): every flag is `False` from configuration alone, zero network calls. On a compatible root, 403/404 from the Context layer means "not enabled"; any other failure raises from the core property (no more `bool | None` tri-state).
- Source `external_access_enabled` = present in the Context layer connector listing; `search_indexing_enabled` = `context_store_readiness.configured_cache_entities` non-empty. Destination `external_access_enabled` = workspace enabled **and** definition in `SQL_PASSTHROUGH_DESTINATION_DEFINITION_IDS`; `search_indexing_enabled` is always `False`.
- Organization / workspace `search_indexing_enabled` currently aliases `external_access_enabled` until an org-level indexing signal ships; this is documented on the properties.
- `search_enabled` renamed to `search_indexing_enabled` (`ConnectorFeature.SEARCH` → `SEARCH_INDEXING`) so it is not confused with a connector's native or passthrough search.
- MCP `airbyte/mcp/cloud.py` is now a pass-through: no Agents API calls or try/except there; `with_feature`, `name_contains`, and `limit` are forwarded to core. Rule documented in `CONTRIBUTING.md`.
- `airbyte.agents._api_util` imports `airbyte._util.deployment` as a module to break an import cycle exposed by the Cloud core now depending on Agents utilities.

## Review Spotlight

- [`CloudWorkspace.list_connectors`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789723467-cloud-agents-feature-parity/airbyte/cloud/workspaces.py#L940): batching + filter order
- [`CloudWorkspace._get_connector_features` / `external_access_enabled`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789723467-cloud-agents-feature-parity/airbyte/cloud/workspaces.py#L351-L470): incompatible-root short circuit, 403/404 handling, destination semantics
- [`CloudOrganization.external_access_enabled`](https://github.com/airbytehq/PyAirbyte/blob/devin/1789723467-cloud-agents-feature-parity/airbyte/cloud/organizations.py#L158): org-level flags
- [`CONTRIBUTING.md` — MCP presentation-layer rule](https://github.com/airbytehq/PyAirbyte/blob/devin/1789723467-cloud-agents-feature-parity/CONTRIBUTING.md#mcp-tools-are-a-presentation-layer)

## Risks / Notes

- Cost: one Context layer call per workspace/organization plus one `inspect` per externally accessible source when listing sources. Same as before; still fine for interactive use.
- Behavior change vs. the earlier revision of this PR: runtime failures on a compatible root now raise instead of returning `None`.

## Test plan

```
ruff format --check . && ruff check .
pytest tests/unit_tests
```

979 passed, 1 skipped locally. New/updated tests in `test_cloud_credentials.py`, `test_mcp_cloud.py`, `test_cloud_client.py` cover: mixed `list_connectors()`, `connector_type` / `with_feature` / `name_contains` / `limit` (post-filter), `limit <= 0`, zero-call behavior on incompatible roots, one shared lookup per list + lazy per-connector caching, destination semantics, org-level flags (enabled / 403 / 404 / 500 / transport), `list_organizations(with_feature=...)`, MCP result fields, malformed payloads, and import-cycle safety. MyPy not run locally (not installed in the env); relying on CI.

## Follow-ups

- `with_docs: bool` on `describe_cloud_source` / `describe_cloud_destination` and a `get_connector_docs` tool — design under discussion.

Link to Devin session: https://app.devin.ai/sessions/a0dfe2d2cab84b2aa7e99654f7f3489d
Open in Devin Desktop: https://app.devin.ai/desktop/session/a0dfe2d2cab84b2aa7e99654f7f3489d?variant=devin
Requested by: @aaronsteers


airbyte/mcp/cloud.py
```diff
@@ -247,6 +282,12 @@ class CloudWorkspaceResult(BaseModel):
     """ID of the organization, if known and available."""
     organization_name: str | None = None
     """Name of the organization (requires ORGANIZATION_READER permission)."""
+    agents_enabled: bool | None = None
+    """Whether the workspace is enabled for Airbyte Agents (the Cloud "Context layer").
+
+    `None` when enablement could not be determined, for example because the workspace has
+    no known organization or the Agents API was unreachable.
+    """
```
aaronsteers: Let's remove for now. For now, just connectors and orgs will have the enablement.

airbyte/mcp/cloud.py
```diff
@@ -996,27 +1030,139 @@ def list_deployed_cloud_source_connectors(
...
-    """List all deployed source connectors in the Airbyte Cloud workspace."""
+    """List all deployed source connectors in the Airbyte Cloud workspace.
+
+    Each source reports `external_access_enabled` (usable by Airbyte Agents) and
+    `search_enabled` (Context Store caching configured); pass `with_feature` to return only
+    sources with one of those features.
+    """
+    if limit is not None and limit <= 0:
+        raise PyAirbyteInputError(message="`limit` must be greater than 0.")
     workspace: CloudWorkspace = _get_cloud_workspace(ctx, workspace_id)
```
aaronsteers: Please ALWAYS put a blank line after any statement that returns early from the context.

```suggestion
    if limit is not None and limit <= 0:
        raise PyAirbyteInputError(message="`limit` must be greater than 0.")

workspace: CloudWorkspace = _get_cloud_workspace(ctx, workspace_id)
```

airbyte/mcp/cloud.py
aaronsteers: Let's use the same names for now at least on the org level: `external_access_enabled` and `search_enabled`.

airbyte/mcp/cloud.py
```diff
@@ -996,27 +1030,145 @@ def list_deployed_cloud_source_connectors(
...
+
+def _resolve_parent_organization_id(workspace: CloudWorkspace) -> str | None:
+    """Return the workspace's parent organization ID, or `None` if it cannot be resolved.
+
+    The Agents API needs an explicit organization when credentials span several organizations.
+    """
+    try:
+        return workspace.get_organization().organization_id
+    except _AGENTS_LOOKUP_ERRORS:
+        return None
```
aaronsteers: Did we not already have something like this? This feels bespoke and wrong here. Should probably just be inline where you need it.

airbyte/mcp/cloud.py
```diff
@@ -996,27 +1030,145 @@ def list_deployed_cloud_source_connectors(
...
+
+
+def _list_agent_source_search_status(workspace: CloudWorkspace) -> dict[str, bool | None] | None:
```
aaronsteers: Why isn't this a method on the workspace class?

airbyte/mcp/cloud.py
```diff
@@ -996,27 +1030,145 @@ def list_deployed_cloud_source_connectors(
...
+            search_by_source_id[connector.connector_id] = None
+            continue
+
+        search_by_source_id[connector.connector_id] = bool(
+            readiness is not None and readiness.configured_cache_entities
+        )
+    return search_by_source_id
+
+
+def _is_agent_workspace(workspace: CloudWorkspace) -> bool | None:
```
aaronsteers: Ditto. Should be a workspace member/method.

aaronsteers (commented):
You are breaking the rule of the MCP module being a presentation layer on top of the core modules. Please also make sure this is correctly explained in the CONTRIBUTING.md and add it to this PR if not.

airbyte/mcp/cloud.py
aaronsteers: CloudWorkspace. We are phasing out agents module.

airbyte/mcp/cloud.py
aaronsteers: Okay. Please double check. But in that case, what did you previously implement at the org level?

airbyte/cloud/workspaces.py
```diff
@@ -116,11 +121,17 @@ def __init__(
...
         config_api_root: str | None = None,
         bearer_token: str | SecretString | None = None,
+        organization_id: str | None = None,
     ) -> None:
-        """Validate and initialize credentials."""
+        """Validate and initialize credentials.
+
+        `organization_id` optionally scopes requests that need an explicit organization (for
+        example Airbyte Agents lookups) when the credentials span several organizations.
+        """
```
aaronsteers, line 137: This isn't a good description. We should not need this from the caller. We can take it perhaps as optional to save a lookup - but arguably, we may not want to trust it - so in that case it just becomes a guard to ensure the caller knows where they are and isn't disoriented.

Tell me why you didn't just look it up rather than require as input?

Again, I'm not against optionally accepting it - but I'm not okay with accepting it for the reason you put here in the docstring.

airbyte/cloud/workspaces.py
```diff
@@ -304,6 +315,87 @@ def get_organization(
...
+        except AirbyteError as error:
+            status_code = (error.context or {}).get("status_code")
+            if status_code in {HTTPStatus.FORBIDDEN, HTTPStatus.NOT_FOUND}:
+                return False
+
+            raise
+
+        return True
+
+    def list_agent_source_search_status(self) -> dict[str, bool | None]:
```
aaronsteers: This is not a valid public method. Looks like it should be internal/private method. Or if it needs to be called by a user, then it needs a better name and we need to discuss.

CONTRIBUTING.md
```diff
@@ -10,6 +10,21 @@ Use **Markdown** formatting in all docstrings — not reStructuredText (reST).
...
+public core classes (for example `airbyte.cloud.CloudWorkspace`), where it is reusable and
+unit-testable on its own. An MCP tool should only:
+
+- Parse and validate its inputs.
+- Call one or more core methods.
+- Convert the result into the tool's Pydantic response model (including mapping expected
+  failures to "unknown" values where the tool contract calls for it).
+
+If you find yourself writing a helper in `airbyte/mcp/` that talks to an API or encodes
+domain rules, move it to the core module and call it from the tool instead.
```
aaronsteers:
```suggestion
## MCP Tools Are a Presentation Layer

Modules under `airbyte/mcp/` are thin wrappers over the core library. Business logic — API
calls, status derivation, cross-referencing between Cloud and Agents data — belongs on the
public core classes (for example `airbyte.cloud.CloudWorkspace`), where it is reusable and
unit-testable on its own.

If you find yourself writing a helper in `airbyte/mcp/` that talks to an API or encodes
domain rules, move it to the most appropriate core module or util module and create the necessary public Python interface first.

As a general rule (with rare exceptions), there shouldn't be anything that you can do through the MCP tools which you couldn't also do with the Public Python interface. (Hence the framing as a "presentation" layer on top of the core modules.)

**Comment by aaronsteers** [comment]
Devin, update to resolve merge conflicts.

File: airbyte/cloud/models.py [hunk 9]
```diff
@@ -11,6 +11,25 @@
...
+"""Destination definition ID -> `sql_dialect` value accepted by the `sql_select` action."""
+
+SQL_PASSTHROUGH_DESTINATION_NAMES: Mapping[str, str] = {
+    SNOWFLAKE_DESTINATION_DEFINITION_ID: "Snowflake",
+    BIGQUERY_DESTINATION_DEFINITION_ID: "BigQuery",
+}
+"""Destination definition ID -> display name of the destination integration."""
+
+SQL_PASSTHROUGH_DESTINATION_DEFINITION_IDS = frozenset(SQL_PASSTHROUGH_DESTINATION_DIALECTS)
+"""Destination definitions AI agents can query through SQL passthrough."""
```
aaronsteers: None of this should be part of the public API surface area. You can prefix them with "_".

airbyte/cloud/workspaces.py
```diff
@@ -304,6 +322,153 @@ def get_organization(
...
+
+    @property
+    def search_indexing_enabled(self) -> bool:
+        """Whether search indexing is available in this workspace.
+
+        Search indexing is available wherever external access is enabled; individual
+        sources report whether indexing is configured via
+        `CloudConnector.search_indexing_enabled`.
+        """
+        return self.external_access_enabled
```
aaronsteers: This is not correct. I think technically this is `False` everywhere until Search Indexing goes online for Fusion in a few days.

airbyte/cloud/workspaces.py
```diff
@@ -304,6 +322,153 @@ def get_organization(
...
+
+    def _list_source_search_indexing_status(self) -> dict[str, bool]:
+        """Map each externally accessible source ID to whether search indexing is configured.
+
+        Sources missing from the mapping are not enabled for external access. The mapping is
+        empty when the Context layer API reports the workspace as forbidden or not found.
+        Raises when the API is unreachable or returns a payload that does not match the
+        expected models.
+        """
```
aaronsteers: Ditto. This seems misguided. I think this is just `False` everywhere for now. Lmk if you think that's not correct.

airbyte/cloud/connectors.py
```diff
@@ -91,6 +92,27 @@ def __repr__(self) -> str:
         )
 
 
+class ConnectorType(str, Enum):
```
aaronsteers: Devin, get the StrEnum implementation from the PR above yours. Pull that in here and refactor to use a proper StrEnum implementation everywhere you've touched or created StrEnums.

Also, as noted in that PR, the `ConnectorType` implementation shouldn't be here. It should be declared *already* somewhere - and if not, it should be declared outside of the `airbyte.cloud` module - as the same StrEnum class should be used in local and registry modules.

airbyte/cloud/workspaces.py
```diff
@@ -37,31 +37,40 @@
...
 from airbyte._util.api_util import get_web_url_root
+from airbyte.agents import _api_util as agents_api_util
+from airbyte.agents._destination_docs import SQL_PASSTHROUGH_DESTINATION_DEFINITION_IDS
+from airbyte.agents.models import AgentConnectorDetails, AgentConnectorInfo
+from airbyte.cloud import organizations as cloud_organizations
 from airbyte.cloud._credentials import _AirbyteCredentials
 from airbyte.cloud.client_config import CloudClientConfig
 from airbyte.cloud.connections import CloudConnection
 from airbyte.cloud.connectors import (
+    CloudConnector,
```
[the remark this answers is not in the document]
aaronsteers: Devin, is this resolved now?

airbyte/cloud/workspaces.py
```diff
@@ -37,31 +37,42 @@
...
 from airbyte._util.api_util import get_web_url_root
+from airbyte.cloud import organizations as cloud_organizations
 from airbyte.cloud._credentials import _AirbyteCredentials
 from airbyte.cloud.client_config import CloudClientConfig
 from airbyte.cloud.connections import CloudConnection
 from airbyte.cloud.connectors import (
+    CloudConnector,
     CloudDestination,
     CloudSource,
+    ConnectorFeature,
```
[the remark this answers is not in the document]
aaronsteers: Devin, is this resolved now? Or can it be?

airbyte/cloud/workspaces.py
```diff
@@ -37,31 +37,42 @@
...
+    ConnectorType,
```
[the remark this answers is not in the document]
aaronsteers: Ditto. Devin, is this resolved now? Or can it be? (Should be.)

airbyte/cloud/workspaces.py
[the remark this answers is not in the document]
aaronsteers: Disregard #1211. From this `workspaces` module, I think we should be able to make CloudConnector/CloudSource/CloudDestination into typing-only imports.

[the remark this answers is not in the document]
aaronsteers: If not, then the reverse: the connectors module should be able to use CloudWorkspace as typing-only. I think there's a fix path here somewhere. Work with me on it in slack if you can't figure it out.

aaronsteers:
Superseded by https://github.com/airbytehq/PyAirbyte/pull/1220:
- https://github.com/airbytehq/PyAirbyte/pull/1220

PR #1192, opened 2026-09-18
feat(mcp): hand OAuth consent off to the branded Keycloak login

cjkenned:
> [!IMPORTANT]
> **Draft — blocked on airbytehq/fastmcp-extensions#119.** The pin here is `fastmcp-extensions>=0.24.0`, which does not exist yet, so `uv lock` cannot run and `uv.lock` is deliberately left stale. CI will fail on lock consistency until that PR is released. Once it is: `uv lock`, commit, un-draft.

## What

Two new args on the interactive `OIDCAuthConfig` in `_create_auth` (`airbyte/mcp/server.py`):

```python
require_authorization_consent="external",
extra_authorize_params={"prompt": "consent"},
```

## Why

`OIDCProxy` shows its own consent page before redirecting to Keycloak. That page is rendered from the FastMCP server's `name`, `icons[0]`, and `website_url` (`fastmcp/server/auth/oauth_proxy/consent.py:378-388`), and this server sets neither `icons` nor `website_url`. So today an interactive user signing in to the hosted Cloud MCP sees:

1. a consent page with the **FastMCP logo** and the server named as the raw slug **`airbyte-mcp`**, not hyperlinked, then
2. the actual Airbyte Keycloak login.

Two consent prompts back to back, the first one unrecognizable as Airbyte.

`"external"` skips the built-in page so Keycloak's branded login is the only thing the user sees. `prompt=consent` keeps the IdP from silently reusing an existing browser session and skipping its page too — without it, "skip our page" can turn into "skip every page".

`"external"` rather than `False`: both skip the screen, but `False` logs a "consent disabled" warning on every startup (`oauth_proxy/proxy.py:398-403`), which misdescribes a flow where consent moved upstream rather than disappeared.

This matches what `sonar/agent-engine-mcp` already does in `app/mcp_instance.py`.

## Scope

Branding the FastMCP-rendered pages themselves (`icons=` / `website_url=` on the `mcp_server(...)` call) is a deliberate follow-up. It still matters after this PR for the OAuth **error** pages, which keep using FastMCP chrome.

Whether the Keycloak realm's login page is itself Airbyte-themed is deployment-side config, outside this repo — worth confirming on `_airbyte-cloud-users` before this ships.

## Test plan

Verified locally against an editable install of the fastmcp-extensions branch, resolving **fastmcp 3.2.0** — the version this repo's lock pins, and the floor that matters for the new kwargs:

- `tests/unit_tests/test_mcp_auth.py` — 31 passed, including a new `test_create_auth_oidc_defers_consent_to_the_idp`
- `tests/unit_tests` — 929 passed, 1 skipped (pre-existing network skip)
- `ruff check` / `ruff format --check` clean on changed files
- Confirmed the real `OIDCProxy.__init__` on fastmcp 3.2.0 binds every kwarg `_build_oidc_proxy` now forwards, so this cannot `TypeError` at startup

🤖 Generated with [Claude Code](https://claude.com/claude-code)


aaronsteers (approved):
Approving to unblock. 

Lmk if you need any assist with testing or running a prerelease.

The `/prerelease` slash command should work with PRs on this repo and also in the `fastmcp-extentions` repo, in case it helps.

cjkenned:
tested locally with the fastmcp-extensions prelease from the other pr. the extensions version is published here: https://github.com/airbytehq/fastmcp-extensions/releases/tag/v0.24.0, and pypi here https://pypi.org/project/fastmcp-extensions/

cjkenned:
my python tests ran successfully but the status wasnt reported: https://github.com/airbytehq/PyAirbyte/actions/runs/35631632180
probably a github problem. going to just admin merge this.

